PR #2, opened 2017-11-17
Rewrite first pass [DO NOT MERGE]

bboughton:
Here is my first pass at rewriting this tool in Go. I have separated the work into multiple commits for easier review. The most interesting commit is 05b3710.

### Notes

* I focused mainly on getting the first task of deleting all EC2 instances completed. I have not taken the time to clean up the code, refactor, or add comments. I have attempted to add comments to the main areas that need to be cleaned up. Please feel free to ask about specific sections you have questions on and I will elaborate as to thinking.

* My general approach was to not try to prematurely optimize, so the code strictly focuses on Listing and Destroying EC2 instances. My thinking is that the code will be refactored as other resources are added to make it more general.

* I was able to get an integration test working, but didn't implement any unit test for this first pass.


internal/awsnuke/client.go
```diff
@@ -0,0 +1,116 @@
...
+
+func New(id string) *Client {
+	return &Client{
+		Id:      id,
+		Regions: DefaultRegions,
+	}
+}
+
+func (c *Client) ListNonProtectedEC2Instances() ([]EC2Instance, error) {
+	ctx := context.TODO()
```
bboughton: `context.TODO` will need to be replaced with a real context that has a timeout.

internal/awsnuke/client.go
```diff
@@ -0,0 +1,116 @@
...
+				Name: aws.String("owner-id"),
+				Values: []*string{
+					aws.String(c.Id),
+				},
+			},
+			{
+				Name: aws.String("instance-state-name"),
+				Values: []*string{
+					aws.String("running"),
+					aws.String("pending"),
```
brikis98: What about `stopped`?

internal/awsnuke/client.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+var DefaultRegions = []string{
```
brikis98: Does the AWS SDK not have a full list of regions?

internal/awsnuke/client.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+	}
+	sess := session.Must(session.NewSession())
+	creds := credentials.NewEnvCredentials()
+	for _, region := range c.Regions {
+		ec2client := ec2.New(sess, &aws.Config{
+			Credentials: creds,
+			Region:      aws.String(region),
+		})
+		err := ec2client.DescribeInstancesPagesWithContext(ctx, input, func(page *ec2.DescribeInstancesOutput, lastPage bool) bool {
```
brikis98: Oh, neat, I didn't know the SDK had an operation that automatically handled paging this way.

internal/awsnuke/client.go
```diff
@@ -0,0 +1,116 @@
...
+				Values: []*string{
+					aws.String("running"),
+					aws.String("pending"),
+				},
+			},
+		},
+	}
+	sess := session.Must(session.NewSession())
+	creds := credentials.NewEnvCredentials()
+	for _, region := range c.Regions {
```
brikis98: Consider extracting the inner part of this for-loop into its own function (e.g., `ListNonProtectedEC2InstancesInRegion`).

internal/awsnuke/client.go
```diff
@@ -0,0 +1,116 @@
...
+			return nil, errors.WithStackTraceAndPrefix(err, "failed to list ec2 instances")
+		}
+	}
+	return instances, nil
+}
+
+func (c *Client) DestroyEC2Instances(instances []EC2Instance) error {
+	ctx := context.TODO()
+	sess := session.Must(session.NewSession())
+	creds := credentials.NewEnvCredentials()
```
brikis98: Why pull out the credentials this way? And require them to be from the env?

internal/awsnuke/client_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/aws-nuke/internal/awsnuke"
+)
+
+var AWSAccountId = "353720269506"
```
brikis98: Might want to call it `DefaultAwsAccountId` and add a comment that this is Gruntwork's "Nuclear Wasteland" account for testing.

internal/awsnuke/client_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	instances, err = nuke.ListNonProtectedEC2Instances()
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if expect, actual := 0, len(instances); expect != actual {
```
brikis98: We use [testify](https://github.com/stretchr/testify) to turn these sorts of checks into one-line `assert.Equals(...)` calls.

internal/awsnuke/client_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+		t.Fatalf("failed to create ec2 instance: %v", err)
+	}
+
+	if len(reservation.Instances) < 1 {
+		t.Fatal("failed to create ec2 instance")
+	}
+
+	instance := reservation.Instances[0]
+	state := "running"
+OUTER:
```
brikis98: Instead of this, see [WaitUntilInstanceRunning](http://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.WaitUntilInstanceRunning).

internal/awsnuke/client_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/aws-nuke/internal/awsnuke"
+)
+
+var AWSAccountId = "353720269506"
+
+func TestDestroyEC2Instances(t *testing.T) {
+	createNewInstance(t, "us-east-1", "ami-6057e21a")
```
brikis98: Add `t.Parallel()`.

internal/awsnuke/client_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if expect, actual := 0, len(instances); expect != actual {
+		t.Errorf("expected all instances to have been destroyed")
+	}
+}
+
+func createNewInstance(t *testing.T, region, ami string) {
+	ec2client := newEC2Client(region)
+	input := &ec2.RunInstancesInput{
```
brikis98: Add a tag to the instance so we know this is from the `aws-nuke` tests.

internal/awsnuke/client_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	err = nuke.DestroyEC2Instances(instances)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	instances, err = nuke.ListNonProtectedEC2Instances()
```
brikis98: It's a bit odd to use the library to check for instances, as the library itself may be broken. Moreover, this will fail if this test is running in parallel with other tests that are creating EC2 instances.

Is there some other check you could do? For example, [Terratest](https://github.com/gruntwork-io/terratest) has a function for a short unique ID you can generate and add as a tag to each instance. You can then search for that tag and make sure you can no longer find it.

main.go
```diff
@@ -0,0 +1,67 @@
...
+
+	app.Flags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "non-interactive",
+			Usage: "Assume yes for all interactive prompts",
+		},
+	}
+
+	log := logging.GetLogger("aws-nuke")
+	app.Action = func(c *cli.Context) error {
```
brikis98: Define this in a separate, standalone function. It's more readable and may be reusable in other contexts (e.g., at test time).

main.go
```diff
@@ -0,0 +1,67 @@
...
+		if c.NArg() < 1 {
+			return cli.NewExitError("insufficient number of arguments", 1)
+		}
+
+		id := c.Args().Get(0)
+
+		nuke := awsnuke.New(id)
```
brikis98: I'd split this into several named functions to make it easier to read (and debug and test): e.g., `showInstances`, `confirmDelete`, `deleteInstances`.

main.go
```diff
@@ -0,0 +1,67 @@
...
+		opts := shell.NewShellOptions()
+		opts.NonInteractive = c.Bool("non-interactive")
+		opts.Logger = log
+		yes, err := shell.PromptUserForYesNo("nuke resources?", opts)
+		if err != nil {
+			return cli.NewExitError("failed to retrieve user input", 1)
+		}
+
+		if !yes {
+			log.Info("exiting without deleting resources")
```
brikis98: You're not actually exiting here

main.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		for _, instance := range ec2Instances {
+			log.Infof("ec2 instance: %s", instance.Id)
+		}
+		log.Infof("aws account: %s", id)
+		opts := shell.NewShellOptions()
+		opts.NonInteractive = c.Bool("non-interactive")
+		opts.Logger = log
+		yes, err := shell.PromptUserForYesNo("nuke resources?", opts)
```
brikis98: This prompt should be surrounded with tons of massive warnings about how this operation is destructive and cannot be undone.

README.md
```diff
@@ -1 +1,40 @@
...
+12. Remove all non-default subnets
+13. Remove all non-default route tables
+14. Remove all Simple Storage Service (S3) buckets
+15. Remove all Simple Queue Service (SQS) queues
+16. Remove all Lambda Functions
+17. Remove all CloudFront distributions
+18. Remove all ElasticCache Instances
+19. Time based filtering
+20. Region based filtering
+21. Opt-Out specific aws resources from being nuked (ie EC2)
```
brikis98: Thanks for jotting down this list!

BTW, minor thing, but in Markdown, you can just use the same number (e.g., `1.` every time) and it'll render into a proper numbered list, with no need for you to maintain the numbers by hand.

bboughton: I do use the `1.` method however I have my editor setup to auto format markdown on save and it automatically numbers the numbered list.

brikis98: Oh, cool. What editor is that?

internal/awsnuke/client.go
bboughton: I will have to check. I only looked into it briefly and found these regions in the [endpoints pkg](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants).

internal/awsnuke/client.go
bboughton: Good catch, I was thinking about EC2 instances started by ASG's and how the ASG will auto terminate a stopped instance.

brikis98: Hm, that brings up another interesting point: if these are instances in an ASG, terminating them will not be all that useful. So in the full version of the tool, we'll need to run the ASG termination code before the EC2 termination code...

internal/awsnuke/client_integration_test.go
bboughton: This was a quick fix to get it working. I agree that the assertion should not rely on the Unit Under Test. Will fix.

README.md
bboughton: Vim, I setup the auto formatting myself. It uses `pandoc` with GitHub flavor formatting. Of course I set it up before GitHub [publish](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown) their new official Markdown flavor so mine may be out of sync.

PR #3, opened 2018-01-11
Add support for deleting Auto Scaling Groups.

josh-padnick:
Note that an ASG cannot be deleted until it contains 0 instances, so we first set the desired capacity down to 0, then wait, then delete the ASG.

Also fixed a bug where we were attempting to terminate EC2 Instances that were already terminated.


aws-nuke
```diff
@@ -93,14 +113,29 @@ function aws_nuke {
...
         nuke_kinesis_streams "$region"
+
+        # Security Groups don't cost money and must be deleted in a sequence that respects their dependencies. Commented
+        # out until proper order of delete is implemented.
+        #nuke_security_groups "$region"
+    done
+
+    # Deleting ASGs after we nuke everything else is a hacky way to give set_asgs_num_desired_instances_to_zero enough
+    # time to complete its actions. We sleep an additional 10 seconds before proceeding just to be sure.
+    sleep 10
```
brikis98: There's no way this will be enough. ASGs can easily take 2-5 min to empty out.

brikis98:
We've moved to Go, and this functionality is now part of #5, so closing.

PR #4, opened 2018-01-22
aws-nuke-v0.0.1

tonerdo:
This PR contains a reimplementation of the original aws-nuke bash shell script in golang.

Changes Include:
* Complete rewrite of cli utility in golang
* Support for nuking only EC2 instances
* Unit tests


README.md
```diff
@@ -1 +1,28 @@
-# aws-nuke
\ No newline at end of file
+# aws-nuke
+
+This repo contains a CLI tool to delete all AWS resources in an account. The currently supported functionality includes:
```
brikis98: It's probably a good idea to add two more things to this intro section:

1. Motivation: explain why aws-nuke exists and you might want to use it. E.g., you have an account you use for testing and need to clean up left over resources so AWS doesn't charge you for them.

1. Warning: big, bold text explaining that this tool is destructive, irreversibly deletes things in your account, and that you must be very careful with its use.

README.md
```diff
@@ -1 +1,28 @@
...
+
+* Deleting all unprotected EC2 instances in an AWS account
+
+## Install
+
+1. Download the latest binary for your OS on the [releases page](https://github.com/gruntwork-io/aws-nuke/releases).
+2. Move the binary to a folder on your `PATH`. E.g.: `mv gruntwork_darwin_amd64 /usr/local/bin/gruntwork`.
```
brikis98: Update name of the binary here.

README.md
```diff
@@ -1 +1,28 @@
...
+Happy Nuking!!!
+
+## Credentials
+
+In order for the `aws-nuke` CLI tool to access your AWS, you will need to provide your AWS credentials. You can used one of the [standard AWS CLI credential mechanisms](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html).
+
+## Running Tests
+
+```shell
+go test -v ./...
```
brikis98: Could you add an MIT license to this repo? See the license section in [fetch](https://github.com/gruntwork-io/fetch) for an example.

aws/aws.go
```diff
@@ -0,0 +1,42 @@
...
+
+var regions []string
+
+// GetAllResources - Lists all aws resources
+func GetAllResources() []string {
+	regions = []string{
+		"us-east-1", "us-east-2", "us-west-1", "us-west-2", "ca-central-1",
+		"eu-west-1", "eu-central-1", "eu-west-2", "ap-southeast-1", "ap-southeast-2",
+		"ap-northeast-2", "ap-northeast-1", "ap-south-1", "sa-east-1",
+	}
```
brikis98: I think you can get the list of regions programmatically from this API: https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/

aws/aws.go
```diff
@@ -0,0 +1,42 @@
...
+
+	var resources []string
+	for _, region := range regions {
+		session, _ := session.NewSession(&awsgo.Config{
```
brikis98: I'm guessing the second parameter is an error object. If so, do not ignore it. Although it's verbose, the idiomatic code style in Go is for every function (including this `GetAllResources` function) to return a value (e.g., `[]string` and an error):

```go
func GetAllResources() ([]string, error) {
  // ...
  session, err := session.NewSession(&awsgo.Config{ ... })
  if err != nil {
    return nil, errors.WithStackTrace(err)
  }
}
```

aws/aws.go
```diff
@@ -0,0 +1,42 @@
...
+			Region: awsgo.String(region)},
+		)
+
+		instances, err := getAllEc2Instances(session, region)
+		if err == nil {
```
brikis98: We should probably return the error immediately if it's not `nil`.

aws/aws.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+	}
+
+	return resources
+}
+
+// NukeAllResources - Nukes all aws resources
+func NukeAllResources() {
+	for _, region := range regions {
+		session, _ := session.NewSession(&awsgo.Config{
```
brikis98: Same comment on error handling

aws/aws.go
```diff
@@ -0,0 +1,42 @@
...
+			Region: awsgo.String(region)},
+		)
+
+		nukeAllEc2Instances(session)
```
brikis98: This function should return an error up up the chain.

aws/ec2.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/aws-nuke/logging"
+)
+
+var ec2Instances = make(map[string][]*string)
```
brikis98: I would recommend strongly against stateful variables of this sort. In the future, we'd most likely want to update this code to do work concurrently, and mutable, shared variables will make that very hard.

Instead, the `GetAllResources` function should return this map of EC2 instances and the `nukeAllResources` function should take the map in as an argument. That'll make it easier to test them too. And improve reusability/composability of those functions.

aws/ec2.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Build string that'll be shown in resources list
+func buildEntryName(instance ec2.Instance) string {
+	return fmt.Sprintf("ec2-%s-%s", *instance.InstanceId, *instance.InstanceType)
```
brikis98: There is an `aws.StringValue` function that is a `nil`-safe way to turn a pointer into a string.

aws/ec2.go
```diff
@@ -0,0 +1,84 @@
...
+// Returns a formatted string of EC2 instance ids
+func getAllEc2Instances(session *session.Session, region string) ([]string, error) {
+	svc := ec2.New(session)
+
+	params := &ec2.DescribeInstancesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name: awsgo.String("instance-state-name"),
+				Values: []*string{
+					awsgo.String("running"), awsgo.String("pending"),
```
brikis98: Any reason to filter by state? For example, any reason not to nuke `stopped` instances?

aws/ec2.go
```diff
@@ -0,0 +1,84 @@
...
+				},
+			},
+		},
+	}
+
+	output, err := svc.DescribeInstances(params)
+	if err != nil {
+		return nil, err
```
brikis98: We wrap all errors returned by external APIs (i.e., not gruntwork code) with `errors.WithStackTrace` from `gruntwork-cli`. This gives us a stack trace for error messages, which is very useful for debugging.

aws/ec2.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	var entries []string
+	for _, reservation := range output.Reservations {
+		for _, instance := range reservation.Instances {
+			instanceID := *instance.InstanceId
+
+			attr, _ := svc.DescribeInstanceAttribute(&ec2.DescribeInstanceAttributeInput{
```
brikis98: Don't ignore errors. It's annoying to have `if err != nil` checks everywhere, but it's far more annoying when things fail silently!

aws/ec2.go
```diff
@@ -0,0 +1,84 @@
...
+	return entries, nil
+}
+
+// Deletes all non protected EC2 instances
+func nukeAllEc2Instances(session *session.Session) error {
+	svc := ec2.New(session)
+	instances := ec2Instances[*session.Config.Region]
+
+	if len(instances) == 0 {
+		return nil
```
brikis98: You should log something in this case (e.g., "no instances to nuke").

aws/ec2_test.go
```diff
@@ -0,0 +1,59 @@
...
+	)
+
+	instance := createTestEC2Instance(t, session)
+	instanceEntries, err := getAllEc2Instances(session, "us-west-2")
+
+	if err != nil {
+		assert.Fail(t, "Unable to fetch list of EC2 Instances")
+	}
+
+	assert.NotEqual(t, 0, len(instanceEntries))
```
brikis98: We often run tests in parallel, and I'm afraid if some other tests happens to fire up an EC2 instance in the same account, this test will pass, even if the code isn't actually working. Consider tagging the instances you launch from the aws-nuke test and make sure the instance you find with `getAllEc2Instances` includes the tag you expect.

In fact, it's probably a good idea to tag ALL resources created by aws-nuke tests with "aws-nuke-test-<UNIQUE_ID>" so if aws-nuke's own tests leave resources around, we know what the cause is. We typically use this [UniqueId](https://github.com/gruntwork-io/terratest/blob/06be4e266055d4512a35d3069c43738133eb71b5/util/random.go#L24) function to generate unique IDs for all our tests.

aws/ec2_test.go
```diff
@@ -0,0 +1,59 @@
...
+	assert.Contains(t, instanceEntries, buildEntryName(instance))
+}
+
+func TestNukeInstances(t *testing.T) {
+	session, _ := session.NewSession(&awsgo.Config{
```
brikis98: error handling

aws/ec2_test.go
```diff
@@ -0,0 +1,59 @@
...
+	)
+
+	nukeAllEc2Instances(session)
+	instanceEntries, err := getAllEc2Instances(session, "us-west-2")
+
+	if err != nil {
+		assert.Fail(t, "Unable to fetch list of EC2 Instances")
+	}
+
+	assert.Equal(t, 0, len(instanceEntries))
```
brikis98: This won't work if other things are deploying into the same account or this test is running multiple times in parallel. See my note on tagging above.

commands/cli.go
```diff
@@ -0,0 +1,52 @@
...
+
+// CreateCli - Create the CLI app with all commands, flags, and usage text configured.
+func CreateCli(version string) *cli.App {
+	app := cli.NewApp()
+
+	app.Name = "aws-nuke"
+	app.HelpName = app.Name
+	app.Author = "Gruntwork <www.gruntwork.io>"
+	app.Version = version
+	app.Usage = "A CLI tool to cleanup AWS resources"
```
brikis98: We should definitely have a big warning text in the usage description!

brikis98: As well as a list of what resources are currently supported

commands/cli.go
```diff
@@ -0,0 +1,52 @@
...
+	if len(resources) == 0 {
+		logging.Logger.Infoln("Nothing to nuke, you're all good!")
+		return nil
+	}
+
+	for _, resource := range resources {
+		logging.Logger.Infoln(resource)
+	}
+
+	prompt := "\nAre you sure you want to nuke all listed resources"
```
brikis98: Add a big, huge WARNING here. Yell at the user to tell them this is irreversible! Make the warning red ([example of using colors on the CLI here](https://github.com/gruntwork-io/gruntwork-cli/blob/master/shell/prompt.go#L13)).

aws/ec2.go
tonerdo: I filter by state so that I don't get EC2 instances that have already been `terminated` but not yet deleted (i.e. they still appear in the instances list). As for `stopped` instances, my thinking is that since they're aren't incurring any charges there's no need to nuke them. Although now that I think about it, in the spirit of the tool it should just nuke 'em all

brikis98: yea, I'd nuke everything to keep it simple :)

aws/ec2_test.go
tonerdo: * This particular test line simply checks that the creation of an instance succeeded and the next line ensures that the instance created by the test is present in the returned list. If in the `TestNukeInstances` we ensure we only delete appropriately tagged instances then, I feel this line will be fine

tonerdo:
> Add a big, huge WARNING here. Yell at the user to tell them this is irreversible! Make the warning red (example of using colors on the CLI here).

The text is printed by the `PromptUserForYesNo` function which defaults to `BRIGHT_GREEN` with no way of specifying a custom color

brikis98:
> The text is printed by the PromptUserForYesNo function which defaults to BRIGHT_GREEN with no way of specifying a custom color

Sure, but you can add a separate logging statement just before that in red! This is a hugely dangerous tools, so we need to very seriously warn users. In fact, even "yes/no" might not be the right prompt here. We might want to force users to type "nuke" or something where you can't possibly do it by accident.

aws/ec2_test.go
brikis98: Consider the following:

1. We have two automated tests running in the same AWS account at the same time. The first test, `foo`, creates an EC2 instance. The second test, this one for `aws-nuke`, tries to create the EC2 Instance, but due to some bug in the test, fails.
1. Now you do your check, which is merely seeing that there are more than 0 instances created. The check will pass, as `foo` created an EC2 Instance, but it *should* fail, as the instance created by `aws-nuke` didn't actually succeed.

README.md
tonerdo: Done!

aws/ec2.go
tonerdo: I updated this to nuke every state except `terminated` and `shutting-down`. `terminated` is for already terminated but not cleaned up instances, `shutting-down` is for instances that are in the process of being terminated I didn't see a reason to include this two since they represent already terminated or being terminated instances

aws/ec2_test.go
tonerdo: Ah makes sense, thanks for the clarification. I've removed this line and I now specifically check for the presence of the instance that was created

tonerdo:
> We might want to force users to type "nuke" or something where you can't possibly do it by accident.

Good idea!

README.md
```diff
@@ -1 +1,36 @@
...
\ No newline at end of file
+# aws-nuke
+
+This repo contains a CLI tool to delete all AWS resources in an account. aws-nuke was created for situations when you might have an account you use for testing and need to clean up left over resources so AWS doesn't charge you for them. Also great for cleaning out accounts with redundant resources.
+
+The currently supported functionality includes:
+
+* Deleting all unprotected EC2 instances in an AWS account
+
+### WARNING: THIS TOOL IS HIGHLY DESTRUCTIVE, ALL SUPPORTED RESOURCES WILL BE DELETED. ITS EFFECTS ARE IRREVERSIBLE AND SHOULD NEVER BE USED IN A PRODUCTION ENVIRONMENT
```
brikis98: 👍

README.md
```diff
@@ -1 +1,36 @@
...
+
+## Install
+
+1. Download the latest binary for your OS on the [releases page](https://github.com/gruntwork-io/aws-nuke/releases).
+2. Move the binary to a folder on your `PATH`. E.g.: `mv gruntwork_darwin_amd64 /usr/local/bin/aws-nuke`.
```
brikis98: s/gruntwork_darwin_amd64/aws-nuke_darwin_amd64/

aws/aws.go
```diff
@@ -0,0 +1,70 @@
...
+		for id := range p.Regions() {
+			regions = append(regions, id)
+		}
+	}
+
+	return regions
+}
+
+// GetAllResources - Lists all aws resources
+func GetAllResources() (map[string]map[string][]*string, error) {
```
brikis98: Hm, that's quite a nested return value. Very hard to know what'll be in it! I had to read through all the code to figure out that it's a map of resource type -> map of region ->  resource unique ID. I think.

This is a perfect opportunity to create some custom `struct` types with named parameters. Example:

```go
type AwsAccountResources {
  ResourcesInRegion map[string]AwsRegionResources 
}

type AwsRegionResources struct {
  Instances []Ec2Instance
}

type Ec2Instance struct {
  InstanceId string
}

func GetAllResources() (*AwsAccountResources, error) {

}
```

Now I can read the function signature and types and know exactly what is being returned.

aws/aws.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		resources["ec2"][region] = instanceIds
+	}
+
+	return resources, nil
+}
+
+// NukeAllResources - Nukes all aws resources
+func NukeAllResources(resourceMaps map[string]map[string][]*string) error {
```
brikis98: Same thing here. To someone who isn't intimately familiar with the code, there's no way to guess what this map of maps should contain. See my comment above on using types to make this easier to understand.

aws/aws.go
```diff
@@ -0,0 +1,70 @@
...
+		session, err := session.NewSession(&awsgo.Config{
+			Region: awsgo.String(region)},
+		)
+
+		if err != nil {
+			return errors.WithStackTrace(err)
+		}
+
+		err = nukeAllEc2Instances(session, resourceMaps["ec2"][region])
+		if err != nil {
```
brikis98: In go, you can combine these into one statement:

```go
if err := nukeAllEc2Instances(session, resourceMaps["ec2"][region]); err != nil {

}
```

aws/ec2.go
brikis98: Sounds good!

aws/ec2.go
```diff
@@ -0,0 +1,91 @@
...
+				return nil, errors.WithStackTrace(err)
+			}
+
+			protected := *attr.DisableApiTermination.Value
+			// Exclude protected EC2 instances
+			if !protected {
+				instanceIds = append(instanceIds, &instanceID)
+			}
+		}
+	}
```
brikis98: I'd take all the code in the loop above and put it into a function called something like `filterOutProtectedInstances`. It'll make it easier to know what this code is doing without having to scan every line.

aws/ec2.go
```diff
@@ -0,0 +1,91 @@
...
+			},
+		},
+	}
+
+	output, err := svc.DescribeInstances(params)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	var instanceIds []*string
```
brikis98: Why an array of pointers? Why not just an array of plain strings?

aws/ec2.go
```diff
@@ -0,0 +1,91 @@
...
+// Build string that'll be shown in resources list
+func buildEntryName(instance ec2.Instance) string {
+	return fmt.Sprintf(
+		"ec2-%s-%s",
+		awsgo.StringValue(instance.InstanceId),
+		awsgo.StringValue(instance.InstanceType),
+	)
+}
+
+// Returns a formatted string of EC2 instance ids
```
brikis98: Update the comment, as these IDs are not formatted. Also, mention this function returns just instances that are not already terminated and not protected.

aws/ec2_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	runResult, err := svc.RunInstances(params)
+	if err != nil {
+		assert.Failf(t, "Could not create test EC2 instance: %s", errors.WithStackTrace(err).Error())
+	}
+
+	// Add test tag to the created instance
+	_, err = svc.CreateTags(&ec2.CreateTagsInput{
+		Resources: []*string{runResult.Instances[0].InstanceId},
```
brikis98: I wonder if, due to AWS eventual consistency, this API call can fail, as the information about this instance ID has not yet propagated... You may need to call [WaitUntilInstanceExists](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.WaitUntilInstanceExists).

aws/ec2_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+var uniqueTestID = "aws-nuke-test-" + uniqueID()
```
brikis98: Create this variable within each test function and not out here, as multiple test functions may run in parallel, and should each have their own unique IDs.

In fact, as a general rule, avoid any sort of global state by default!

aws/ec2_test.go
```diff
@@ -0,0 +1,132 @@
...
+func TestNukeInstances(t *testing.T) {
+	session, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String("us-west-2")},
+	)
+
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+
+	output, err := ec2.New(session).DescribeInstances(&ec2.DescribeInstancesInput{})
```
brikis98: This code here, plus the for-loop below, should be extracted into a method called `findEC2InstancesByNameTag(name string)` to make it easier to read this code.

aws/ec2_test.go
```diff
@@ -0,0 +1,132 @@
...
+					if *tag.Value == uniqueTestID {
+						instanceIds = append(instanceIds, &instanceID)
+					}
+				}
+			}
+
+		}
+	}
+
+	nukeAllEc2Instances(session, instanceIds)
```
brikis98: Check for an error returned by this method!

commands/cli.go
```diff
@@ -0,0 +1,63 @@
...
+
+	for resource, resourceMap := range resources {
+		for region, instances := range resourceMap {
+			for _, instance := range instances {
+				logging.Logger.Infof("%s-%s-%s", resource, *instance, region)
+			}
+		}
+	}
+
+	prompt := "\nAre you sure you want to nuke all listed resources? Enter 'nuke' to confirm: "
```
brikis98: I would still add a log line before this with a big red warning. It's very important to get the user's attention here.

aws/ec2.go
tonerdo: The call to terminate instances uses an array of string pointers. Easier to just get them here as opposed to trying to build the pointer array from a plain string array

aws/ec2.go
tonerdo: I'm removing this function as it's no longer being used. My last refactor removed the need for it, just forgot to delete it

aws/ec2.go
brikis98: I recommend optimizing the code for readability and maintenance. It's easier to work with an array of plain strings and it communicates the intent better; an array of pointers makes it sound like some of them could be `nil` or you should be modifying them for some reason, neither of which is true.

It's very easy to turn an array of strings into an array of string pointers for the AWS API call: just wrap it with [aws.StringSlice](https://docs.aws.amazon.com/sdk-for-go/api/aws/#StringSlice).

aws/ec2_test.go
tonerdo: Done

aws/aws.go
```diff
@@ -0,0 +1,85 @@
...
+			return nil, errors.WithStackTrace(err)
+		}
+
+		regionResources := AwsRegionResources{
+			RegionName:          region,
+			ResourceIdentifiers: awsgo.StringValueSlice(instanceIds),
+		}
+
+		account.Resources["ec2"] = append(account.Resources["ec2"], regionResources)
```
brikis98: I'm not sure I like having a map with arbitrary resource strings in it. It won't be obvious to a maintainer later what kind of resources may be in the map and it's prone to typos ("EC2" vs "ec2"). 

I'd recommend using the type system here and having lists of specific resources types, such as `EC2Instance` and (in the future), `ASG`, `ELB`, etc.

brikis98: Ah, in fact, given your need to loop over all the resources and nuke them, it probably makes sense to do something like this:

```go
interface AwsResources {
  Nuke(session *aws.Session) error 
}

type AwsRegionResources struct {
	RegionName         string
	Resources          []AwsResources
}

type EC2Instances struct {
  InstanceIds []string
}

// Implement the AwsResources interface
func (instance EC2Instance) Nuke(session *aws.Session) error {
    // This is what's currently in the nukeAllEc2Instances method
}
```

Now `NukeAllResources` can just loop over all the resources:

```go
for _, region := range getAllRegions() {
  resourcesInRegion := account.Resources[region]
  for _, resources := resourcesInRegion.Resources {
    if err := resources.Nuke(session); err != nil {
      // ...
    }
  }
}
```

aws/ec2_test.go
tonerdo: Done

PR #5, opened 2018-02-05
aws-nuke v0.0.2

tonerdo:
Adds support for nuking:

- Auto Scale Groups
- Elastic Load Balancers (Classic and v2)
- EBS Volumes

### DIFF
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -51,6 +51,7 @@ func createTestAutoScalingGroup(t *testing.T, session *session.Session, name str
 }
 
 func TestListAutoScalingGroups(t *testing.T) {
+	t.Parallel()
 	session, err := session.NewSession(&awsgo.Config{
 		Region: awsgo.String("us-west-2")},
 	)
@@ -71,6 +72,7 @@ func TestListAutoScalingGroups(t *testing.T) {
 }
 
 func TestNukeAutoScalingGroups(t *testing.T) {
+	t.Parallel()
 	session, err := session.NewSession(&awsgo.Config{
 		Region: awsgo.String("us-west-2")},
 	)
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -72,6 +72,9 @@ func TestListAutoScalingGroups(t *testing.T) {
 	}
 
 	assert.Contains(t, awsgo.StringValueSlice(groupNames), groupName)
+
+	// clean up after this test
+	defer nukeAllAutoScalingGroups(session, []*string{&groupName})
 }
 
 func TestNukeAutoScalingGroups(t *testing.T) {
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/aws-nuke/util"
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 	"github.com/stretchr/testify/assert"
 )
@@ -62,7 +63,7 @@ func TestListAutoScalingGroups(t *testing.T) {
 		assert.Fail(t, errors.WithStackTrace(err).Error())
 	}
 
-	groupName := "aws-nuke-test-" + uniqueID()
+	groupName := "aws-nuke-test-" + util.UniqueID()
 	createTestAutoScalingGroup(t, session, groupName)
 
 	groupNames, err := getAllAutoScalingGroups(session, region)
@@ -86,7 +87,7 @@ func TestNukeAutoScalingGroups(t *testing.T) {
 		assert.Fail(t, errors.WithStackTrace(err).Error())
 	}
 
-	groupName := "aws-nuke-test-" + uniqueID()
+	groupName := "aws-nuke-test-" + util.UniqueID()
 	createTestAutoScalingGroup(t, session, groupName)
 
 	_, err = svc.DescribeAutoScalingGroups(&autoscaling.DescribeAutoScalingGroupsInput{
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -50,6 +50,8 @@ func TestListAutoScalingGroups(t *testing.T) {
 
 	groupName := "aws-nuke-test-" + util.UniqueID()
 	createTestAutoScalingGroup(t, session, groupName)
+	// clean up after this test
+	defer nukeAllAutoScalingGroups(session, []*string{&groupName})
 
 	groupNames, err := getAllAutoScalingGroups(session, region)
 	if err != nil {
@@ -57,9 +59,6 @@ func TestListAutoScalingGroups(t *testing.T) {
 	}
 
 	assert.Contains(t, awsgo.StringValueSlice(groupNames), groupName)
-
-	// clean up after this test
-	defer nukeAllAutoScalingGroups(session, []*string{&groupName})
 }
 
 func TestNukeAutoScalingGroups(t *testing.T) {
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -1,0 +1,112 @@
+package aws
+
+import (
+	"testing"
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func createTestAutoScalingGroup(t *testing.T, session *session.Session, name string) {
+	svc := autoscaling.New(session)
+	instance := createTestEC2Instance(t, session, name)
+
+	// EC2 Instance must be in a running state before ASG can be created
+	err := ec2.New(session).WaitUntilInstanceRunning(&ec2.DescribeInstancesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name:   awsgo.String("instance-id"),
+				Values: []*string{instance.InstanceId},
+			},
+		},
+	})
+
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+
+	param := &autoscaling.CreateAutoScalingGroupInput{
+		AutoScalingGroupName: &name,
+		InstanceId:           instance.InstanceId,
+		MinSize:              awsgo.Int64(1),
+		MaxSize:              awsgo.Int64(2),
+	}
+
+	_, err = svc.CreateAutoScalingGroup(param)
+	if err != nil {
+		assert.Failf(t, "Could not create test ASG: %s", errors.WithStackTrace(err).Error())
+	}
+
+	err = svc.WaitUntilGroupExists(&autoscaling.DescribeAutoScalingGroupsInput{
+		AutoScalingGroupNames: []*string{&name},
+	})
+
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+}
+
+func TestListAutoScalingGroups(t *testing.T) {
+	session, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String("us-west-2")},
+	)
+
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+
+	groupName := "aws-nuke-test-" + uniqueID()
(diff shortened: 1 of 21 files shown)


aws/ebs_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func createTestEBSVolume(t *testing.T, session *session.Session, name string) ec2.Volume {
+	svc := ec2.New(session)
+	volume, err := svc.CreateVolume(&ec2.CreateVolumeInput{
+		AvailabilityZone: awsgo.String("us-west-2a"),
```
brikis98: I just noticed that these tests all hard-code the region. The downside to this is if we have some region-specific bug, we won't catch it.

You can use [this method](https://github.com/gruntwork-io/terratest/blob/master/aws/region.go#L20) to pick a random one.

aws/ebs_test.go
```diff
@@ -0,0 +1,117 @@
...
+	return *volume
+}
+
+func findEBSVolumesByNameTag(output *ec2.DescribeVolumesOutput, name string) []*string {
+	var volumeIds []*string
+	for _, volume := range output.Volumes {
+		// Retrive only IDs of instances with the unique test tag
+		for _, tag := range volume.Tags {
+			if *tag.Key == "Name" {
+				if *tag.Value == name {
```
brikis98: Instead of a second level of nesting, you could combine the two if-statements: `if *tag.Key == "Name" && *tag.Value == name`.

brikis98: Also, I wonder if there is a chance `tag.Key` or `tag.Value` could be a nil pointer? Might be safer to use `aws.StringValue`.

aws/ebs_test.go
```diff
@@ -0,0 +1,117 @@
...
+	createTestEC2Instance(t, session, uniqueTestID)
+
+	output, err := ec2.New(session).DescribeVolumes(&ec2.DescribeVolumesInput{})
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+
+	volumeIds := findEBSVolumesByNameTag(output, uniqueTestID)
+
+	if err := nukeAllEbsVolumes(session, volumeIds); err != nil {
```
brikis98: Does this method wait for all volumes to be deleted? Or is it possible for it to return while deletion is still in progress, causing the next check to fail?

aws/ec2.go
```diff
@@ -85,6 +85,20 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
 		return errors.WithStackTrace(err)
 	}
 
+	err = svc.WaitUntilInstanceTerminated(&ec2.DescribeInstancesInput{
```
brikis98: Ah, yea, this seems like a good addition. I'd recommend adding equivalent "wait until XXX deleted" to the EBS, ASG, and ELB code.

aws/elb_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	elbNames, err := getAllElbInstances(session, "us-west-2")
+	if err != nil {
+		assert.Fail(t, "Unable to fetch list of Auto Scaling Groups")
+	}
+
+	assert.Contains(t, awsgo.StringValueSlice(elbNames), elbName)
+}
+
+func TestNukeELBs(t *testing.T) {
```
brikis98: How long do the tests take to run? It seems like it would take a while.

You should probably add:

```go
t.Parallel()
```

As the first line of each test. This will get them all to run in parallel. It will also flush out any concurrency bugs.

aws/elbv2_test.go
```diff
@@ -0,0 +1,116 @@
...
+			subnet2.SubnetId,
+		},
+	}
+
+	result, err := svc.CreateLoadBalancer(param)
+	if err != nil {
+		assert.Failf(t, "Could not create test ELBv2: %s", errors.WithStackTrace(err).Error())
+	}
+
+	balancer := *result.LoadBalancers[0]
```
brikis98: Could this list be empty? Prob worth checking and returning an error if it is.

aws/asg_test.go
```diff
@@ -0,0 +1,112 @@
...
+	err = svc.WaitUntilGroupExists(&autoscaling.DescribeAutoScalingGroupsInput{
+		AutoScalingGroupNames: []*string{&name},
+	})
+
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+}
+
+func TestListAutoScalingGroups(t *testing.T) {
```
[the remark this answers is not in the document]
tonerdo: Typically I run aws-nuke to clean up the resources the tests leave around. I feel it's a good exercise of the functionality of the tool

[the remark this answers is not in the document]
brikis98: Well, we'll be hooking up the tests to run automatically after every commit to the `aws-nuke` repo, so we can't rely on manual cleanup. I'd rather have the tests make an effort to clean up after themselves *and* we'll have a scheduled run of aws-nuke in our AWS accounts.

aws/ec2.go
tonerdo: This has been done for everything except classic ELB which doesn't have this functionality

aws/ebs_test.go
tonerdo: It does now

aws/ebs.go
```diff
@@ -0,0 +1,57 @@
...
+	logging.Logger.Infof("Deleting all EBS volumes in region %s", *session.Config.Region)
+
+	for _, volumeID := range volumeIds {
+		params := &ec2.DeleteVolumeInput{
+			VolumeId: volumeID,
+		}
+
+		_, err := svc.DeleteVolume(params)
+		if err != nil {
+			// Ignore not found errors, some volumes are deleted along with EC2 Instances
```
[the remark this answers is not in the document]
tonerdo: This is done as well

aws/elb_test.go
tonerdo: Tests now run and pass concurrently

aws/ec2.go
brikis98: If there's no built-in `WaitUntilXXX` method, you can create your own! You just create a retry loop (see [DoWithRetry](https://github.com/gruntwork-io/terratest/blob/06be4e266055d4512a35d3069c43738133eb71b5/util/retry.go#L30)), looking up the ELB by name until it no longer exists or max retries is exceeded.

aws/aws.go
```diff
@@ -1,34 +1,57 @@
...
-func GetAllResources() (*AwsAccountResources, error) {
+func GetAllResources(excludedRegions []string) (*AwsAccountResources, error) {
 	account := AwsAccountResources{
 		Resources: make(map[string]AwsRegionResource),
 	}
 
 	for _, region := range getAllRegions() {
+		// Ignore all cli excluded regions
+		if collections.ListContainsElement(excludedRegions, region) {
+			continue
```
brikis98: Might be a good idea to log either "Processing region XXX" or "Skipping region XXX".

aws/aws.go
```diff
@@ -39,6 +62,46 @@ func GetAllResources() (*AwsAccountResources, error) {
 
 		resourcesInRegion := AwsRegionResource{}
 
+		// ASG Names
+		groupNames, err := getAllAutoScalingGroups(session, region)
```
brikis98: You might want to add a comment near the top of this method that the order of these resources matters due to dependencies between them: e.g., ASGs first, then EC2 instances, then EBS volumes.

aws/ebs.go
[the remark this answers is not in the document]
brikis98: It's probably also a good idea to add a log statement with the ID/name of *each* EBS volume, instance, ASG, etc. that gets nuked. That'll be very useful for debugging and seeing the progress of the code.

aws/ebs_test.go
```diff
@@ -0,0 +1,125 @@
...
+		assert.Failf(t, "Could not tag EBS volume: %s", errors.WithStackTrace(err).Error())
+	}
+
+	return *volume
+}
+
+func findEBSVolumesByNameTag(output *ec2.DescribeVolumesOutput, name string) []*string {
+	var volumeIds []*string
+	for _, volume := range output.Volumes {
+		// Retrive only IDs of instances with the unique test tag
```
brikis98: s/Retrive/Retrieve/

aws/ec2_test.go
```diff
@@ -1,49 +1,66 @@
...
-
-	const BASE_62_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
-	const UNIQUE_ID_LENGTH = 6 // Should be good for 62^6 = 56+ billion combinations
-
-	var out bytes.Buffer
-
-	rand := rand.New(rand.NewSource(time.Now().UnixNano()))
-	for i := 0; i < UNIQUE_ID_LENGTH; i++ {
-		out.WriteByte(BASE_62_CHARS[rand.Intn(len(BASE_62_CHARS))])
+func getLinuxAmiIDForRegion(region string) string {
```
brikis98: This will be a pain to maintain as regions are added or AMIs removed. You should be able to find the AMI automatically using Amazon's APIs, similar to Packer's [source_ami_filter](https://www.packer.io/docs/builders/amazon-ebs.html#source_ami_filter).

commands/cli.go
```diff
@@ -29,7 +35,7 @@ func CreateCli(version string) *cli.App {
 func awsNuke(c *cli.Context) error {
 	logging.Logger.Infoln("Retrieving all active AWS resources")
 
-	account, err := aws.GetAllResources()
+	account, err := aws.GetAllResources(c.StringSlice("exclude"))
```
brikis98: You should validate the list of regions passed in to ensure each one corresponds to a valid AWS region. We don't want to blow away a region because someone made a typo (e.g., us-wes-1).

aws/asg_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func createTestAutoScalingGroup(t *testing.T, session *session.Session, name string) {
+	svc := autoscaling.New(session)
+	instance := createTestEC2Instance(t, session, name)
+
+	// EC2 Instance must be in a running state before ASG can be created
+	err := ec2.New(session).WaitUntilInstanceRunning(&ec2.DescribeInstancesInput{
```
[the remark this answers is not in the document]
tonerdo: Ah missed this. Thanks

aws/ebs.go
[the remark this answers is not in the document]
tonerdo: Logging has been updated to reflect this

aws/asg_test.go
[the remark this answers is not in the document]
tonerdo: List tests now cleanup after themselves

aws/elb.go
```diff
@@ -0,0 +1,77 @@
...
+
+func waitUntilElbDeleted(svc *elb.ELB, input *elb.DescribeLoadBalancersInput) error {
+	for i := 0; i < 30; i++ {
+		_, err := svc.DescribeLoadBalancers(input)
+		if err != nil {
+			// an error is returned when ELB no longer exists
+			return nil
+		}
+
+		time.Sleep(1 * time.Second)
```
brikis98: Log here (debug) that you're waiting for the EBS volume to be deleted so it's clear what's going on.

aws/elb.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	panic("ELBs failed to delete")
```
brikis98: Return an error. Don't panic.

commands/cli.go
```diff
@@ -19,17 +22,32 @@ func CreateCli(version string) *cli.App {
...
 // Nuke it all!!!
 func awsNuke(c *cli.Context) error {
-	logging.Logger.Infoln("Retrieving all active AWS resources")
+	regions := aws.GetAllRegions()
+	excludedRegions := c.StringSlice("exclude-region")
+
+	for _, excludedRegion := range excludedRegions {
+		if !collections.ListContainsElement(regions, excludedRegion) {
+			fmt.Println(excludedRegion + "is not a valid AWS Region")
+			return InvalidFlagError{}
```
brikis98: Instead of the error message in a println and a separate error with no info, update `InvalidFlagError` to have the flag name and value as fields and to have an `Error()` method that clearly shows them to the user.

commands/cli.go
```diff
@@ -19,17 +22,32 @@ func CreateCli(version string) *cli.App {
...
+		}
+	}
 
-	account, err := aws.GetAllResources()
+	fmt.Println("Retrieving all active AWS resources")
```
brikis98: Use the logger rather than `fmt.Println`. Applies everywhere in this code.

PR #8, opened 2018-02-14
Add nuking resources older than a particular time

tonerdo:


README.md
```diff
@@ -30,6 +30,14 @@ You can use the `--exclude-region` flag to exclude resources in certain regions
 aws-nuke --exclude-region ap-south-1 --exclude-region ap-south-2
 ```
 
+### Excluding Resources by Age
+
+You can use the `--exclude-since` flag to exclude resources created after a certain time, any resources older than that time will be deleted. For example the following command does not nuke resources created after the midnight on 1st Jan, 2017:
+
+```shell
+aws-nuke --exclude-since '01-01-2017 12:00AM'
```
brikis98: Perhaps `--exclude-after` would be better? Perhaps in the future, it might make sense to add an `--exclude-before` as well.

commands/cli.go
```diff
@@ -41,14 +46,26 @@ func awsNuke(c *cli.Context) error {
...
+	var err error
+
+	if c.String("exclude-since") != "" {
+		excludeSince, err = time.Parse("01-02-2006 03:04AM", c.String("exclude-since"))
+		if err != nil {
+			return errors.WithStackTrace(err)
+		}
+	} else {
+		excludeSince = time.Now()
+	}
```
brikis98: I'd extract the logic above into its own method with a signature like `parseTime(paramName string, c *cli.Context) (time.Time, error)`. It would be more readable, reusable, testable, and you could then simply return the parsed value or an error without any mutable variables.

aws/asg_test.go
```diff
@@ -53,7 +54,7 @@ func TestListAutoScalingGroups(t *testing.T) {
 	// clean up after this test
 	defer nukeAllAutoScalingGroups(session, []*string{&groupName})
 
-	groupNames, err := getAllAutoScalingGroups(session, region)
+	groupNames, err := getAllAutoScalingGroups(session, region, time.Now().Add(1*time.Hour))
```
brikis98: It's VERY important to check this logic actually excludes things properly!

I'd add a check where you create an instance, sleep for 10 seconds, and then call this method with the `excludeSince` param set to 1 minute ago. You can then check that the instance is NOT returned. If that works, you then set `excludeSince` to 1 hour from now and make sure it IS returned.

The same applies to the tests for all the other resources (ELBs, ASGs, etc) as well.

README.md
tonerdo: `--exclude-after` sounds better, I had a hard time looking for a better name

README.md
```diff
@@ -30,6 +30,14 @@ You can use the `--exclude-region` flag to exclude resources in certain regions
 aws-nuke --exclude-region ap-south-1 --exclude-region ap-south-2
 ```
 
+### Excluding Resources by Age
+
+You can use the `--exclude-after` flag to exclude resources created after a certain time, any resources older than that time will be deleted. For example the following command does not nuke resources created after the midnight on 1st Jan, 2017:
+
+```shell
+aws-nuke --exclude-after '01-01-2017 12:00AM'
```
josh-padnick: Would supporting a duration make more sense? I'm imagining a CI job that would always want to delete all resources older than, say, a day, rather than having to manually compute the date each time.

aws/ec2.go
```diff
@@ -26,7 +28,7 @@ func filterOutProtectedInstances(svc *ec2.EC2, output *ec2.DescribeInstancesOutp
 
 			protected := *attr.DisableApiTermination.Value
 			// Exclude protected EC2 instances
-			if !protected {
+			if !protected && excludeAfter.After(*instance.LaunchTime) {
 				filteredIds = append(filteredIds, &instanceID)
 			}
```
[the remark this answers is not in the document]
tonerdo: Good catch! Thanks

README.md
tonerdo: I thought of this approach but it was hard for me to come up with a good syntax approach to that. For example if we had an `--older-than` flag what would the smallest unit be? days, minutes, seconds? Regardless of the one we go with there's gonna be some math involved if the user wants to ignore within a time period which I believe is more cognitive load than just looking up a date. More thoughts on this are welcome

brikis98: I actually like @josh-padnick's idea quite a bit. You can use syntax like `1d`, `5h`. This can be parsed using: https://golang.org/pkg/time/#ParseDuration

README.md
```diff
@@ -30,6 +30,14 @@ You can use the `--exclude-region` flag to exclude resources in certain regions
 aws-nuke --exclude-region ap-south-1 --exclude-region ap-south-2
 ```
 
+### Excluding Resources by Age
+
+You can use the `--older-than` flag to nuke resources that were created before a certain period, the possible values are all valid values for [ParseDuration](https://golang.org/pkg/time/#ParseDuration) For example the following command nukes resources that are at least one day old:
```
brikis98: "to **only** nuke resources that were created before a certain period"

aws/asg_test.go
```diff
@@ -53,7 +54,14 @@ func TestListAutoScalingGroups(t *testing.T) {
 	// clean up after this test
 	defer nukeAllAutoScalingGroups(session, []*string{&groupName})
 
-	groupNames, err := getAllAutoScalingGroups(session, region)
+	groupNames, err := getAllAutoScalingGroups(session, region, time.Now().Add(1*time.Hour*-1))
```
brikis98: Remove the `-1` and use `time.Now().Sub()`.

commands/cli.go
```diff
@@ -27,12 +28,30 @@ func CreateCli(version string) *cli.App {
 			Name:  "exclude-region",
 			Usage: "regions to exclude",
 		},
+		cli.StringFlag{
+			Name:  "older-than",
+			Usage: "duration of resources' creation date to delete from",
```
brikis98: A bit hard to understand. Perhaps "Only delete resources older than this specified value. Can be any valid Go duration, such as 1d or 8h."

commands/cli.go
```diff
@@ -27,12 +28,30 @@ func CreateCli(version string) *cli.App {
...
+func parseDurationParam(paramValue string) (*time.Time, error) {
+	duration, err := time.ParseDuration(paramValue)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	// make it negative so it goes back in time
+	duration = -1 * duration
+
+	excludeAfter := time.Now().Add(duration)
```
brikis98: No need to use negatives. Just do `time.Now().Sub(...)`.

aws/asg_test.go
brikis98: This applies to all the code below as well.

aws/ec2.go
```diff
@@ -27,7 +29,9 @@ func filterOutProtectedInstances(svc *ec2.EC2, output *ec2.DescribeInstancesOutp
 			protected := *attr.DisableApiTermination.Value
 			// Exclude protected EC2 instances
 			if !protected {
-				filteredIds = append(filteredIds, &instanceID)
+				if excludeAfter.After(*instance.LaunchTime) {
```
brikis98: I believe this code originally had a bug in it that Josh caught. Is it possible to add a test case that would've caught that?

commands/cli.go
tonerdo: The thing is `Sub` is used to subtract a time from another and return a duration. I'd still have to do some sort of preparation and won't be able to use `duration` directly

brikis98: Oh, how weird that `Add` and `Sub` aren't symmetric! OK, carry on then :)

aws/ec2.go
tonerdo: Tests added

brikis98:
Don't forget to clean up the branch and cut a new release :)

PR #11, opened 2018-02-20
Ignore EBS volumes attached to protected instances

tonerdo:


aws/ebs.go
```diff
@@ -47,16 +47,19 @@ func nukeAllEbsVolumes(session *session.Session, volumeIds []*string) error {
 
 		_, err := svc.DeleteVolume(params)
 		if err != nil {
-			// Ignore not found errors, some volumes are deleted along with EC2 Instances
-			if !strings.Contains(err.Error(), "InvalidVolume.NotFound") {
-				logging.Logger.Errorf("[Failed] %s", err)
-				return errors.WithStackTrace(err)
+			if awsErr, isAwsErr := err.(awserr.Error); isAwsErr && awsErr.Code() == "VolumeInUse" {
```
brikis98: This error message could mean other things, so I'm not sure assuming it's a protected instance is the right move here. 

Shouldn't we instead update the `getAllEbsVolumes` to filter out volumes that are attached to instances? In general, `aws-nuke` should have terminated all instances *before* deleting volumes, so the only volumes still attached are probably those that we aren't supposed to delete anyway...

aws/ebs_test.go
```diff
@@ -108,7 +108,7 @@ func TestNukeEBSVolumes(t *testing.T) {
 	}
 
 	uniqueTestID := "aws-nuke-test-" + util.UniqueID()
-	createTestEC2Instance(t, session, uniqueTestID, false)
+	createTestEBSVolume(t, session, uniqueTestID)
```
brikis98: Why was the test passing if it wasn't doing the right thing? 

One of the most important ideas from TDD is that you should have your tests fail *before* you implement the feature. This ensures the tests are actually checking what you think they are and have clear errors when they fail.

aws/ebs.go
tonerdo: The would be a good approach if `getAllEbsVolumes` wasn't called _before_ any resources are deleted. The `getAll*` functions are used to populate the list the user sees before they confirm a nuke. The result of those functions are also reused for nuking so we don't have to make a duplicate api call. One way I could go is use check if the volume has any `VolumeAttachments` and going ahead to delete if it doesn't

aws/ebs_test.go
tonerdo: Definitely a lapse of judgement on my part here. It never actually created the EBS volume so when it called nuke and checked that the volume doesn't exist (which it won't) the tests pass. Will update this test shortly to be a lot more solid and will also do the same for other tests.

aws/ebs.go
brikis98: Ah, gotcha. Yea, I think just before deleting each volume, you'll have to check if it's attached, and skip it (while logging a warning) if it is.

tonerdo: I'm curious about something. `VolumeInUse` is thrown when a volume is still in use (whether or not it's a protected EC2 instance using it). If we're only checking for EC2 instances it means we could miss other scenarios that prevents a volume from being deleted which will again cause aws-nuke to fail. Granted my error message was a bit too specific, but my thinking is that it's better to catch all cases when it might be in use and update the log message to something a bit more generic e.g. "EBS volume vol-061bc7f7561ab1934 is attached to an active resource"

brikis98: That's a fair point. I think logging a warning with that more generic error message is good enough!

tonerdo: Log message has been updated

aws/ebs.go
```diff
@@ -47,16 +47,19 @@ func nukeAllEbsVolumes(session *session.Session, volumeIds []*string) error {
...
+				logging.Logger.Infof("EBS volume %s is attached to an active resource", *volumeID)
```
brikis98: Please make this a `Warnf` and say explicitly the volume can't be deleted.

aws/ebs_test.go
```diff
@@ -117,16 +117,74 @@ func TestNukeEBSVolumes(t *testing.T) {
...
-		assert.Fail(t, "Unable to fetch list of EC2 Instances")
+		assert.Fail(t, "Unable to fetch list of EBS Volumes")
 	}
 
-	for _, volumeID := range volumeIds {
-		assert.NotContains(t, volumes, *volumeID)
+	assert.NotContains(t, awsgo.StringValueSlice(volumeIds), awsgo.StringValue(volume.VolumeId))
+}
+
+func TestNukeEBSVolumesInUse(t *testing.T) {
```
brikis98: Good idea to add this test!

aws/ec2_test.go
```diff
@@ -93,6 +93,16 @@ func createTestEC2Instance(t *testing.T, session *session.Session, name string,
 	return *runResult.Instances[0]
 }
 
+func removeEC2InstanceProtection(svc *ec2.EC2, instance *ec2.Instance) {
+	// make instance unprotected so it can be cleaned up
+	svc.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
```
brikis98: I'm guessing this method can return an error. You should check for it and fail the test if it does.

brikis98:
OK, great, changes LGTM! Once the tests pass, feel free to merge and cut a new release.

PR #12, opened 2018-02-20
update circleci config to run aws-nuke nightly

tonerdo:


circle.yml
```diff
@@ -70,7 +78,12 @@ workflows:
           filters:
             tags:
               only: /^v.*/
-
+      - run:
+          requires:
+            - test
+          filters:
+            tags:
+              only: /^v.*/
```
brikis98: This will run `aws-nuke` every time we do a release. We probably only want to do it nightly, right?

circle.yml
```diff
@@ -42,6 +42,14 @@ jobs:
           root: .
           paths: bin
 
+  run:
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: /go/src/github.com/gruntwork-io/aws-nuke
+      - run: go run main.go
```
brikis98: Shouldn't this exclude `us-west-2`? And only delete resources older than, say, 2 days?

circle.yml
```diff
@@ -42,6 +42,14 @@ jobs:
           root: .
           paths: bin
 
+  run:
```
brikis98: Call this something more descriptive, like `nuke_phx_devops`.

circle.yml
tonerdo: I think 2 days is fair, although I'm interested to know how you arrived at that figure because I was gonna go with 24hrs. What's the frequency at which resources are typically created in the Phx DvOps account?

circle.yml
tonerdo: Thanks for catching this, left over from a copy paste

circle.yml
brikis98: 2 days is just a guess. We may have to tweak it later. Resources that are left from tests would ideally be cleaned up nightly. However, we also use Phx DevOps for manual testing, and occasionally, we leave things running a bit longer.

tonerdo:
@brikis98 the `nuke_phx_devops` job fails because circleci isn't equipped to provide input via stdin. I was thinking that we could add another cli flag that skips the prompt all together. @josh-padnick also thinks the same and even suggested `--force`. Thoughts?

circle.yml
```diff
@@ -70,7 +78,9 @@ workflows:
           filters:
             tags:
               only: /^v.*/
-
+      - nuke_phx_devops:
+          requires:
+            - test
```
brikis98: This will run aws-nuke after every commit... to delete resources more than 48h old. What's the purpose of that? I don't think we need this here; we should *only* have the `nuke_phx_devops` run once nightly.

brikis98:
> @brikis98 the nuke_phx_devops job fails because circleci isn't equipped to provide input via stdin. I was thinking that we could add another cli flag that skips the prompt all together. @josh-padnick also thinks the same and even suggested --force. Thoughts?

Yes, the tool definitely need a non-interactive mode. A `--force` flag to disable all interactive prompts sounds like a good idea.

commands/cli.go
```diff
@@ -33,6 +33,10 @@ func CreateCli(version string) *cli.App {
 			Usage: "Only delete resources older than this specified value. Can be any valid Go duration, such as 10m or 8h.",
 			Value: "0s",
 		},
+		cli.BoolFlag{
+			Name:  "force",
+			Usage: "Skip nuke confirmation prompt",
```
brikis98: You should put a big "WARNING: this will automatically delete all resources without any confirmation" here :)

commands/cli.go
```diff
@@ -93,19 +97,27 @@ func awsNuke(c *cli.Context) error {
...
 
-	if strings.ToLower(input) == "nuke" {
-		aws.NukeAllResources(account, regions)
+		if strings.ToLower(input) == "nuke" {
+			if err := aws.NukeAllResources(account, regions); err != nil {
+				return err
+			}
+		}
+	} else {
+		if err := aws.NukeAllResources(account, regions); err != nil {
```
brikis98: Add a log statement here that says "The --force flag is set, so proceeding without confirmation."

brikis98:
Make sure to delete the branch and cut a new release when you're ready.

Also, please give `aws-nuke` a manual run with the `--older-than 48h --exclude-region us-west-2` params, as there are a bunch of resources that need clean up now.

Finally, do one more experiment: run the tests and make sure that all the resources they create are cleaned up by the tests themselves. There are a lot of aws-nuke resources in the Phx DevOps account, so I suspect some test is failing to clean up after itself. The nightly run will eventually take care of that, but only after 2 days, so it would be better if the tests did the cleanup properly and aws-nuke only picked up the slack on case of (hopefully) rare bugs/crashes/etc.

commands/cli.go
josh-padnick: One other safety measure to consider: When `aws-nuke` is run with `--force`, show the warning but sleep for, say, 5 - 10 seconds so that an interactive user has an opportunity to abort the run.

brikis98: A 10 second "oh shit" pause is a very good idea. https://github.com/gruntwork-io/aws-nuke/issues/14

PR #15, opened 2018-02-23
Ensure tests cleanup after themselves

### LINKED ISSUE #13: EC2 Instances from Automated Tests Are Left Around
I'm pretty consistently seeing EC2 Instances in every region with names like:

* `aws-nuke-test-3nOZAN`
* `aws-nuke-test-iBLgDR`

It looks like `aws-nuke-test-3nOZAN` was launched at February 20, 2018 at 7:49:30 AM UTC-7, so I'm pretty confident that aws-nuke tests aren't cleaning up after themselves. Either that, or tests are continually failing, but it [doesn't look like that's the case](https://circleci.com/gh/gruntwork-io/aws-nuke), so the test clean up is probably the culprit.

To see for yourself, check out the PhxDevOps AWS account in any region.

tonerdo:
Fixes #13

### DIFF
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -25,7 +25,7 @@ func createTestAutoScalingGroup(t *testing.T, session *session.Session, name str
 
 	_, err := svc.CreateAutoScalingGroup(param)
 	if err != nil {
-		assert.Fail(t, "Could not create test ASG: %v", err)
+		assert.Failf(t, "Could not create test ASG", errors.WithStackTrace(err).Error())
 	}
 
 	err = svc.WaitUntilGroupExists(&autoscaling.DescribeAutoScalingGroupsInput{
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -25,7 +25,7 @@ func createTestAutoScalingGroup(t *testing.T, session *session.Session, name str
 
 	_, err := svc.CreateAutoScalingGroup(param)
 	if err != nil {
-		assert.Failf(t, "Could not create test ASG: %s", errors.WithStackTrace(err).Error())
+		assert.Fail(t, "Could not create test ASG: %v", err)
 	}
 
 	err = svc.WaitUntilGroupExists(&autoscaling.DescribeAutoScalingGroupsInput{
--- a/aws/asg_test.go
+++ b/aws/asg_test.go
@@ -49,25 +49,25 @@ func TestListAutoScalingGroups(t *testing.T) {
 		assert.Fail(t, errors.WithStackTrace(err).Error())
 	}
 
-	groupName := "aws-nuke-test-" + util.UniqueID()
-	createTestAutoScalingGroup(t, session, groupName)
+	uniqueTestID := "aws-nuke-test-" + util.UniqueID()
+	createTestAutoScalingGroup(t, session, uniqueTestID)
 	// clean up after this test
-	defer nukeAllAutoScalingGroups(session, []*string{&groupName})
-	defer nukeAllEc2Instances(session, findEC2InstancesByNameTag(t, session, groupName))
+	defer nukeAllAutoScalingGroups(session, []*string{&uniqueTestID})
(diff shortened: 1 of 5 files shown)


aws/asg_test.go
```diff
@@ -85,6 +86,9 @@ func TestNukeAutoScalingGroups(t *testing.T) {
 	groupName := "aws-nuke-test-" + util.UniqueID()
 	createTestAutoScalingGroup(t, session, groupName)
 
+	// clean up ec2 instance created by the above call
+	defer nukeAllEc2Instances(session, findEC2InstancesByNameTag(t, session, groupName))
```
brikis98: Should you also call `nukeAllAutoScalingGroups`?

aws/elb_test.go
```diff
@@ -54,14 +54,14 @@ func TestListELBs(t *testing.T) {
 
 	elbNames, err := getAllElbInstances(session, region, time.Now().Add(1*time.Hour*-1))
 	if err != nil {
-		assert.Fail(t, "Unable to fetch list of Auto Scaling Groups")
+		assert.Fail(t, "Unable to fetch list of ELBs")
```
brikis98: In all of these `assert.Fail`, you should include the error so we can tell what went wrong.

```go
assert.Fail(t, "Unable to fetch list of ELBs: %v", err)
```

aws/elbv2_test.go
```diff
@@ -25,20 +25,36 @@ func createTestELBv2(t *testing.T, session *session.Session, name string) elbv2.
 		assert.Fail(t, "Needs at least 2 subnets to create ELBv2")
 	}
 
-	subnet1 := *subnetOutput.Subnets[0]
-	subnet2 := *subnetOutput.Subnets[1]
+	var firstSubnet ec2.Subnet
+	var secondSubnet ec2.Subnet
+
+	// ensure we select subnets in different availability zones
```
brikis98: Any time you find yourself marking of blocks of code with a comment, you're typically better off extracting that code into a method with a descriptive name, such as `getSubnetsInDifferentAZs(...)`.

aws/elbv2_test.go
```diff
@@ -25,20 +25,36 @@ func createTestELBv2(t *testing.T, session *session.Session, name string) elbv2.
...
+	for _, subnet1 := range subnetOutput.Subnets {
+		for _, subnet2 := range subnetOutput.Subnets {
+			if subnet1.AvailabilityZone != subnet2.AvailabilityZone && subnet1.SubnetId != subnet2.SubnetId {
```
brikis98: If `ec2.Subnet` is a struct, then all of these are empty strings initially, aren't they? And if it's a pointer, won't this cause a `nil` pointer dereference?

aws/elbv2_test.go
```diff
@@ -25,20 +25,36 @@ func createTestELBv2(t *testing.T, session *session.Session, name string) elbv2.
...
+				firstSubnet = *subnet1
+				secondSubnet = *subnet2
+				break
+			}
+
+			if subnet1 != nil && subnet2 != nil {
+				break
+			}
```
brikis98: I find the logic above a bit confusing. Couldn't you just set `subnet1 = *subnetOutput.Subnets[0]` and then just loop over `*subnetOutput.Subnets[1 ... n]` until you find one in a different AZ? Also, you have a check above for at least 2 subnets, but if they are in the same AZ, you may need more, so you probably need to check after this loop that two subnets were actually found!

aws/asg_test.go
tonerdo: Oh no, I'm deleting the ec2 instance created before the autoscaling group could be created. I'll update the variable name so it's less confusing

aws/elbv2_test.go
tonerdo: Yeah, good catch. I should be checking that the pointers aren't nil and then failing the tests if they are

PR #17, opened 2018-03-13
Sleep for 10 seconds when force flag is set

tonerdo:
Fixes #17

### DIFF
--- a/commands/cli.go
+++ b/commands/cli.go
@@ -116,9 +116,9 @@ func awsNuke(c *cli.Context) error {
 			}
 		}
 	} else {
-		logging.Logger.Infoln("The --force flag is set, so waiting for 10 seconds before proceeding to nuke everything your account. If you don't want to proceed, hit CTRL+C now!!")
-		for i := 0; i < 10; i++ {
-			fmt.Printf("%d...", i+1)
+		logging.Logger.Infoln("The --force flag is set, so waiting for 10 seconds before proceeding to nuke everything in your account. If you don't want to proceed, hit CTRL+C now!!")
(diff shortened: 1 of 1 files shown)


commands/cli.go
```diff
@@ -115,7 +115,8 @@ func awsNuke(c *cli.Context) error {
 			}
 		}
 	} else {
-		logging.Logger.Infoln("The --force flag is set, so proceeding without confirmation.")
+		logging.Logger.Infoln("The --force flag is set, so proceeding after 10 seconds without confirmation.")
```
brikis98: I'd update this to say: "The --force flag is set, so waiting for 10 seconds before proceeding to nuke everything your account. If you don't want to proceed, hit CTRL+C now!!"

commands/cli.go
```diff
@@ -115,7 +116,13 @@ func awsNuke(c *cli.Context) error {
 			}
 		}
 	} else {
-		logging.Logger.Infoln("The --force flag is set, so proceeding without confirmation.")
+		logging.Logger.Infoln("The --force flag is set, so waiting for 10 seconds before proceeding to nuke everything your account. If you don't want to proceed, hit CTRL+C now!!")
```
brikis98: "everything **in** your account"

commands/cli.go
```diff
@@ -115,7 +116,13 @@ func awsNuke(c *cli.Context) error {
...
+		for i := 0; i < 10; i++ {
+			fmt.Printf("%d...", i+1)
+			time.Sleep(1 * time.Second)
```
brikis98: I feel like a countdown from 10 down to 1 would make more sense here...

PR #18, opened 2018-03-14
Nuke User owned AMIs

tonerdo:
This PR does the following:

- Adds support for listing and nuking Amazon Machine Images owned by the user
- Adds tests for this functionality

### DIFF
--- a/aws/ami.go
+++ b/aws/ami.go
@@ -6,6 +6,7 @@ import (
 	awsgo "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/aws-nuke/logging"
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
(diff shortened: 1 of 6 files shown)


aws/ami.go
```diff
@@ -0,0 +1,69 @@
...
+
+	logging.Logger.Infof("Terminating all AMIs in region %s", *session.Config.Region)
+
+	for _, imageID := range imageIds {
+		params := &ec2.DeregisterImageInput{
+			ImageId: imageID,
+		}
+
+		_, err := svc.DeregisterImage(params)
+		if err != nil {
```
brikis98: If you try to delete an AMI that's actively in use, do you get an error? If so, we should catch it, log it, and keep going.

aws/ami_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/aws-nuke/util"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func createTestAMI(t *testing.T, session *session.Session, name string) ec2.Image {
+	svc := ec2.New(session)
+	instance := createTestEC2Instance(t, session, name, false)
```
brikis98: Should this `defer` the deletion of this instance?

tonerdo: It does the defer on line 62. If the `defer` is done here, the EC2 instance could have been nuked before we get a chance to create the Image. The safest bet is to just clean up after the test function itself

aws/ami.go
tonerdo: Will try this out

tonerdo: From my tests an AMI can be deleted even when there exists an EC2 instance that was created with it. I believe once the VM is provisioned with that machine image there's no link between them and that ami is exists just for re-provisioning new EC2 instances

brikis98: Good to know!

Ship it 👍

tonerdo: Sweet, I'll merge once I fix the test

josh-padnick: @tonerdo Is deleting an AMI actually a two-step procedure as [I had to do](https://github.com/josh-padnick/ec2-snapper/blob/master/delete_command.go#L258) in ec2-snapper? At that time, I first had to deregister the AMI (as you do here), and second, had to delete the AMI's corresponding EBS snapshot.

PR #22, opened 2018-03-20
Add potential support other cloud providers

tonerdo:


commands/cli.go
```diff
@@ -58,6 +58,29 @@ func parseDurationParam(paramValue string) (*time.Time, error) {
...
+				}
+			case "gcp":
+				return UnsupportedProviderError{
+					Name: "aws",
+				}
+			}
+		}
+	}
+
+	return UnsupportedProviderError{}
```
brikis98: If there are no args, you should prob just show the help text (`urfave/cli` has some method to do that).

PR #23, opened 2018-03-21
Restructure cli, make provider names subcommands

tonerdo:


commands/cli.go
```diff
@@ -23,24 +23,30 @@ func CreateCli(version string) *cli.App {
...
-		cli.StringFlag{
-			Name:  "older-than",
-			Usage: "Only delete resources older than this specified value. Can be any valid Go duration, such as 10m or 8h.",
-			Value: "0s",
-		},
-		cli.BoolFlag{
-			Name:  "force",
-			Usage: "Skip nuke confirmation prompt. WARNING: this will automatically delete all resources without any confirmation",
+	app.Usage = "A CLI tool to cleanup cloud resources (AWS, Azure, GCP). THIS TOOL WILL COMPLETELY REMOVE ALL RESOURCES AND ITS EFFECTS ARE IRREVERSIBLE!!!"
+	app.Commands = []cli.Command{
```
brikis98: Does `cloud-nuke` (without any args) still show the help text?

commands/cli.go
```diff
@@ -23,24 +23,30 @@ func CreateCli(version string) *cli.App {
...
+				},
+				cli.StringFlag{
+					Name:  "older-than",
+					Usage: "Only delete resources older than this specified value. Can be any valid Go duration, such as 10m or 8h.",
+					Value: "0s",
+				},
+				cli.BoolFlag{
+					Name:  "force",
+					Usage: "Skip nuke confirmation prompt. WARNING: this will automatically delete all resources without any confirmation",
+				},
```
brikis98: If you're worried about having to define the same args for all future cloud providers, you can extract each one into its own var and then pass those vars in as a list:

```go
excludeRegionFlag := cli.StringSliceFlag{ ... }
olderThanFlag := cli.StringFlag{ ... }
allFlags := []cli.Flag{excludeRegionFlag, olderThanFlag}
```

commands/cli.go
tonerdo: Yeah it does

brikis98: OK, great

commands/cli.go
tonerdo: I think I'll keep it as it is for now, it seems cleaner. In the near future when I add deleting specific resources e.g. `cloud-nuke aws ami` chances are things might get hairy especially if we see that some options could be unique to a specific provider

brikis98: Fine to keep it as-is for now. But remember, these are all variables, so you can choose which of them to pass to `aws`, which to `gcp`, etc. You can combine and compose however you want :)

PR #24, opened 2018-03-22
implement custom waituntilavailable method for AMIs

tonerdo:


aws/ami_test.go
```diff
@@ -7,11 +7,31 @@ import (
...
+		if *output.Images[0].State == "available" {
+			return nil
+		}
+
+		time.Sleep(5 * time.Second)
+		logging.Logger.Debug("Waiting for ELB to be available")
+	}
+
+	// return ElbDeleteError{}
+	return nil
```
brikis98: Wait, shouldn't this return an error, as you weren't able to find the image you wanted even after 50 retries?

aws/ami_test.go
```diff
@@ -7,11 +7,31 @@ import (
...
+		if err != nil {
+			return err
+		}
+
+		if *output.Images[0].State == "available" {
+			return nil
+		}
+
+		time.Sleep(5 * time.Second)
+		logging.Logger.Debug("Waiting for ELB to be available")
```
brikis98: I would put the logging statement first, so it's clear what's going on.

aws/ami_test.go
tonerdo: This was a complete oversight. Thanks for the call out

PR #25, opened 2018-03-27
Support deleting in batch requests to avoid AWS API request limits

### LINKED ISSUE #21: RequestLimitExceeded error when resources are a lot

tonerdo:
Fixes #21

### DIFF
--- a/aws/ami_types.go
+++ b/aws/ami_types.go
@@ -22,6 +22,7 @@ func (image AMIs) ResourceIdentifiers() []string {
 }
 
 func (image AMIs) MaxBatchSize() int {
+	// Tentative batch size to ensure AWS doesn't throttle
 	return 200
 }
 
--- a/aws/ami_types.go
+++ b/aws/ami_types.go
@@ -21,6 +21,10 @@ func (image AMIs) ResourceIdentifiers() []string {
 	return image.ImageIds
 }
 
+func (image AMIs) MaxBatchSize() int {
+	return 200
+}
+
 // Nuke - nuke 'em all!!!
 func (image AMIs) Nuke(session *session.Session, identifiers []string) error {
 	if err := nukeAllAMIs(session, awsgo.StringSlice(identifiers)); err != nil {
--- a/aws/ami_types.go
+++ b/aws/ami_types.go
@@ -22,16 +22,7 @@ func (image AMIs) ResourceIdentifiers() []string {
 }
 
 // Nuke - nuke 'em all!!!
-func (image AMIs) Nuke(session *session.Session) error {
-	if err := nukeAllAMIs(session, awsgo.StringSlice(image.ImageIds)); err != nil {
-		return errors.WithStackTrace(err)
-	}
-
-	return nil
-}
-
-// NukeBatch - nuke some!!!
-func (image AMIs) NukeBatch(session *session.Session, identifiers []string) error {
+func (image AMIs) Nuke(session *session.Session, identifiers []string) error {
 	if err := nukeAllAMIs(session, awsgo.StringSlice(identifiers)); err != nil {
 		return errors.WithStackTrace(err)
 	}
--- a/aws/ami_types.go
+++ b/aws/ami_types.go
@@ -30,6 +30,15 @@ func (image AMIs) Nuke(session *session.Session) error {
 	return nil
 }
 
+// NukeBatch - nuke some!!!
+func (image AMIs) NukeBatch(session *session.Session, identifiers []string) error {
+	if err := nukeAllAMIs(session, awsgo.StringSlice(identifiers)); err != nil {
+		return errors.WithStackTrace(err)
+	}
+
+	return nil
+}
+
 type ImageAvailableError struct{}
 
 func (e ImageAvailableError) Error() string {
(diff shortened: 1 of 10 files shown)


aws/aws.go
```diff
@@ -177,8 +191,21 @@ func NukeAllResources(account *AwsAccountResources, regions []string) error {
 
 		resourcesInRegion := account.Resources[region]
 		for _, resources := range resourcesInRegion.Resources {
-			if err := resources.Nuke(session); err != nil {
-				return errors.WithStackTrace(err)
+			if len(resources.ResourceIdentifiers()) >= 500 {
+				logging.Logger.Infoln("Terminating Resources in batches")
+				batches := split(resources.ResourceIdentifiers(), 200)
```
brikis98: Where does `500` and `200` come from? These are "magic" numbers and should be extracted into constants with some explanation above them (e.g., AWS throttles after XXX API calls...).

aws/aws.go
```diff
@@ -177,8 +191,21 @@ func NukeAllResources(account *AwsAccountResources, regions []string) error {
 
 		resourcesInRegion := account.Resources[region]
 		for _, resources := range resourcesInRegion.Resources {
-			if err := resources.Nuke(session); err != nil {
-				return errors.WithStackTrace(err)
+			if len(resources.ResourceIdentifiers()) >= 500 {
+				logging.Logger.Infoln("Terminating Resources in batches")
```
brikis98: "There are more than XXX resources to terminate (YYY total), so terminating resources in batches"

aws/aws.go
brikis98: In fact, each resource type should probably have a `MaxBatchSize()` function.

aws/aws.go
```diff
@@ -177,8 +191,21 @@ func NukeAllResources(account *AwsAccountResources, regions []string) error {
...
+				batches := split(resources.ResourceIdentifiers(), 200)
+
+				for _, batch := range batches {
+					if err := resources.NukeBatch(session, batch); err != nil {
```
brikis98: It feels a bit weird to have a nearly identical `NukeBatch` and `Nuke` method in each resource type. Perhaps there should solely be a `Nuke` method that takes a list of identifiers as a parameter?

aws/aws.go
tonerdo: Sorry about the "magic" numbers. Was trying stuff out and forgot to abstract it.

brikis98: AWS is vague on what the limits are: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html. However, they vary per service and even per region. See also https://stackoverflow.com/questions/23728591/ec2-api-throttling-limits. 

I suspect we will need the flexibility to use different limits with different services. Also, it may turn out simple batching doesn't work and we need an [exponential back off](https://docs.aws.amazon.com/general/latest/gr/api-retries.html), but let's try the simple version first.

tonerdo: Ok, Gotcha. Thanks for the links

aws/ami_types.go
```diff
@@ -21,9 +21,13 @@ func (image AMIs) ResourceIdentifiers() []string {
 	return image.ImageIds
 }
 
+func (image AMIs) MaxBatchSize() int {
+	return 200
```
brikis98: Please put a comment that this is a best guess!

brikis98: Same for all the other impls of this method

aws/aws.go
```diff
@@ -41,6 +41,20 @@ func getRandomRegion() string {
 	return allRegions[randIndex]
 }
 
+func split(identifiers []string, limit int) [][]string {
```
brikis98: Please add a bunch of unit tests for this method to check boundary conditions: empty list, 1 item list, odd number of items, even number of items, small limit, large limit, etc.

aws/aws.go
```diff
@@ -166,6 +180,8 @@ func GetAllResources(regions []string, excludedRegions []string, excludeAfter ti
 
 // NukeAllResources - Nukes all aws resources
 func NukeAllResources(account *AwsAccountResources, regions []string) error {
+	// Resources more than 500 in number could cause AWS to throttle api calls
```
brikis98: Again, please mention that 500 is a rule of thumb based on experience, but the actual number will vary based on resource type, region, etc.

aws/aws.go
```diff
@@ -177,8 +193,23 @@ func NukeAllResources(account *AwsAccountResources, regions []string) error {
 
 		resourcesInRegion := account.Resources[region]
 		for _, resources := range resourcesInRegion.Resources {
-			if err := resources.Nuke(session); err != nil {
-				return errors.WithStackTrace(err)
+			length := len(resources.ResourceIdentifiers())
+			if length >= MaxRequestSize {
```
brikis98: Actually, it doesn't seem like you should even do this check. Instead, always batch things. If there are fewer items than the `MaxBatchSize`, it will all end up in one batch anyway.

tonerdo: I'll remove this and up the batch size of all resources to 500

PR #27, opened 2018-04-04
Use string comparison to determine RequestLimitExceeded error

tonerdo:


aws/aws.go
```diff
@@ -207,18 +207,19 @@ func NukeAllResources(account *AwsAccountResources, regions []string) error {
 			for i := 0; i < len(batches); i++ {
 				batch := batches[i]
 				if err := resources.Nuke(session, batch); err != nil {
-					if awsErr, isAwsErr := err.(awserr.Error); isAwsErr && awsErr.Code() == "RequestLimitExceeded" {
+					if strings.Contains(err.Error(), "RequestLimitExceeded") {
```
brikis98: What was the error type if not `awserr.Error`? You can find out with `reflect.TypeOf(...)`.

tonerdo: Definitely didn't consider using reflection. The only challenge here is that now that our resources are all cleared, I won't hit this error soon enough to be able to get the type. Might have to resort to a bit of research to figure it out

brikis98: I ask because I prefer to understand what caused the bug in the first place. Most errors I've seen from the AWS SDK are `awserr.Error`. If this one isn't, it seems valuable to know why. I'm OK merging this fix as-is with a TODO while we figure it out, but I suspect if we don't understand it fully, this bug will bite us again in the future.

PR #28, opened 2018-04-06
Adds Elastic IP nuking support

tonerdo:


aws/eip.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	logging.Logger.Infof("Deleting all Elastic IPs in region %s", *session.Config.Region)
+
+	for _, allocationID := range allocationIds {
+		params := &ec2.ReleaseAddressInput{
+			AllocationId: allocationID,
+		}
+
+		_, err := svc.ReleaseAddress(params)
```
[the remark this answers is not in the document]
tonerdo: This doesn't prevent the deletion of the the Elastic IP

aws/eip.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// Returns a formatted string of EIP allocation ids
+func getAllEIPAddresses(session *session.Session, region string, excludeAfter time.Time) ([]*string, error) {
```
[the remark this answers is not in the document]
tonerdo: Weirdly enough, this is the one resource that doesn't have a created time info attached to it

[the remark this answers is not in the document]
brikis98: That's... quite a problem. That means if we were to delete these, cloud-nuke could delete EIPs that are currently in use in automated tests, manual testing, etc.

I'm trying to think of workarounds...

1. We could  make `cloud-nuke` stateful. Each time it runs, it records what EIPs it saw and a timestamp of "first seen" for that EIP in DynamoDB. Next time it runs, it finds the previous entries and deletes anything it "first saw" longer ago than `excludeAfter`. This would work for any resource that doesn't track a creation date.

1. Same as above, except instead of using DynamoDB, aws-nuke adds a "first-seen" timestamp as a tag (`cloud-nuke-first-seen-timestamp`) on the EIP. This would work for any resource that doesn't have a Creation Date, but does have tags.

#2 seems fairly elegant to me...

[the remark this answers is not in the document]
tonerdo: Option (2) is definitely the better way to go. Having to use a DB brings in another point of failure

aws/aws.go
```diff
@@ -153,6 +153,19 @@ func GetAllResources(regions []string, excludedRegions []string, excludeAfter ti
...
+			return nil, errors.WithStackTrace(err)
+		}
+
+		eipAddresses := EIPAddresses{
+			AllocationIds: awsgo.StringValueSlice(allocationIds),
+		}
+
+		resourcesInRegion.Resources = append(resourcesInRegion.Resources, eipAddresses)
+		// End EIP Addresses
+
```
brikis98: This method is now 130+ lines long. That's too long. In a follow-up PR, consider the following:

1. Add a new method to `AwsResources` along the lines of `func GetAllResources(session, region, excludeAfter) []string, error`.
1. Implement that method in each resource using its existing `getAllXXX` method (e.g., `getAllEIPAddresses`).
1. At the top of this method, create `allResources []AwsResources{...}`, where the `...` is an instance of each type of resource we support.
1. This method now reduces down to a single loop over `allResources` that calls `GetAllResources` and adds them to a list.

aws/eip.go
```diff
@@ -0,0 +1,94 @@
...
+	"time"
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func getOrSetFirstSeenTag(svc *ec2.EC2, address ec2.Address) (time.Time, error) {
```
brikis98: I'm not a fan of a method that does two things. It's harder to reason about it and harder to test. If you break it up into two methods—`getFirstSeenTag` and `setFirstSeenTag`—it's clearer what each does and you can have unit tests for each.

aws/eip.go
```diff
@@ -0,0 +1,94 @@
...
+	layout := "2006-01-02 15:04:05"
+
+	tags := address.Tags
+	for _, tag := range tags {
+		if *tag.Key == key {
+			return time.Parse(layout, *tag.Value)
+		}
+	}
+
+	now := time.Now().UTC()
```
brikis98: Also, if you have a separate `setFirstSeenTag`, it can take a timestamp as a param, which will make it even easier to test that the method does the right thing, as you wouldn't be bound to the system clock.

aws/eip.go
```diff
@@ -0,0 +1,94 @@
...
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func getOrSetFirstSeenTag(svc *ec2.EC2, address ec2.Address) (time.Time, error) {
+	key := "cloud-nuke-first-seen"
+	layout := "2006-01-02 15:04:05"
```
brikis98: These should probably be `const` outside of this method.

aws/eip.go
brikis98: Also, please add a comment to this method explaining how and why we tag EIPs

aws/eip_test.go
```diff
@@ -0,0 +1,84 @@
...
+	})
+
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+
+	return *output.Addresses[0]
+}
+
+func TestListEIPAddress(t *testing.T) {
```
brikis98: Do you have a test that checks if the tagging actually works? E.g., a test that creates an EIP, waits 2 min, creates another, and sets `excludeAfter` to 1 min, and makes sure only one of the EIPs is returned? You also want to check that tagging works properly, so if that test waits another min, both should be returned.

aws/aws.go
tonerdo: #29 has been created to track this

aws/eip.go
tonerdo: I'll make them accept these values as arguments instead so the constants aren't global to the package

aws/eip.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return nil
+}
+
+func getFirstSeenTag(svc *ec2.EC2, address ec2.Address, key string, layout string) (*time.Time, error) {
+	tags := address.Tags
+	for _, tag := range tags {
+		if *tag.Key == key {
+			firstSeenTime, err := time.Parse(layout, *tag.Value)
+			return &firstSeenTime, err
```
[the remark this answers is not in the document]
tonerdo: > Will taking the address of it cause an error?

No, `time.Parse` will default to the Unix Epoch

> I think it communicates the intent better to do

Agreed

PR #31, opened 2018-05-14
update circle-ci config to nuke sandbox account nightly

tonerdo:


circle.yml
```diff
@@ -52,6 +52,16 @@ jobs:
...
 
+  nuke_sandbox:
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: /go/src/github.com/gruntwork-io/cloud-nuke
+      - run: export AWS_ACCESS_KEY_ID=SANDBOX_AWS_ACCESS_KEY_ID
+      - run: export AWS_SECRET_ACCESS_KEY=SANDBOX_AWS_SECRET_ACCESS_KEY
+      - run: go run main.go aws --older-than 1h --force
```
brikis98: Sandbox should only be older than 24h, right?

circle.yml
```diff
@@ -90,7 +100,7 @@ workflows:
               ignore: /.*/
 
 
-  nightly:
+  frequently:
```
brikis98: What's that do?

circle.yml
tonerdo: Ah thanks. Good catch

PR #34, opened 2018-10-04
Improve nuking strategy

tonerdo:
Currently when iterating over resource IDs and nuking each, cloud-nuke returns from the function when it encounters an error. This results in the other resources further down the list not being deleted on that pass.

This PR updates cloud-nuke to log when it encounters an error and continue with other resources in the list. This should greatly reduce the amount of leftover resources we see and should slightly reduce our monthly costs

### DIFF
--- a/aws/asg.go
+++ b/aws/asg.go
@@ -38,6 +38,7 @@ func nukeAllAutoScalingGroups(session *session.Session, groupNames []*string) er
 	}
 
 	logging.Logger.Infof("Deleting all Auto Scaling Groups in region %s", *session.Config.Region)
+	var deletedGroupNames []*string
 
 	for _, groupName := range groupNames {
 		params := &autoscaling.DeleteAutoScalingGroupInput{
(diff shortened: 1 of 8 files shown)


aws/asg.go
```diff
@@ -48,20 +49,20 @@ func nukeAllAutoScalingGroups(session *session.Session, groupNames []*string) er
...
-		logging.Logger.Infof("Deleted Auto Scaling Group: %s", *groupName)
 	}
 
 	err := svc.WaitUntilGroupNotExists(&autoscaling.DescribeAutoScalingGroupsInput{
-		AutoScalingGroupNames: groupNames,
+		AutoScalingGroupNames: deletedGroupNames,
 	})
 
 	if err != nil {
 		return errors.WithStackTrace(err)
```
brikis98: Should this also just log the error?

aws/ebs.go
```diff
@@ -49,27 +50,25 @@ func nukeAllEbsVolumes(session *session.Session, volumeIds []*string) error {
...
-		logging.Logger.Infof("Deleted EBS Volume: %s", *volumeID)
 	}
 
 	err := svc.WaitUntilVolumeDeleted(&ec2.DescribeVolumesInput{
-		VolumeIds: volumeIds,
+		VolumeIds: deletedVolumeIDs,
 	})
 
 	if err != nil {
 		return errors.WithStackTrace(err)
```
brikis98: Log the error?

aws/ec2.go
```diff
@@ -102,6 +98,10 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
 		},
 	})
 
+	for _, instanceID := range instanceIds {
+		logging.Logger.Infof("Terminated EC2 Instance: %s", *instanceID)
+	}
+
 	if err != nil {
 		logging.Logger.Errorf("[Failed] %s", err)
 		return errors.WithStackTrace(err)
```
brikis98: Log the error?

aws/elb.go
```diff
@@ -65,20 +66,20 @@ func nukeAllElbInstances(session *session.Session, names []*string) error {
...
-		logging.Logger.Infof("Deleted ELB: %s", *name)
 	}
 
 	err := waitUntilElbDeleted(svc, &elb.DescribeLoadBalancersInput{
-		LoadBalancerNames: names,
+		LoadBalancerNames: deletedNames,
 	})
 
 	if err != nil {
 		return errors.WithStackTrace(err)
```
brikis98: Log the error?

aws/elbv2.go
```diff
@@ -46,20 +47,20 @@ func nukeAllElbv2Instances(session *session.Session, arns []*string) error {
...
-		logging.Logger.Infof("Deleted ELBv2: %s", *arn)
 	}
 
 	err := svc.WaitUntilLoadBalancersDeleted(&elbv2.DescribeLoadBalancersInput{
-		LoadBalancerArns: arns,
+		LoadBalancerArns: deletedArns,
 	})
 
 	if err != nil {
 		return errors.WithStackTrace(err)
```
brikis98: Log the error?

Issue #32, opened 2018-06-12
ECS resources are not nuked


sarkis:
Resources like ECSCluster, Tasks, Services should all be nuked.

sarkis:
@brikis98 would love to implement this one. Won't commit until I have the time for it, though :)

tonerdo:
@sarkis, go for it! 👍

sarkis:
Just tested #36 and it works perfect. ECS Cluster deletion issue is a non-issue in terms of cloud-nuke and documented - marking this issue as closed. Thanks @yorinasub17!

PR #36, opened 2018-10-11
Nuke ECS services and ECS clusters

yorinasub17:
First attempt at addressing https://github.com/gruntwork-io/cloud-nuke/issues/32

This implements nuking of:
- ECS tasks (indirectly, by draining ECS services)
- ECS services
- ~ECS clusters~

This does NOT implement nuking of:
- ECS task definitions
- Target groups for LB
- Service discovery services and namespaces


aws/ecs_cluster.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// getAllEcsClusters - Returns a string of ECS Cluster ARNs, which uniquely identifies the cluster.
+// NOTE: AWS api doesn't provide the necessary information to
+//       implement `excludeAfter` filter at the cluster
+//       level, so we will implement it at the service level.
```
brikis98: Wait, so does that mean we delete all ECS clusters, no matter how old they are?

aws/ecs_service.go
```diff
@@ -0,0 +1,157 @@
...
+	return ecsServiceArns, ecsServiceClusterMap, nil
+}
+
+// Deletes all provided ECS Services. At a high level this involves two steps:
+// 1.) Drain all tasks from the service so that nothing is
+//     running.
+// 2.) Delete service object once no tasks are running.
+// Note that this will swallow failed deletes and continue along, logging the
+// service ARN so that we can find it later.
+func nukeAllEcsServices(awsSession *session.Session, ecsServiceClusterMap map[string]string, ecsServiceArns []*string) error {
```
brikis98: This is a really long function with big comment blocks in the middle. Consider breaking the contents into several smaller functions: e.g., `startDrainingEcsServices`, `waitForServicesToFinishDraining`, etc.

aws/ecs_cluster.go
brikis98: There are other resources that don't expose a "created date." For those, we use tags to only delete resources older than a certain date. E.g., see the EIP code: https://github.com/gruntwork-io/cloud-nuke/blob/master/aws/eip.go

yorinasub17: I was actually going to go that route originally, except in my original search I couldn't find any way to set tags in ECS. But I guess I didn't try hard enough then because now I found that they are called something else in ECS (`Attribute`), which explains why I didn't find it the first time: I was looking for the wrong thing :(

... except unfortunately, it looks like it won't even let you set attributes on a cluster:

```
houston exec sandbox -- aws --region us-east-1 ecs put-attributes --cluster "$CLUSTER_ARN" --attributes name="cloud-nuke-test",value="1234",targetId="$CLUSTER_ARN"

An error occurred (InvalidParameterException) when calling the PutAttributes operation: Unexpected targetType 'cluster'.
```

I can't find anything else in the ECS API suite that lets me set custom values on a resource. The only thing we have control of is the `Name`, which isn't something we should overload.

Should I just remove the cluster nuking for now?

aws/ecs_service.go
yorinasub17: Done

aws/ecs_cluster.go
brikis98: Yea, probably... We should be conservative in a tool like this and never delete more than expected.

yorinasub17: Make sense. Will take it out.

yorinasub17: Added a caveat section to the README to signal that we have considered this and made the conscious choice not to pursue this at the moment.

yorinasub17:
@brikis98 Is it ok to merge this? I have taken out ECS cluster deleting as requested. This now only contains deleting ECS services, which does have a timestamp we can use.

brikis98:
Yup, ship it! 👍

PR #43, opened 2018-11-30
Nuke EKS clusters

yorinasub17:
Given how expensive EKS is (one cluster = $0.20/hr), and how frequent the tests fail right now as I actively develop, I am afraid I might forget to clean up one of these so decided to go ahead and implement the nuking.

Note that this required upgrading `aws-sdk-go`. I tried the latest and there were issues, so downgraded to what we have in terratest.

### DIFF
--- a/aws/eks.go
+++ b/aws/eks.go
@@ -1,0 +1,95 @@
+package aws
+
+import (
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// getAllEksClusters returns a list of strings of EKS Cluster Names that uniquely identify each cluster.
+func getAllEksClusters(awsSession *session.Session, excludeAfter time.Time) ([]*string, error) {
+	svc := eks.New(awsSession)
+	result, err := svc.ListClusters(&eks.ListClustersInput{})
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+	filteredClusters, err := filterOutRecentEksClusters(svc, result.Clusters, excludeAfter)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+	return filteredClusters, nil
+}
+
+// filterOutRecentEksClusters will take in the list of clusters and filter out any clusters that were created after
+// `excludeAfter`.
+func filterOutRecentEksClusters(svc *eks.EKS, clusterNames []*string, excludeAfter time.Time) ([]*string, error) {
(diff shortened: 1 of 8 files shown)


aws/eks_utils_for_test.go
```diff
@@ -0,0 +1,106 @@
...
+func getRandomEksSupportedRegion() string {
+	supportedRegions := []string{
+		"us-east-1",
+		"us-east-2",
+		"us-west-2",
+		"eu-west-1",
+	}
+	rand.Seed(time.Now().UnixNano())
+	randIndex := rand.Intn(len(supportedRegions))
+	return supportedRegions[randIndex]
```
brikis98: There's a terratest `GetRandomRegion` helper that takes in a whitelist and/or blacklist of regions you can use...

aws/eks.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// getAllEksClusters returns a list of strings of EKS Cluster Names that uniquely identify each cluster.
+func getAllEksClusters(awsSession *session.Session, excludeAfter time.Time) ([]*string, error) {
```
brikis98: Don't you need to call this EKS code from somewhere? I think there's a struct of some sort you need to add and then some extra API calls in the main CLI app...

yorinasub17: Yup... I can't believe I missed that...

yorinasub17: I was so tied up in the tests and it was getting late that I got a bit too trigger happy I guess.

yorinasub17:
One downside of this: the tests now have an extra 20 minutes because launching and tearing down EKS clusters take that long...

brikis98:
> One downside of this: the tests now have an extra 20 minutes because launching and tearing down EKS clusters take that long...

Does it run in parallel with the other tests so the whole build takes 20 min? Or is it sequential?

yorinasub17:
It runs parallel, so the whole build takes 20 minutes. For comparison, it was ~5 mins before.

brikis98:
Understood. Not much we can do, TBH. Eventually, we'll want to nuke RDS, Elasticsearch, and other slow stuff too, so the build will only get slower. Such is DevOps.

rfvermut:
Can anything be done to finish this?

autero1:
We don't seem to have verbose output in the tests. Are there any particular tests that keep failing?

tonerdo:
@autero1 see #50

yorinasub17:
Thanks @tonerdo for the test fixes! I took a quick scan and I believe I addressed all the comments, so I am going to go ahead and merge this and release!

PR #44, opened 2018-11-30
Fix panic from segfaults caused by continuing from a failed condition

yorinasub17:
This replaces the `if err != nil { assert.Fail() }` calls with `require.NoError`. Aside from being slightly more readable, the switch to using `require` as opposed to `assert` allows the test to exit when it shouldn't continue. By doing so, we can avoid panics halting the test and leaving resources lying around like in [this failure](https://circleci.com/gh/gruntwork-io/cloud-nuke/5857).


yorinasub17:
Yup! Jim pointed it out to me on one of the PRs and have since been using it religiously :)

yorinasub17:
Merging!

Issue #42, opened 2018-11-28
cloud-nuke gcp


acrisci:
Hi,

I'm interested in using cloud-nuke for google cloud platform. I see it on the roadmap in the README. Are you accepting contributions for this feature?

josh-padnick:
We'd love contributions for GCP or any other cloud! We use the name `cloud-nuke` because we wantd to add other clouds, but we just haven't gotten around to it yet.

wyardley:
Might make some sense to look at doing this via Magic Modules:
https://github.com/GoogleCloudPlatform/magic-modules
since a lot of the scaffolding (and information needed) is already there.

wyardley:
@josh-padnick I can take a look at what’s involved.

sjmiller609:
Also relevant - a gcloud CLI feature to list all resources. https://cloud.google.com/sdk/gcloud/reference/alpha/resources/list

koalalorenzo:
Is anybody looking/working on this?

acrisci:
I'm not working on it. Look at #45.

arehmandev:
I wrote something rather quick in case anyone wants a look: https://github.com/arehmandev/gcp-nuke - happy for any contributors

hongil0316:
Closing this issue as there seems to be other open source tools that tackle this problem and there are no immediate plans to implement this feature in cloud-nuke

PR #45, opened 2018-12-02
Add support for GCP compute instances

acrisci:
This lays the groundwork for general GCP support for cloud-nuke with
support for deleting compute instances.

Add the `gcp` package and CLI argument to nuke GCP resources.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -47,3 +47,11 @@
 [[constraint]]
   name = "github.com/fatih/color"
   version = "1.5.0"
+
+[[constraint]]
+  branch = "master"
+  name = "google.golang.org/api"
+
+[[constraint]]
+  branch = "master"
+  name = "golang.org/x/net"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -52,13 +52,13 @@
   branch = "master"
   name = "google.golang.org/api"
 
+[[override]]
+  name = "golang.org/x/sys"
+  revision = "1c9583448a9c3aa0f9a6a5241bf73c0bd8aafded"
+
 [[constraint]]
   branch = "master"
   name = "golang.org/x/net"
-
-[[constraint]]
-  name = "cloud.google.com/go"
-  version = "0.26.0"
 
 [[constraint]]
   name = "github.com/gruntwork-io/terratest"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -55,3 +55,7 @@
 [[constraint]]
   branch = "master"
   name = "golang.org/x/net"
+
+[[constraint]]
+  name = "github.com/gruntwork-io/terratest"
+  version = "0.13.22"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -57,5 +57,9 @@
   name = "golang.org/x/net"
 
 [[constraint]]
+  name = "cloud.google.com/go"
+  version = "0.26.0"
+
+[[constraint]]
   name = "github.com/gruntwork-io/terratest"
   version = "0.13.22"
--- a/commands/cli.go
+++ b/commands/cli.go
@@ -116,6 +116,8 @@ func regionIsValid(ctx *gcp.GcpContext, region string) bool {
 }
 
 func gcpNuke(c *cli.Context) error {
+	// TODO accept multiple credentials and nuke resources on all the projects
+	// specified by a command line parameter we have authorization for.
 	ctx, err := gcp.DefaultContext()
 	if err != nil {
 		return errors.WithStackTrace(err)
--- a/commands/cli.go
+++ b/commands/cli.go
@@ -111,6 +111,10 @@ func promptForConfirmationBeforeNuking(force bool) (bool, error) {
 	}
 }
 
+func regionIsValid(ctx *gcp.GcpContext, region string) bool {
+	return ctx.ContainsRegion(region)
+}
+
 func gcpNuke(c *cli.Context) error {
 	ctx, err := gcp.DefaultContext()
 	if err != nil {
@@ -122,9 +126,9 @@ func gcpNuke(c *cli.Context) error {
 	excludedRegions := c.StringSlice("exclude-region")
 
 	for _, excludedRegion := range excludedRegions {
-		if !ctx.ContainsRegion(excludedRegion) {
+		if !regionIsValid(ctx, excludedRegion) {
 			return InvalidFlagError{
-				Name:  "exclude-regions",
+				Name:  "exclude-region",
 				Value: excludedRegion,
 			}
 		}
--- a/commands/cli.go
+++ b/commands/cli.go
@@ -7,8 +7,10 @@ import (
 
 	"github.com/gruntwork-io/gruntwork-cli/collections"
 
+	goerrors "errors"
 	"github.com/fatih/color"
 	"github.com/gruntwork-io/cloud-nuke/aws"
+	"github.com/gruntwork-io/cloud-nuke/gcp"
 	"github.com/gruntwork-io/cloud-nuke/logging"
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 	"github.com/gruntwork-io/gruntwork-cli/shell"
@@ -45,6 +47,26 @@ func CreateCli(version string) *cli.App {
 				},
 			},
 		},
+		{
+			Name:   "gcp",
+			Usage:  "Clean up GCP resources (GCE instances)",
+			Action: errors.WithPanicHandling(gcpNuke),
+			Flags: []cli.Flag{
+				cli.StringSliceFlag{
+					Name:  "exclude-region",
+					Usage: "regions to exclude",
+				},
+				cli.StringFlag{
+					Name:  "older-than",
+					Usage: "Only delete resources older than this specified value. Can be any valid Go duration, such as 10m or 8h.",
+					Value: "0s",
+				},
+				cli.BoolFlag{
+					Name:  "force",
+					Usage: "Skip nuke confirmation prompt. WARNING: this will automatically delete all resources without any confirmation",
+				},
+			},
+		},
 	}
 
 	return app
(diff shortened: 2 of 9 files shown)


acrisci:
#42 

Hi,

I am not an expert in Go or GCP so please let me know if I'm doing anything weird.

I tried to follow the conventions in the AWS example as much as possible, but the GCP libraries are different which made some simplifications in the data structures possible but required some extra state here and there.

Feedback on the approach would be appreciated.

Gopkg.toml
```diff
@@ -47,3 +47,11 @@
 [[constraint]]
   name = "github.com/fatih/color"
   version = "1.5.0"
+
+[[constraint]]
+  branch = "master"
```
acrisci: I don't think this is versioned.

gcp/gcp.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		for i != 0 {
+			result := <-results
+			if result.Err != nil {
+				logging.Logger.Warnf("could not delete resource: %s: %s Region=%s %s=%s",
+					result.Resource.ResourceName(), result.Resource.Name(), result.Resource.Region(),
+					result.Resource.LocationName(), result.Resource.Location())
+				// TODO certain errors may require a retry
+				//resourceList.PushFront(result.Resource)
```
acrisci: I think rate limiting is done in the library automatically, but still need to look into that a bit more. I have a strong feeling we'll need retries anyway for some reason though, hence why this is a bit hairy.

Gopkg.toml
brikis98: Oh, weird, they seem to version the actual code _within_ the repo: i.e., every package has a version number in its name, such as `spanner/v1` and `calendar/v3`. I guess we'll end up always pulling down latest and then have to `import` the right versions...

commands/cli.go
```diff
@@ -63,6 +84,85 @@ func parseDurationParam(paramValue string) (*time.Time, error) {
...
+		}
+
+		fmt.Println()
+		err := ctx.NukeAllResources(resources)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
brikis98: This if/else prompt code is prob identical to the AWS version, so perhaps extract into a reusable `promptForConfirmationBeforeNuking()` function.

gcp/gcp.go
```diff
@@ -0,0 +1,124 @@
...
+	results := make(chan NukeWorkerResult, 100)
+	resourceList := list.New()
+
+	for _, resource := range resources {
+		resourceList.PushFront(resource)
+	}
+
+	for resourceList.Len() != 0 {
+		i := 0
+		for i < batchSize {
```
brikis98: `for i := 0; i < batchSize; i++ { ... }`

gcp/gcp.go
```diff
@@ -0,0 +1,124 @@
...
+
+		for i != 0 {
+			result := <-results
+			if result.Err != nil {
+				logging.Logger.Warnf("could not delete resource: %s: %s Region=%s %s=%s",
+					result.Resource.ResourceName(), result.Resource.Name(), result.Resource.Region(),
+					result.Resource.LocationName(), result.Resource.Location())
+				// TODO certain errors may require a retry
+				//resourceList.PushFront(result.Resource)
+				return result.Err
```
brikis98: I think with AWS, we have been logging errors, but allowing the code to run to the end before returning the errors. That way, a failed delete early in the process doesn't prevent deletion of everything later in the process...

gcp/gce_instance.go
```diff
@@ -0,0 +1,85 @@
...
+			if creationTime.After(excludeAfter) {
+				continue
+			}
+
+			instance := GceInstanceResource{
+				InstanceName: apiInstance.Name,
+				Zone:         zone,
+				RegionName:   region,
+			}
+			var r GcpResource = instance
```
brikis98: Why this extra intermediate variable?

gcp/gce_instance.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+
+			zone, err := zoneFromUrl(apiInstance.Zone)
+			if err != nil {
+				return nil, err
+			}
+
+			region, err := regionFromZone(ctx, zone)
+			if err != nil {
+				return nil, err
```
brikis98: Consider logging errors and collecting them in a list, but not returning them until the very end. That way, a single failed instance delete early on doesn't prevent all others from being deleted later.

gcp/gce_instance.go
```diff
@@ -0,0 +1,85 @@
+package gcp
+
+import (
+	"errors"
+	"strings"
+	"time"
+)
+
+func zoneFromUrl(url string) (string, error) {
```
brikis98: Add unit tests for this method. It's important that we know it works correctly so we don't nuke the wrong zone/region.

gcp/gce_instance.go
```diff
@@ -0,0 +1,85 @@
...
+	split := strings.Split(url, "/")
+	if len(split) == 0 {
+		return "", errors.New("got invalid zone url: " + url)
+	}
+	return split[len(split)-1], nil
+}
+
+func regionFromZone(ctx *GcpContext, zone string) (string, error) {
```
brikis98: Add unit tests for this method. It's important that we know it works correctly so we don't nuke the wrong zone/region.

gcp/gce_instance_test.go
```diff
@@ -0,0 +1,128 @@
...
+	return err
+}
+
+func TestNukeInstances(t *testing.T) {
+	t.Parallel()
+
+	ctx, err := DefaultContext()
+	if err != nil {
+		assert.Fail(t, gruntworkerrors.WithStackTrace(err).Error())
+		return
```
brikis98: Use `t.Fail` and you don't need the `return`.

gcp/gce_instance_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+		_, err = ctx.Service.Instances.Delete(ctx.Project, testZone(), protectedInstanceName).Do()
+		if err != nil {
+			assert.Fail(t, gruntworkerrors.WithStackTrace(err).Error())
+		}
+	}()
+
+	err = createTestInstance(ctx, instanceName, false)
+	if err != nil {
+		assert.Fail(t, gruntworkerrors.WithStackTrace(err).Error())
```
brikis98: Use `t.Fail` so the test stops running at this stage. Same below.

gcp/gce_instance_test.go
```diff
@@ -0,0 +1,128 @@
...
+	err = ctx.NukeAllResources(instances)
+	if err != nil {
+		assert.Fail(t, gruntworkerrors.WithStackTrace(err).Error())
+	}
+
+	// status doesn't update immediately, so give it a minute to show it is
+	// terminating
+	tries := 0
+	lastStatus := ""
+	for tries < 20 {
```
brikis98: Looks like you're not incrementing `tries`? Use `for tries := 0; tries < 20; tries++`.

gcp/gce_instance_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+		_, err := call.Do()
+		if err != nil {
+			assert.Fail(t, gruntworkerrors.WithStackTrace(err).Error())
+		}
+
+		_, err = ctx.Service.Instances.Delete(ctx.Project, testZone(), protectedInstanceName).Do()
+		if err != nil {
+			assert.Fail(t, gruntworkerrors.WithStackTrace(err).Error())
+		}
```
brikis98: Extract the contents of this `defer` into a named function to make it clearer what it does.

gcp/gcp.go
```diff
@@ -0,0 +1,124 @@
...
+
+	regions, err := service.Regions.List(creds.ProjectID).Do()
+	if err != nil {
+		return nil, err
+	}
+
+	context := &GcpContext{
+		Client:  client,
+		Service: service,
+		Project: creds.ProjectID,
```
Etiene: I wonder if it's possible to specify the project id and nuke multiple projects instead of fetching the default one?

gcp/gce_instance_test.go
acrisci: That keeps going as well. What about [FailNow](https://godoc.org/github.com/stretchr/testify/assert#FailNow) on testify?

brikis98: Sorry, I meant `t.Fatal`!

brikis98: Also, the `assert` library has a flavor called `require` in the same library. So for thre code above, you could run `require.NoError(t, err)` and if `err` is not `nil`, it'll fail the test right there and stop.

gcp/gce_instance.go
acrisci: It sounds like this comment was meant for the nuke code. I updated the nuke code to log errors and keep going.

acrisci:
I've implemented the feedback and ready for the next round of review.

brikis98:
This code looks great to me, thanks @acrisci.

@josh-padnick Could you review the GCP specifics? If it looks good, feel free to merge. Note that you'll need to configure the CircleCI build to work with GCP.

josh-padnick:
'cc @robmorgan @autero1 as this relates to GCP

acrisci:
Thanks for the review.

I'm adopting Terratest as well and have some plans for contributions there. I'd rather do that in one go, so I'll make issues and comments.

acrisci:
I ran into some complications with putting terratest into the test suite. Running tests gives:

```
# github.com/gruntwork-io/cloud-nuke/vendor/google.golang.org/grpc/internal/channelz
../vendor/google.golang.org/grpc/internal/channelz/types_linux.go:41:15: undefined: unix.GetsockoptLinger
../vendor/google.golang.org/grpc/internal/channelz/types_linux.go:44:15: undefined: unix.GetsockoptTimeval
../vendor/google.golang.org/grpc/internal/channelz/types_linux.go:47:15: undefined: unix.GetsockoptTimeval
FAIL	github.com/gruntwork-io/cloud-nuke/gcp [build failed]
```

There is a bug in the grpc library in version 0.26.0 of cloud.google.com/go dependency of terratest that we must ensure to import terratest. Updating this to 0.35.1 in terratest and cloud-nuke resolves the issue.

josh-padnick:
@robmorgan Could you take the lead on this PR from here? Or if busy perhaps @autero1 could do so?

robmorgan:
@josh-padnick sure.

@acrisci Do you want to bump the `cloud.google.com/go` dependency to at least 0.35.1 then?

I'll look into updating it in Terratest.

acrisci:
Ok I took a fresh look at it and it turns out what actually needed to be done was to lock `golang.org/x/sys` at a specific version required by grpc. I'm not sure how I got it to work the other way.

Anyway I think this works now.

tonerdo:
Hi @acrisci, many thanks for this contribution. I just made a fix to the tests so please merge in the latest master and fix conflicts so that we can get this PR shipped. Thanks you!!!

acrisci:
Yeah all that's needed is to fix the build errors. I got pulled onto another project but I should be back on this soon.

innovia:
has this PR fallen behind?

acrisci:
@innovia The way things turned out for me, I don't really have a use for this. I feel bad about not finishing it, but being merged with a maintainer with no use case might be worse. And plus the build errors were kind of annoying.

I think the approach I used here is good and I hope someone else who may need this takes it on and benefits from the work I've done here.

PR #46, opened 2018-12-03
Adding support for nuking EFS file systems

robmorgan:
This PR is a **WIP**, however I wanted to open one to get early feedback.

### My questions

1. As an EFS file system has associated mount targets, should I delete them first (in `aws.go`. See the commented out code) or as part of the `nukeAllEfsFileSystems` function? Note: Essentially we need the EFS File System ID in order to be able to remove mount targets.
2. I have combined two seperate types in the `efs_types.go` file. Is that the correct behaviour or should I split them across two files.

### TODO

- [ ] Automated tests

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -166,6 +166,32 @@ func GetAllResources(regions []string, excludedRegions []string, excludeAfter ti
 		resourcesInRegion.Resources = append(resourcesInRegion.Resources, ebsVolumes)
 		// End EBS Volumes
 
+		// EFS Mount Targets
+		//mountTargetIds, err := getAllEfsMountTargets(session, region, excludeAfter)
+		//if err != nil {
+		//	return nil, errors.WithStackTrace(err)
+		//}
+
+		//efsMountTargets := EFSMountTargets{
+		//	MountTargetIds: awsgo.StringValueSlice(mountTargetIds),
+		//}
+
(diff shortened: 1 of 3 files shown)


brikis98:
Just to make sure I understand the issue:

1. To be able to delete an EFS file system, you must first delete all the mount targets.
1. To be able to delete mount targets, you need all the EFS file system IDs first.
1. So you're asking if those two should be deleted as one unit or two separate items that, in principle, could be run independently?

brikis98:
OK, thanks for the clarification. So is there ever a use case for _just_ deleting mount targets that's not associated with EFS? Or vice versa, deleting the EFS file system without first deleting mount targets? If there's no real-world scenario where that's likely, may as well keep them together and keep the code simple.

robmorgan:
@brikis98 No I don't think there is and I guess the purpose of cloud-nuke is to destroy everything. You cannot delete the file system without first deleting the mount targets (and the mount targets are essentially useless without the FS).

Etiene:
Just throwing in my two cents that then I also think it's OK to delete the mount targets together and have both types in one file.

robmorgan:
@Etiene okay, noted 👍

robmorgan:
I'm closing this for now and I'll either pick it back up again in the future, or somebody else can.

PR #47, opened 2018-12-04
Nuke VPC NAT Gateways

robmorgan:
**WIP**

This PR adds support for nuking VPC NAT Gateways.

It successfully blocks until the gateway reaches the `deleted` state, so we can then remove the associated Elastic IP address.

### TODO

- [ ] Automated Tests

### DIFF
--- a/aws/nat_gateway.go
+++ b/aws/nat_gateway.go
@@ -22,12 +22,14 @@ func waitUntilNatGatewayDeleted(svc *ec2.EC2, input *ec2.DescribeNatGatewaysInpu
 			return err
 		}
 
-		if *result.NatGateways[0].State == "deleted" {
-			return nil
+		if result.NatGateways != nil && len(result.NatGateways) > 0 {
+			if *result.NatGateways[0].State == "deleted" {
+				return nil
+			}
 		}
 
+		logging.Logger.Debug("NAT Gateway still not deleted. Will sleep for 5 seconds and check again")
 		time.Sleep(5 * time.Second)
-		logging.Logger.Debug("Waiting for NAT Gateway to be deleted")
 	}
 
 	return NatGatewayDeleteError{}
--- a/aws/nat_gateway.go
+++ b/aws/nat_gateway.go
@@ -1,0 +1,107 @@
+package aws
+
+import (
+	"time"
+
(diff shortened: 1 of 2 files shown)


aws/nat_gateway.go
```diff
@@ -0,0 +1,107 @@
...
+		result, err := svc.DescribeNatGateways(input)
+		if err != nil {
+			if awsErr, isAwsErr := err.(awserr.Error); isAwsErr && awsErr.Code() == "NatGatewayNotFound" {
+				return nil
+			}
+
+			return err
+		}
+
+		if *result.NatGateways[0].State == "deleted" {
```
brikis98: Check `NatGateways` isn't empty or `nil`.

aws/nat_gateway.go
```diff
@@ -0,0 +1,107 @@
...
+			return nil
+		}
+
+		time.Sleep(5 * time.Second)
+		logging.Logger.Debug("Waiting for NAT Gateway to be deleted")
```
brikis98: To make the logs easier to follow, move log statement before sleep and say something like, "NAT Gateway still not deleted. Will sleep for 5 seconds and check again."

aws/nat_gateway.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if len(deletedNatGatewayIDs) > 0 {
+		for _, natGatewayID := range deletedNatGatewayIDs {
+			err := waitUntilNatGatewayDeleted(svc, &ec2.DescribeNatGatewaysInput{
+				NatGatewayIds: []*string{natGatewayID},
+			})
+
+			if err != nil {
+				logging.Logger.Errorf("[Failed] %s", err)
+				return errors.WithStackTrace(err)
```
brikis98: I think we typically log errors, but try not to exit with an error so that we delete as much as possible each run. @tonerdo can confirm.

aws/nat_gateway.go
robmorgan: ok

wandergeek:
want

robmorgan:
I'm closing this for now and I'll either pick it up again in the future or somebody else can.

PR #50, opened 2019-02-21
Fix AuthFailure Error

tonerdo:
This PR fixes the constant `AuthFailure: AWS was not able to validate the provided access credentials` errors when running tests. It turns out that AWS added a new GovCloud region `us-gov-east-1` which I wasn't explicitly ignoring as I did the China and other GovCloud regions.

The simple change was to add `us-gov-east-1` to the `reservedRegions` array so we don't try to get resources from it

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -17,7 +17,7 @@ import (
 func GetAllRegions() []string {
 	// chinese and government regions are not accessible with regular accounts
 	reservedRegions := []string{
-		"cn-north-1", "cn-northwest-1", "us-gov-west-1",
+		"cn-north-1", "cn-northwest-1", "us-gov-west-1", "us-gov-east-1",
 	}
 
 	resolver := endpoints.DefaultResolver()


aws/aws.go
```diff
@@ -17,7 +17,7 @@ import (
 func GetAllRegions() []string {
 	// chinese and government regions are not accessible with regular accounts
 	reservedRegions := []string{
-		"cn-north-1", "cn-northwest-1", "us-gov-west-1",
+		"cn-north-1", "cn-northwest-1", "us-gov-west-1", "us-gov-east-1",
```
brikis98: Should this be configurable? That is, other cloud-nuke users may use these regions... Should there be some CLI flag to specify which ones to use or ignore? If so, it's not a high priority change, so feel free to file an issue for it so we can come back to it later.

tonerdo: I think that's a separate feature from this. This just concerns the regions in China and GovCloud that you need special credentials to access. We already have an `exclude-region` CLI option that can be used with the regular AWS regions

robmorgan:
LGTM 👍

PR #53, opened 2019-03-02
Add checksums to release files

conn:
This is to address an unopened issue similar to:
https://github.com/gruntwork-io/terragrunt/issues/610

The checksums of all binaries within `bin/` will be saved within
`SHA256SUMS`. The integrity of binaries can now be verified with this
command: `sha256sum --check --ignore-missing SHA256SUMS`.

In the future, Gruntwork should consider using a PGP key to
cryptographically sign this checksums file and to distribute the
signature along with the rest of the release files. Doing so will likely
require using private CI/CD infrastructure instead of CircleCI so that
Gruntwork may have full control over their private signing key.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,7 @@ jobs:
       - checkout
       - attach_workspace:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
+      - run: cd bin && sha256sum * > SHA256SUMS
       - run: upload-github-release-assets bin/*
 
 workflows:


.circleci/config.yml
```diff
@@ -67,6 +67,7 @@ jobs:
       - checkout
       - attach_workspace:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
+      - run: cd bin && sha256sum * > SHA256SUMS
```
tonerdo: Wouldn't this be better as `sha256sum bin/* > bin/SHA256SUMS`? I'm not sure if each `run` in CircleCI is executed in its own shell instance because if it isn't the below command will fail as there'll be no `bin` folder in the `bin` folder

tonerdo: This https://github.com/gruntwork-io/health-checker/pull/14#issuecomment-469244605 answers my question

Issue #55, opened 2019-03-08
Add ability to nuke specific regions


saady:
Instead of passing all the regions to the exclude-region flag, we need to add a flag to specify a list of regions to be nuked.

rraiatsysgain:
+1

saurabh-hirani:
@autero1 This can be closed as it is addressed by https://github.com/gruntwork-io/cloud-nuke/pull/71 in the latest release.

autero1:
Closing in favor of #71

PR #56, opened 2019-03-08
Add ability to select a region to nuke

saady:
Ref #55

### DIFF
--- a/commands/cli.go
+++ b/commands/cli.go
@@ -43,6 +43,10 @@ func CreateCli(version string) *cli.App {
 					Name:  "force",
 					Usage: "Skip nuke confirmation prompt. WARNING: this will automatically delete all resources without any confirmation",
 				},
+				cli.StringSliceFlag{
+					Name:  "region",
+					Usage: "Specific region to nuke",
+				},
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


commands/cli.go
```diff
@@ -64,7 +68,21 @@ func parseDurationParam(paramValue string) (*time.Time, error) {
 }
 
 func awsNuke(c *cli.Context) error {
+	specificRegion := c.StringSlice("region")
```
Etiene: Wait, so you want to specify multiple regions or one region?

saady: specify one or multiple regions `--region eu-west-3 --region eu-west-1` here I want to nuke the resources in eu-west-1 and eu-west-3 instead of going through all regions.

brikis98: That makes sense. Could you update the README to mention this?

commands/cli.go
```diff
@@ -43,6 +43,10 @@ func CreateCli(version string) *cli.App {
 					Name:  "force",
 					Usage: "Skip nuke confirmation prompt. WARNING: this will automatically delete all resources without any confirmation",
 				},
+				cli.StringSliceFlag{
+					Name:  "region",
+					Usage: "Specific region to nuke",
```
brikis98:
```suggestion
					Usage: "Specific region(s) to nuke",
```

commands/cli.go
brikis98: Also, perhaps rename the variable from `specificRegion` to something like `selectedRegions`?

autero1:
Closing in favor of #71

PR #59, opened 2019-04-11
Attempt to stabilize tests

yorinasub17:
This is an attempt at stabilizing the tests:

- [This build](https://circleci.com/gh/gruntwork-io/cloud-nuke/9036) appeared to have failed because it picked some malformed subnets. Upon checking `getSubnetsInDifferentAZs`, it looked like it was pulling in all subnets for the region, so I added in a filter to only get the default subnets to see if that makes the build more consistent.
- [This build](https://circleci.com/gh/gruntwork-io/cloud-nuke/9020) failed because the ECS service could not be found when draining. I am guessing this is an eventual consistency issue, so added a wait right after creating the ECS service.


yorinasub17:
Ok going to merge this in to see if we start getting more greens on the frequent builds.

PR #60, opened 2019-04-25
Fix date boundary test failure

yorinasub17:
This fixes another test failure that occurs frequently. Specifically, anytime the tests run at midnight, `TestParseDuration` will fail. See https://circleci.com/gh/gruntwork-io/cloud-nuke/9388 for a sample failure output.


yorinasub17:
Note that this test will likely fail at midnight on the last day of the month, but I think we can live with the monthly failure, as opposed to a daily failure.

yorinasub17:
Merging this in. Thanks for the review! Let's see if it passes for the next iteration of midnight testing of this.

PR #61, opened 2019-04-30
More fixes for test stabilization

yorinasub17:
Frequently the ECS tests fail timing out waiting for the service. I noticed that when this happens, the ECS service actually does show up eventually, so adding a retry loop to see if that makes the tests more stable. So far, ran the build 6 times and it only failed once for a different test (https://circleci.com/gh/gruntwork-io/workflows/cloud-nuke/tree/yori-fix-tests-round-3).


yorinasub17:
Thanks for the review! Will merge and hopefully, the number of failed master builds start to decrease.

PR #66, opened 2019-08-14
Add defaults-aws subcommand

bwhaley:
This PR adds a new subcommand, `cloud-nuke defaults-aws`. The command does the following:
* Discovers and removes all default VPCs from an account
* Discovers all default security groups and removes their default ingress/egress rules

#### Background
New AWS accounts have a [Default VPC](https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html) in each region. If you create a resource (e.g. an EC2 instance) without specifying a VPC ID and subnet, it will be created within the default VPC by default. For production accounts, indeed for any account which is used for purposes other than simple sandbox tests, the default VPC can be considered a risk as it may accumulate unused/undesired resources over time.

Additional, all VPCs, not just default VPCs, include a default security group. The default security group has an ingress rule that allows all traffic within the same group, and an egress rule that permits all outbound traffic (e.g. all ports, all protocols, to destination `0.0.0.0/0`). This is an antipattern for obvious reasons. Removing this rule is a requirement for compliance with some security standards, such as the CIS AWS Foundations Benchmark.

#### Usage
I've added the `defaults-aws` subcommand to remove these defaults. Example output:
```
cloud-nuke defaults-aws --force   ✔  ⚙  7615  07:52:17
INFO[2019-08-14T07:52:26-07:00] Identifying enabled regions
INFO[2019-08-14T07:52:27-07:00] Found enabled region eu-north-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region ap-south-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region eu-west-3
INFO[2019-08-14T07:52:27-07:00] Found enabled region eu-west-2
INFO[2019-08-14T07:52:27-07:00] Found enabled region eu-west-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region ap-northeast-2
INFO[2019-08-14T07:52:27-07:00] Found enabled region ap-northeast-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region sa-east-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region ca-central-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region ap-southeast-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region ap-southeast-2
INFO[2019-08-14T07:52:27-07:00] Found enabled region eu-central-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region us-east-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region us-east-2
INFO[2019-08-14T07:52:27-07:00] Found enabled region us-west-1
INFO[2019-08-14T07:52:27-07:00] Found enabled region us-west-2
INFO[2019-08-14T07:52:27-07:00] Discovering default VPCs
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-09f5580fdcbb2f45c eu-north-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-02a9144962943118c ap-south-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-0b8fa626656a2d406 eu-west-3
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-09dc63dd3e9ef59b5 eu-west-2
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-08748b3914dbe20f3 eu-west-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-07e471a2781ce611e ap-northeast-2
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-0d756684f04e8b539 ap-northeast-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-0d6a8e45677eb1adb sa-east-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-0b41d4bc601d51418 ca-central-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-026a2ea431bde67aa ap-southeast-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-08bd2b462fcdb511c ap-southeast-2
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-090846f0c0d7886c7 eu-central-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-033904a71040f7305 us-east-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-0da379ffd7cb43d90 us-east-2
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-001199441593c91bd us-west-1
INFO[2019-08-14T07:52:37-07:00] * Default VPC vpc-0c2dd0d4dbf70d539 us-west-2
INFO[2019-08-14T07:52:37-07:00] Nuking VPC vpc-09f5580fdcbb2f45c in region eu-north-1
INFO[2019-08-14T07:52:38-07:00] ...detaching Internet Gateway igw-01faf4477ddc2aaa1
INFO[2019-08-14T07:52:38-07:00] ...deleting Internet Gateway igw-01faf4477ddc2aaa1
INFO[2019-08-14T07:52:38-07:00] ...deleting subnet subnet-0c57018894f22cd3b
INFO[2019-08-14T07:52:39-07:00] ...deleting subnet subnet-03f333046f700fdc1
INFO[2019-08-14T07:52:39-07:00] ...deleting subnet subnet-0c6b2a41d26840a04
INFO[2019-08-14T07:52:40-07:00] ...deleting Security Group sg-08f25d8ce6b83c2ba
INFO[2019-08-14T07:52:40-07:00] ...deleting VPC vpc-09f5580fdcbb2f45c
INFO[2019-08-14T07:52:40-07:00] Nuking VPC vpc-02a9144962943118c in region ap-south-1
INFO[2019-08-14T07:52:41-07:00] ...detaching Internet Gateway igw-06381f015a43a06f0
INFO[2019-08-14T07:52:41-07:00] ...deleting Internet Gateway igw-06381f015a43a06f0
INFO[2019-08-14T07:52:42-07:00] ...deleting subnet subnet-024ee83d56944842f
INFO[2019-08-14T07:52:42-07:00] ...deleting subnet subnet-07616d8c3e6edf7cb
INFO[2019-08-14T07:52:43-07:00] ...deleting subnet subnet-0549721d4262b3c5e
INFO[2019-08-14T07:52:44-07:00] ...deleting Security Group sg-09f2a83c6e69ec27a
INFO[2019-08-14T07:52:44-07:00] ...deleting VPC vpc-02a9144962943118c
INFO[2019-08-14T07:52:44-07:00] Nuking VPC vpc-0b8fa626656a2d406 in region eu-west-3
INFO[2019-08-14T07:52:45-07:00] ...detaching Internet Gateway igw-078c47bce8bf47af8
INFO[2019-08-14T07:52:45-07:00] ...deleting Internet Gateway igw-078c47bce8bf47af8
[snip]
INFO[2019-08-14T07:53:22-07:00] Finished nuking default VPCs in all regions
INFO[2019-08-14T07:53:22-07:00] Discovering default security groups
INFO[2019-08-14T07:53:25-07:00] * Default rules for SG sg-00d959b8d5d11f48b default eu-west-3
INFO[2019-08-14T07:53:25-07:00] * Default rules for SG sg-070fc543465a33a83 default eu-west-1
INFO[2019-08-14T07:53:25-07:00] * Default rules for SG sg-03947dbd85c595888 default ca-central-1
INFO[2019-08-14T07:53:25-07:00] ...revoking default rules from Security Group sg-00d959b8d5d11f48b
INFO[2019-08-14T07:53:26-07:00] Egress rule not present (ok)
INFO[2019-08-14T07:53:26-07:00] Ingress rule not present (ok)
INFO[2019-08-14T07:53:26-07:00] ...revoking default rules from Security Group sg-070fc543465a33a83
INFO[2019-08-14T07:53:27-07:00] Egress rule not present (ok)
INFO[2019-08-14T07:53:28-07:00] Ingress rule not present (ok)
INFO[2019-08-14T07:53:28-07:00] ...revoking default rules from Security Group sg-03947dbd85c595888
INFO[2019-08-14T07:53:28-07:00] Egress rule not present (ok)
INFO[2019-08-14T07:53:28-07:00] Ingress rule not present (ok)
INFO[2019-08-14T07:53:28-07:00] Finished nuking default Security Groups in all regions
```

The only supported flag is `--force` to allow for non-interactive use. 

#### Testing
One challenge with this change was that many of the tests rely on the presence of a default VPC and/or security group to function. e.g. new EC2 instances, when launched by the tests, are launched in to the default VPC. This command removes the default VPC. Furthermore, any other activities in the same account, such as tests from other code repositories, would have issues.

Therefore I used the `ec2iface.EC2API` and the [GoMock](https://github.com/golang/mock) framework to mock the APIs for this testing. This was a cumbersome effort and required quite a lot of code, but the end result is that we have thorough testing for this change without impacting "real" AWS resources.

#### Closing
Please give this a careful review. The nature of this command is destructive, obviously - deleting VPCs and security groups. I've tested it myself, but bugs in this feature could be very costly.


aws/ec2.go
```diff
@@ -110,3 +113,394 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
...
+		}
+		if vpcId != "" {
+			vpcs[i].VpcId = vpcId
+		} else {
+			// Ignore regions that don't have a default VPC
+			vpcs = append(vpcs[:i], vpcs[i+1:]...)
+			i--
+		}
+	}
+	return vpcs, nil
```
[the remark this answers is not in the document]
yorinasub17: Also, I am confused about the input var. Is that all VPCs that this has detected, or is the input supposed to be one example VPC per region?

aws/ec2.go
```diff
@@ -110,3 +113,394 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
...
+		logging.Logger.Infof("...deleting Internet Gateway %s", awsgo.StringValue(igw.InternetGateways[0].InternetGatewayId))
+		_, err = v.svc.DeleteInternetGateway(
+			&ec2.DeleteInternetGatewayInput{
+				InternetGatewayId: igw.InternetGateways[0].InternetGatewayId,
+			},
+		)
+		if err != nil {
+			return errors.WithStackTrace(err)
+		}
+	}
```
yorinasub17: What if there is more than one? We should probably add a guard here to log some warning and the fact that we are skipping it.

aws/ec2.go
```diff
@@ -110,3 +113,394 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
 	logging.Logger.Infof("[OK] %d instance(s) terminated in %s", len(instanceIds), *session.Config.Region)
 	return nil
 }
+
+func GetEc2ServiceClient(region string) ec2iface.EC2API {
+	return ec2.New(newSession(region))
+}
+
```
yorinasub17: I am inclined to move all the VPC stuff to a different file, but don't have super strong feelings about it.

aws/ec2.go
```diff
@@ -110,3 +113,394 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
...
+	}
+	for _, securityGroup := range securityGroups.SecurityGroups {
+		if *securityGroup.GroupName == "default" {
+			groupIds = append(groupIds, awsgo.StringValue(securityGroup.GroupId))
+		}
+	}
+	return groupIds, nil
+}
+
+func GetSecurityGroups(regions []string) ([]DefaultSecurityGroup, error) {
```
yorinasub17:
```suggestion
func GetDefaultSecurityGroups(regions []string) ([]DefaultSecurityGroup, error) {
```
So it is clear you are only getting the default security groups.

aws/ec2.go
```diff
@@ -110,3 +113,394 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
...
+}
+
+type DefaultSecurityGroup struct {
+	GroupName string
+	GroupId   string
+	Region    string
+	svc       ec2iface.EC2API
+}
+
+func DescribeSecurityGroups(svc ec2iface.EC2API) ([]string, error) {
```
yorinasub17:
```suggestion
func DescribeDefaultSecurityGroups(svc ec2iface.EC2API) ([]string, error) {
```

So it is clear you are only getting the default security groups.

commands/cli.go
```diff
@@ -135,3 +142,107 @@ func awsNuke(c *cli.Context) error {
...
+	logging.Logger.Infoln("Identifying enabled regions")
+	regions, err := aws.GetEnabledRegions()
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+	for _, region := range regions {
+		logging.Logger.Infof("Found enabled region %s", region)
+	}
+
+	err = defaultVpcs(c, regions)
```
yorinasub17: NIT: the function name should probably indicate that it will nuke the default VPCs. At a glance, it is not exactly clear what this is expected to do.

commands/cli.go
```diff
@@ -135,3 +142,107 @@ func awsNuke(c *cli.Context) error {
...
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+
+	err = defaultSecurityGroups(c, regions)
```
yorinasub17: NIT: same as above re:naming.

commands/cli.go
```diff
@@ -135,3 +142,107 @@ func awsNuke(c *cli.Context) error {
...
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+	return nil
+}
+
+func defaultVpcs(c *cli.Context, regions []string) error {
+	logging.Logger.Infof("Discovering default VPCs")
+	vpcPerRegion := aws.NewVpcPerRegion(regions)
+	vpcPerRegion, err := aws.GetDefaultVpcs(vpcPerRegion)
```
yorinasub17: Out of curiosity, why can't we pass in the regions and lookup the defaults in each one there?

aws/ec2_test.go
```diff
@@ -222,3 +243,438 @@ func TestNukeInstances(t *testing.T) {
...
+	return &ec2.DeleteSecurityGroupInput{
+		GroupId: awsgo.String(securityGroupId),
+	}
+}
+
+func getDeleteVpcInput(vpcId string) *ec2.DeleteVpcInput {
+	return &ec2.DeleteVpcInput{
+		VpcId: awsgo.String(vpcId),
+	}
+}
```
yorinasub17: Maybe extract these helpers into a `mock_ec2_utils_for_test.go` file? (the name is a hack to ensure it is only compiled and added to the binary during testing).

aws/ec2_test.go
```diff
@@ -222,3 +243,438 @@ func TestNukeInstances(t *testing.T) {
 		assert.NotContains(t, instances, *instanceID)
 	}
 }
+
```
yorinasub17: Since these mock based tests are completely different from any other test in cloud-nuke, I think they should be extracted into their own file (e.g `ec2_unit_test.go`) with a comment at the top to help guide someone with the nature of the test. It took me a bit to understand what was going on in these tests (setup mocks to expect a bunch of calls => run the function)

aws/aws.go
```diff
@@ -1,44 +1,95 @@
...
+		regionNames = append(regionNames, awsgo.StringValue(region.RegionName))
 	}
 
-	return regions
+	return regionNames, nil
 }
 
 func getRandomRegion() string {
-	allRegions := GetAllRegions()
+	allRegions, _ := GetEnabledRegions()
```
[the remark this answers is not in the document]
bwhaley: Good point!

aws/ec2.go
bwhaley: There can only be 0 or 1. I will add a statement about skipping if there are 0.

aws/ec2.go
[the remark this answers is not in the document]
bwhaley: The input var is really just a `Vpc{}` object for each region with the `Region` and `svc` fields already assigned. The reason to do it this way (and it does feel a bit funky to me) is that the `svc` object is overridden by the mock object in the tests. in the CLI we call `NewVpcPerRegion()` followed by `GetDefaultVpcs()`. The tests skip the `NewVpcPerRegion()` (which assigns `svc` to the "real" service object and instead assigns the mock object itself.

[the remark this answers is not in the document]
bwhaley: Re: the loop above - nice catch. This was an experiment I had neglected to replace. I'll go with something similar to your suggestion.

commands/cli.go
bwhaley: See comment above re: the service client. This is different between the default VPCs and the security group because with default VPCs, there is exactly one per region. With default security groups, there is one per VPC, and potentially many VPCs per region. That's why it's handled differently between the two.

aws/ec2.go
bwhaley: The reason I kept it all in `ec2.go` is because that's how AWS has it organized. All the VPC methods are in the `ec2` package. Not that we necessarily have to follow their lead, which arguably doesn't make sense to customers, but that's the reasoning.

aws/ec2_test.go
bwhaley: Sure.

aws/ec2.go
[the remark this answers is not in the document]
yorinasub17: > The reason to do it this way (and it does feel a bit funky to me) is that the svc object is overridden by the mock object in the tests.

Ah ok. Probably worth adding an extended comment there since I am bound to come in and try to "optimize" this at a later point in time and shoot myself in the foot!

aws/ec2.go
yorinasub17: Ah gotcha. It's less for customers (customers won't see this code) and more for us. I find it more intuitive for them to be separate because I spend more time in the console than the AWS Go SDK, and VPC has its own console vs EC2.

Come to think of it, I wonder why the go SDK has VPC under EC2 when the console separates them out...

bwhaley:
Thanks for the review.

I'm open to ideas on how to include integration tests as that's my preference as well. In fact, you may have noticed that I actually wrote integration tests before I realized that it would break other tests that run in parallel. Then I realized that it might also run in parallel with other test/work in the same AWS account that also depends on default VPCs. Given the distributed nature of the tests and multiple parallel uses of an account, I can't really see a good way to do integration testing of this feature (other than having it spin up a separate account and do stuff there as you suggested we might do for the other compliance work).

commands/cli.go
yorinasub17: Thanks for the explanation!

yorinasub17:
> (other than having it spin up a separate account and do stuff there as you suggested we might do for the other compliance work).

I think Jim mentioned that we can't quite do this in an automated test because of how hard it is to shut down an account (requires email verification and what not). Otherwise, I would have suggested adding it in here.

bwhaley:
We may want to discuss how best to do it with the broader team, perhaps in an IaC team meeting or async via slack.

In any case, the mock testing is better than no tests :)

aws/ami_test.go
```diff
@@ -76,7 +76,10 @@ func createTestAMI(t *testing.T, session *session.Session, name string) (*ec2.Im
 func TestListAMIs(t *testing.T) {
 	t.Parallel()
 
-	region := getRandomRegion()
+	region, err := getRandomRegion()
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
yorinasub17: FYI you can use `assert.NoError(t, err)`, which will basically do the above.

aws/ec2_unit_test.go
```diff
@@ -0,0 +1,284 @@
+// These tests use GoMock and the ec2iface to provide a mock framework for testing the EC2 API
+// Unlike other tests in cloud-nuke, nuking the default VPCs and security groups is not an option.
+// Other tests within cloud-nuke depend on the default VPCs/SGs to function, and other projects
+// may be using the same AWS account at the same time. Deleting the default VPCs would break things.
+// Therefore, the default VPC/SG nuke testing is mocked as unit tests.
+// To generate the EC2API mock, install https://github.com/golang/mock, then use the following:
+// mockgen -source vendor/github.com/aws/aws-sdk-go/service/ec2/ec2iface/interface.go -destination aws/mocks/gomock-EC2API.go
```
yorinasub17: Ah didn't realize that mock file was autogen'd. Good call on including those instructions!

bwhaley: You thought I wrote all 20,920 lines?! I'm flattered!

yorinasub17: Oh no I thought you copied pasted it from the dark web 😄

bwhaley:
Thanks for the review! Merging and releasing this one.

Issue #26, opened 2018-03-28
Invalid response on prompt leads to immediate exit.


josh-padnick:
Cloud-nuke takes a while to identify all the resources it plans to delete in AWS and the finally displays this prompt:

```
Are you sure you want to nuke all listed resources? Enter 'nuke' to confirm: 
```

I accidentally entered `yes` instead of `nuke`, and the program immediately exited, which means I have to re-scan all AWS resources. Instead catch the error and give the user another chance to enter `nuke`. Also consider advising the user to use `CTRL+C` to cancel.

marinalimeira:
I believe this was solved at #73! Closing the issue.

PR #73, opened 2019-10-10
Prompt for 'nuke' again on invalid responses

aqche:
Here is my take on addressing issue #26. Instead of exiting if the user types in an invalid input to the nuke prompt, the user will be notified that the input was invalid and prompted again. A note that you can exit with ^C is added as well.

### DIFF
--- a/commands/cli.go
+++ b/commands/cli.go
@@ -150,8 +150,8 @@ func awsNuke(c *cli.Context) error {
 	}
 
 	if !c.Bool("force") {
-		prompt := "\nAre you sure you want to nuke all listed resources? Enter 'nuke' to confirm: "
-		proceed, err := confirmationPrompt(prompt)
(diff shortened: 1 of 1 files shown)


autero1:
Thanks for the PR @aqche !

The change makes sense, will run the test suite now.

commands/cli.go
```diff
@@ -262,19 +262,26 @@ func nukeDefaultSecurityGroups(c *cli.Context, regions []string) error {
 	return nil
 }
 
-func confirmationPrompt(prompt string) (bool, error) {
+func confirmationPrompt(prompt string, maxPrompts int) (bool, error) {
```
autero1: Maybe add an explanation here why we're allowing multiple tries.

aqche:
@autero1 I added a comment explaining the reasoning behind multiple tries. Also, I removed the unneeded `else` block just to make things slightly cleaner. Thanks!

autero1:
Thanks! I'll merge this in now.

PR #84, opened 2020-01-07
Add a plan for the RDS nuke

marinalimeira:

PR #86, opened 2020-01-07
Add feature to delete RDS instances

marinalimeira:
Following #84.

From the first part:
- [X] Create a new RDS instance
- [X] The lookup for the instances based on region
- [X] Nuke the instance based on the lookup results (with SkipFinalSnapshot: true)
- [x] Run from CLI
- [x] Delete RDS remaining instances on AWS 
- [x] Add the RDS to be searchable by the --resource-type flag.
- [x] Get instances by age (older than xx)

### DIFF
--- a/aws/rds.go
+++ b/aws/rds.go
@@ -11,7 +11,7 @@ import (
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
-func getAllRdsInstances(session *session.Session, region string, excludeAfter time.Time) ([]*string, error) {
+func getAllRdsInstances(session *session.Session, excludeAfter time.Time) ([]*string, error) {
 	svc := rds.New(session)
 
 	result, err := svc.DescribeDBInstances(&rds.DescribeDBInstancesInput{})
@@ -35,7 +35,7 @@ func nukeAllRdsInstances(session *session.Session, names []*string) error {
 	svc := rds.New(session)
 
 	if len(names) == 0 {
-		logging.Logger.Infof("No RDS DB Instanceto nuke in region %s", *session.Config.Region)
+		logging.Logger.Infof("No RDS DB Instance to nuke in region %s", *session.Config.Region)
 		return nil
 	}
 
@@ -51,7 +51,7 @@ func nukeAllRdsInstances(session *session.Session, names []*string) error {
 		_, err := svc.DeleteDBInstance(params)
 
 		if err != nil {
-			logging.Logger.Errorf("[Failed] %s", err)
+			logging.Logger.Errorf("[Failed] %s: %s", *name, err)
 		} else {
 			deletedNames = append(deletedNames, name)
 			logging.Logger.Infof("Deleted RDS DB Instance: %s", awsgo.StringValue(name))
--- a/aws/rds.go
+++ b/aws/rds.go
@@ -72,7 +72,7 @@ func nukeAllRdsInstances(session *session.Session, names[]*string) error {
       logging.Logger.Errorf("[Failed] %s", err)
     } else {
       deletedNames = append(deletedNames, name)
-      logging.Logger.Infof("Deleted ELB: %s", *name)
+      logging.Logger.Infof("Deleted RDS: %s", *name)
     }
   }
 
--- a/aws/rds.go
+++ b/aws/rds.go
@@ -23,9 +23,9 @@ func getAllRdsInstances(session *session.Session, region string, excludeAfter ti
 	var names []*string
 
 	for _, database := range result.DBInstances {
-		//if excludeAfter.After(*database.InstanceCreateTime) {
-		names = append(names, database.DBInstanceIdentifier)
-		//}
+		if excludeAfter.After(*database.InstanceCreateTime) {
+		  names = append(names, database.DBInstanceIdentifier)
+		}
 	}
 
 	return names, nil
--- a/aws/rds.go
+++ b/aws/rds.go
@@ -23,7 +23,7 @@ func waitUntilRdsDeleted(svc *rds.RDS, input *rds.DescribeDBInstancesInput) erro
     }
 
     time.Sleep(1 * time.Second)
-    logging.Logger.Debug("Waiting for ELB to be deleted")
+    logging.Logger.Debug("Waiting for RDS to be deleted")
   }
 
   return RdsDeleteError{}
--- a/aws/rds.go
+++ b/aws/rds.go
@@ -1,33 +1,16 @@
 package aws
 
 import (
+	"context"
 	"time"
 
 	awsgo "github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
+	// "github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/rds"
 	"github.com/gruntwork-io/cloud-nuke/logging"
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
-
-func waitUntilRdsDeleted(svc *rds.RDS, input *rds.DescribeDBInstancesInput) error {
-	for i := 0; i < 240; i++ {
-		_, err := svc.DescribeDBInstances(input)
-		if err != nil {
-			if awsErr, isAwsErr := err.(awserr.Error); isAwsErr && awsErr.Code() == "DBInstanceNotFound" {
-				return nil
-			}
-
-			return err
-		}
-
-		time.Sleep(1 * time.Second)
-		logging.Logger.Debug("Waiting for RDS to be deleted")
-	}
-
-	return RdsDeleteError{}
-}
 
 func getAllRdsInstances(session *session.Session, region string, excludeAfter time.Time) ([]*string, error) {
 	svc := rds.New(session)
@@ -76,10 +59,14 @@ func nukeAllRdsInstances(session *session.Session, names []*string) error {
 		}
 	}
 
+	timeout := 240 * time.Second
+	ctx := awsgo.BackgroundContext()
+	ctx, _ = context.WithTimeout(ctx, timeout)
+
 	if len(deletedNames) > 0 {
 		for _, name := range deletedNames {
 
-			err := waitUntilRdsDeleted(svc, &rds.DescribeDBInstancesInput{
+			err := svc.WaitUntilDBInstanceDeletedWithContext(ctx, &rds.DescribeDBInstancesInput{
 				DBInstanceIdentifier: name,
 			})
 
--- a/aws/rds.go
+++ b/aws/rds.go
@@ -1,7 +1,6 @@
 package aws
 
 import (
-	"context"
 	"time"
 
 	awsgo "github.com/aws/aws-sdk-go/aws"
@@ -59,14 +58,10 @@ func nukeAllRdsInstances(session *session.Session, names []*string) error {
 		}
 	}
 
-	timeout := 240 * time.Second
-	ctx := awsgo.BackgroundContext()
-	ctx, _ = context.WithTimeout(ctx, timeout)
-
 	if len(deletedNames) > 0 {
 		for _, name := range deletedNames {
 
-			err := svc.WaitUntilDBInstanceDeletedWithContext(ctx, &rds.DescribeDBInstancesInput{
+			err := svc.WaitUntilDBInstanceDeleted(&rds.DescribeDBInstancesInput{
 				DBInstanceIdentifier: name,
 			})
 
--- a/aws/rds.go
+++ b/aws/rds.go
@@ -1,0 +1,95 @@
+package aws
+
+import (
+  "time"
+
+  awsgo "github.com/aws/aws-sdk-go/aws"
+  "github.com/aws/aws-sdk-go/aws/awserr"
+  "github.com/aws/aws-sdk-go/aws/session"
+  "github.com/aws/aws-sdk-go/service/rds"
+  "github.com/gruntwork-io/cloud-nuke/logging"
+  "github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func waitUntilRdsDeleted(svc *rds.RDS, input *rds.DescribeDBInstancesInput) error {
+  for i := 0; i < 240; i++ {
+    _, err := svc.DescribeDBInstances(input)
+    if err != nil {
+      if awsErr, isAwsErr := err.(awserr.Error); isAwsErr && awsErr.Code() == "DBInstanceNotFound" {
+        return nil
+      }
+
+      return err
+    }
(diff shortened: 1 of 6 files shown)


aws/rds_types.go
```diff
@@ -0,0 +1,8 @@
+package aws
+
+type RdsDeleteError struct{}
```
yorinasub17: This error struct should probably have a parameter indicating which RDS instance failed (preferably ID). That way, users can investigate later what RDS instance failed to delete, whether it is systematic, or if `cloud-nuke` just didn't wait long enough.

aws/rds.go
```diff
@@ -0,0 +1,95 @@
...
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func waitUntilRdsDeleted(svc *rds.RDS, input *rds.DescribeDBInstancesInput) error {
+	for i := 0; i < 240; i++ {
```
yorinasub17: Would be good to add a comment here indicating how long this will wait for the RDS instance to be deleted.

aws/rds_types.go
```diff
@@ -0,0 +1,7 @@
...
+
+func (e RdsDeleteError) Error() string {
+	return "RDS was not deleted"
```
yorinasub17: If you incorporate the above, then this message should include the ID.

aws/rds.go
```diff
@@ -0,0 +1,95 @@
...
+	result, err := svc.DescribeDBInstances(&rds.DescribeDBInstancesInput{})
+
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	var names []*string
+
+	for _, database := range result.DBInstances {
+		//if excludeAfter.After(*database.InstanceCreateTime) {
```
yorinasub17: Should this still be commented out?

aws/rds.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	return names, nil
+}
+
+func nukeAllRdsInstances(session *session.Session, names []*string) error {
+	svc := rds.New(session)
+
+	if len(names) == 0 {
+		logging.Logger.Infof("No Relational Database Service to nuke in region %s", *session.Config.Region)
```
yorinasub17: NIT:

```suggestion
		logging.Logger.Infof("No RDS DB Instances to nuke in region %s", *session.Config.Region)
```

RDS (Relational Database Service) is the AWS service, and what we are nuking are the DB instances that are a part of that.

aws/rds.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil
+	}
+
+	logging.Logger.Infof("Deleting all Relational Database Services in region %s", *session.Config.Region)
+	var deletedNames []*string
```
yorinasub17: NIT: Go idiom is to typically initialize the variable inline with the declaration. That way, it is more explicit what is the initial value. You typically only declare variables if you use `if` blocks to initialize.

```suggestion
	deletedNames := []*string{}
```

aws/rds.go
```diff
@@ -0,0 +1,95 @@
...
+			SkipFinalSnapshot:    awsgo.Bool(true),
+		}
+
+		_, err := svc.DeleteDBInstance(params)
+
+		if err != nil {
+			logging.Logger.Errorf("[Failed] %s", err)
+		} else {
+			deletedNames = append(deletedNames, name)
+			logging.Logger.Infof("Deleted RDS: %s", *name)
```
yorinasub17: NIT:

```suggestion
			logging.Logger.Infof("Deleted RDS DB Instance: %s", awsgo.StringValue(name))
```

Note that it is better to use `awsgo.StringValue` instead of dereferencing the pointer directly because it accounts for nil gracefully.

aws/rds.go
yorinasub17: Looks like this is in several places, so be sure to update all references as well.

aws/rds.go
```diff
@@ -0,0 +1,95 @@
...
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func waitUntilRdsDeleted(svc *rds.RDS, input *rds.DescribeDBInstancesInput) error {
```
yorinasub17: Do you think it is possible to use [WaitUntilDBInstanceDeleted](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#RDS.WaitUntilDBInstanceDeleted) instead? I know the AWS wait functions can be finicky, but it might work here.

Note: not worth it for the v1, but probably should capture as a comment to indicate the possibility!

aws/rds.go
```diff
@@ -0,0 +1,95 @@
...
+	for i := 0; i < 240; i++ {
+		_, err := svc.DescribeDBInstances(input)
+		if err != nil {
+			if awsErr, isAwsErr := err.(awserr.Error); isAwsErr && awsErr.Code() == "DBInstanceNotFound" {
```
yorinasub17: NIT: I think AWS has constants in the SDK for the error codes. E.g in [the sdk docs](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/), if you search for this string, you will find a reference to `ErrCodeDBInstanceNotFoundFault`. Not sure where the exact import is though...

aws/rds_test.go
```diff
@@ -0,0 +1,92 @@
...
+	input := &rds.DescribeDBInstancesInput{
+		DBInstanceIdentifier: name,
+	}
+
+	for i := 0; i < 240; i++ {
+		instance, err := svc.DescribeDBInstances(input)
+		status := instance.DBInstances[0].DBInstanceStatus
+
+		// If SkipFinalSnapshot = false on delete, should also wait for "backing-up" also to finish
+		if *status != "creating" {
```
yorinasub17: NIT: Use `awsgo.StringValue`

aws/rds_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/rds"
+
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/cloud-nuke/util"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func waitUntilRdsCreated(svc *rds.RDS, name *string) error {
```
yorinasub17: Like the comment above about `waitUntilRdsDeleted`, possibly replaced with [WaitUntilDBInstanceAvailable](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#RDS.WaitUntilDBInstanceAvailable).

aws/rds_test.go
```diff
@@ -0,0 +1,92 @@
...
+		Engine:               awsgo.String("postgres"),
+		MasterUsername:       awsgo.String("gruntwork"),
+		MasterUserPassword:   awsgo.String("password"),
+	}
+
+	_, err := svc.CreateDBInstance(params)
+
+	if err != nil {
+		assert.Fail(t, err.Error())
+	}
```
yorinasub17: NIT: I know you got this from other code, but more recently we learned that we can replace this whole if condition with `require.NoError(t, err)`. The import for `require` is `"github.com/stretchr/testify/require"`.

aws/rds.go
marinalimeira: Looks like the import is just `rds.ErrCodeDBInstanceNotFoundFault`
https://github.com/aws/aws-sdk-go/blob/master/service/rds/examples_test.go#L79 :D

aws/rds.go
marinalimeira: It definitly does, maybe using `WaitUntilDBInstanceDeletedWithContext` that I can set timeouts.

aws/rds.go
marinalimeira: Thank you! I wasn't sure how to represent this.

aws/rds.go
marinalimeira: Yes because somehow it wasn't working, it was some type issue. I've added to the checklist.

marinalimeira:
Thank you for the review @yorinasub17 ! I push the simpler issues and I'll finish the others later, there are still a few improvements for this to be working from the cli.

aws/rds_test.go
marinalimeira: I don't know why but using this function it just never ends :( I'll leave with the custom one.

brikis98:
@marinalimeira Could you go through the existing comments on the PR and respond to / resolve them as appropriate? And update the TODO list in the README description? That will make follow-up review rounds easier.

aws/rds_types.go
marinalimeira: This type is only being used on a function at `rds_test`. Anyway I improved the error message used on the main function to return the instance's name.

yorinasub17: Ah gotcha! It still helps to include the instance ID though so that we can debug the test failure. With something like cloud-nuke tests where we constantly spin up and down RDS instances, it can be difficult to know which one was associated with which test failure.

aws/rds_test.go
yorinasub17: Ok sounds good! Can you add a comment indicating that you've tried to use that function, but ran into mysterious infinite loops? It would help set the context in the future when we come back here and someone has the same thought as me.

aws/rds_test.go
```diff
@@ -0,0 +1,89 @@
...
+	waitUntilRdsCreated(svc, &name)
+}
+
+func TestListRDS(t *testing.T) {
+	t.Parallel()
+
+	region := "us-east-1" //, err := getRandomRegion()
+	//if err != nil {
+	//assert.Fail(t, errors.WithStackTrace(err).Error())
+	//}
```
yorinasub17: NIT: Can you remove dead code?

aws/rds_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	session, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String(region)},
+	)
+
+	rdsName := "cloud-nuke-test" + util.UniqueID()
+	createTestRDSInstance(t, session, rdsName)
+
+	eds, err := getAllRdsInstances(session, time.Now())
```
yorinasub17: NIT: 
```suggestion
	rds, err := getAllRdsInstances(session, time.Now())
```

Can you also add a test that makes sure the time logic works? See https://github.com/gruntwork-io/cloud-nuke/blob/master/aws/ec2_test.go#L196 for example

aws/rds_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if err != nil {
+		assert.Failf(t, "Unable to fetch list of RDS DB Instances", errors.WithStackTrace(err).Error())
+	}
+
+	assert.Contains(t, awsgo.StringValueSlice(eds), strings.ToLower(rdsName))
+
+	nukeAllRdsInstances(session, eds)
```
yorinasub17: Can you call this in a `defer` right after the retrieval? That way, it will run even if the test assertions fail and we will clean it up.

yorinasub17:
Looks like the test failed with a segfault:

```
--- FAIL: TestListRDS (271.95s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xdce9cc]

goroutine 41 [running]:
testing.tRunner.func1(0xc420252690)
	/usr/local/go/src/testing/testing.go:711 +0x2d2
panic(0xfcfd40, 0x1afca50)
	/usr/local/go/src/runtime/panic.go:491 +0x283
github.com/gruntwork-io/cloud-nuke/aws.getAllRdsInstances(0xc420364700, 0xbf87323febcaca2e, 0x3f4bf6a4c0, 0x1b096e0, 0x6, 0xc42031e6c0, 0x15, 0x13b5cb8, 0xf)
	/go/src/github.com/gruntwork-io/cloud-nuke/aws/rds.go:26 +0x11c
github.com/gruntwork-io/cloud-nuke/aws.TestListRDS(0xc420252690)
	/go/src/github.com/gruntwork-io/cloud-nuke/aws/rds_test.go:76 +0x18e
testing.tRunner(0xc420252690, 0x11b9298)
	/usr/local/go/src/testing/testing.go:746 +0xd0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:789 +0x2de
```

marinalimeira:
@yorinasub17 hey, the tests are passing now, can you take a final look?

aws/rds_test.go
```diff
@@ -73,17 +73,19 @@ func TestListRDS(t *testing.T) {
...
-	nukeAllRdsInstances(session, eds)
+	defer func() {
+		nukeAllRdsInstances(session, rds)
 
-	rdsNames, err := getAllRdsInstances(session, time.Now().Add(1*time.Hour*-1))
+		rdsNames, _ := getAllRdsInstances(session, time.Now().Add(1*time.Hour))
 
-	assert.NotContains(t, awsgo.StringValueSlice(rdsNames), strings.ToLower(rdsName))
+		assert.NotContains(t, awsgo.StringValueSlice(rdsNames), strings.ToLower(rdsName))
+	}()
```
yorinasub17: The `defer` should actually happen earlier, right after the instance is created. I suspect you had trouble getting it to work because you want the reference to `rds`, which is returned by `getAllRdsInstances`.

To explain, `defer` is a golang keyword to get things to happen at  the end of the function, right before it exits. This works even if the function exits early due to a `t.Fatal` or `return` call. However, it is only scheduled when the execution pass reaches the `defer` routine. Since we always want to attempt to clean up the RDS instance we created, regardless of whether or not the `getAllRdsInstances` function is working, we want to schedule the routine immediately after the create call. But that means you need the `rds` variable to be set (for `nukeAllRdsInstances`) before the call to `getAllRdsInstances`. So how do you do that?

My suggestion would be to decouple the `nukeAllRdsInstances` call from `getAllRdsInstances`. That is, you want to call `nuke` on the RDS instance that you created, regardless of what `get` returns. So instead of nuking what is returned by get, you want the `create` call to return the rds instance it created and pass that to `nuke` in the `defer` call.

Does that make sense?

marinalimeira: Actually it doesn't make much sense to me, can we make a call?

PR #91, opened 2020-01-24
Move to go modules

### LINKED ISSUE #90: Move to go modules
Right now go is with go 1.13.6, many repos have been moved to go modules which is the future go toolchain. So suggesting move this repo to also use go modules and deprecate `dep`.

chenrui333:
fixes #90


yorinasub17:
This was merged as a part of #115

chenrui333:
Thanks @yorinasub17!!

PR #94, opened 2020-02-11
Add feature to delete RDS Clusters

marinalimeira:
This is a complement of #86.

The previous PR was deleting all types of RDS but the Aurora.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -359,9 +359,8 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		}
 		// End EKS resources
 
-		// RDS DB resources
+		// RDS DB Instances
 		dbInstances := DBInstances{}
-
 		if IsNukeable(dbInstances.ResourceName(), resourceTypes) {
 			instanceNames, err := getAllRdsInstances(session, excludeAfter)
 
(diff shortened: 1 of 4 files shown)


aws/rds_cluster_test.go
```diff
@@ -0,0 +1,63 @@
...
+		DBClusterIdentifier: awsgo.String(name),
+		Engine:               awsgo.String("aurora-mysql"),
+		MasterUsername:       awsgo.String("gruntwork"),
+		MasterUserPassword:   awsgo.String("password"),
+	}
+
+	_, err := svc.CreateDBCluster(params)
+	require.NoError(t, err)
+
+	//waitUntilRdsCreated(svc, &name)
```
robmorgan: Do we need to uncomment that?

aws/aws.go
```diff
@@ -374,7 +373,25 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 				resourcesInRegion.Resources = append(resourcesInRegion.Resources, dbInstances)
 			}
 		}
-		// End RDS DB resources
+		// End RDS DB Instances
+
+		// RDS DB Clusters
+		// These reference the Aurora Clusters, for the use it's the same resource (rds), but AWS
+		// has different abstractions for each.
+		dbClusters := DBInstances{}
```
yorinasub17: I think you need a different struct here so that the right nuking function gets called.

aws/rds_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func waitUntilRdsClusterDeleted(svc *rds.RDS, input *rds.DescribeDBClustersInput) error {
+	// wait up to 60 seconds
```
yorinasub17: I don't think 60 seconds is enough to wait for an RDS cluster to be deleted. This may be enough time for Aurora Serverless, but definitely not enough for regular Aurora. This should probably be more like 15 mins.

aws/aws.go
marinalimeira: Wops, it was calling the `nukeDBInstances` thanks!

aws/rds_cluster.go
marinalimeira: ✅

aws/rds_cluster_test.go
marinalimeira: No, it was remaining from copy + paste 😅

marinalimeira:
@yorinasub17 can you take a look again? :D

aws/rds_cluster.go
```diff
@@ -0,0 +1,97 @@
...
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func waitUntilRdsClusterDeleted(svc *rds.RDS, input *rds.DescribeDBClustersInput) error {
+	// wait up to 15 minutes
+	for i := 0; i < 900; i++ {
```
yorinasub17: NIT: 900 with 1 second sleep in between is likely to be very chatty when waiting this long. Should probably be switched to 90 and 10 second inbetween.

PR #97, opened 2020-03-02
Fix source of panic in tests

yorinasub17:
We've had a few test failures where cloud-nuke panics:

- https://circleci.com/gh/gruntwork-io/cloud-nuke/16853
- https://circleci.com/gh/gruntwork-io/cloud-nuke/16865
- https://circleci.com/gh/gruntwork-io/cloud-nuke/16860

This is bad, because panic stops all test threads, even the ones unrelated to the cause of panic, meaning that we skip any cleanup routines we have in testing. Normally, this isn't an issue as we have `cloud-nuke` to handle the cleanup. However, some of the tests create termination protected EC2 instances to test cloud-nuke's functionality to ignore those, and if we don't clean those up, then nothing will.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,9 @@ jobs:
       - checkout
       - attach_workspace:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
-      - run: go run main.go aws --older-than 1h --force
+      - run:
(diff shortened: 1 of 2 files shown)


yorinasub17:
Thanks for review! Merging this in now so I can release and we can start preventing this from happening.

PR #98, opened 2020-03-02
Add logic for paging in ECS cluster nuking capability

yorinasub17:
We weren't handling pagination in ECS service nuking before, so we weren't picking up the Fargate ECS services in our account, where we have more than 100 ECS clusters.


yorinasub17:
Thanks for review! Merging.

PR #99, opened 2020-03-03
Use t3.micro for testing, since t2.micro is largely deprecated in most regions

yorinasub17:
Everytime cloud nuke selects a new region (e.g `eu-north-1`), the tests fail because they don't support `t2.micro`. This fixes that by using `t3.micro` for testing.


yorinasub17:
Thanks for review! Merging this in now.

PR #100, opened 2020-03-06
Fix source of panic in RDS module

yorinasub17:
One of our recent nuking builds panicked when looking up RDS instances. Looking at the stack trace (https://circleci.com/gh/gruntwork-io/cloud-nuke/16971), it looks like the cause was an RDS instance with `InstanceCreateTime` set to `nil`. Not sure when this is `nil`, but nonetheless added the requisite guards to avoid it in the future.

Also ran `go fmt` and removed dead code (commented out import).


yorinasub17:
Thanks for review! Merging and releasing.

PR #96, opened 2020-02-29
Adding S3 nuke support

saurabh-hirani:
This PR adds support to delete S3 buckets as requested in https://github.com/gruntwork-io/cloud-nuke/issues/58 

It ensures that the AWS S3 list bucket operation - which lists all buckets irrespective of region - is not called multiple times for each region and uses cached entries as highlighted in the above issue discussion. 

Deletion also uses ```WaitUntilBucketNotExists``` to ensure bucket is deleted. It also supports edge cases where in bucket is deleted by some other means after listing but before deletion.

**Demo:**

1. Added s3 to list of resources

```
go run main.go aws --list-resource-types | grep s3

s3
```

2. Run S3 related tests:

```
go test -run S3 -v ./...

=== RUN   TestListS3Buckets
time="2020-02-29T22:52:35+08:00" level=info msg="Random region chosen: ap-northeast-1"
time="2020-02-29T22:52:42+08:00" level=info msg="Awaiting creation - Bucket: cloud-nuke-test-gnd7gn"
time="2020-02-29T22:52:47+08:00" level=info msg="Deleting - 1 S3 Buckets in region ap-northeast-1"
time="2020-02-29T22:52:47+08:00" level=info msg="Deleting - 1/1 - S3 Bucket: cloud-nuke-test-gnd7gn"
time="2020-02-29T22:52:48+08:00" level=info msg="Awaiting deletion - 1/1 - Bucket: cloud-nuke-test-gnd7gn"
time="2020-02-29T22:52:48+08:00" level=info msg="[OK] - 1/1 - Bucket: cloud-nuke-test-gnd7gn - deleted"
--- PASS: TestListS3Buckets (13.35s)
```

TestListS3Buckets - creates, verifies S3 bucket existence, deletes and verifies S3 bucket non existence. 

```
=== RUN   TestNukeS3Buckets
time="2020-02-29T22:52:52+08:00" level=info msg="Random region chosen: ca-central-1"
time="2020-02-29T22:52:55+08:00" level=info msg="Awaiting creation - Bucket: cloud-nuke-test-npymlv"
time="2020-02-29T22:52:56+08:00" level=info msg="Deleting - 1 S3 Buckets in region ca-central-1"
time="2020-02-29T22:52:56+08:00" level=info msg="Deleting - 1/1 - S3 Bucket: cloud-nuke-test-npymlv"
[...]

PR #101, opened 2020-04-01
Prepare #96 for merge

yorinasub17:
This is an extension of #96 to prepare for merging into our environment. Specifically, our accounts are not ready to consider S3 buckets for nuking, but we also don't want to block the PR on our inaction to prep. So this extends the PR with a stopgap solution to exclude `s3` from consideration.

The main commit to look at and review is 15f7288, which introduces the `--exclude-resource-type` CLI arg and updates the circleci config to make use of it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,13 @@ jobs:
       - attach_workspace:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
       - run:
-          command: go run main.go aws --older-than 1h --force
+          command: |
+            # We explicitly list the resource types we want to nuke, as we are not ready to nuke some resource types
+            # (e.g., S3)
+            go run main.go aws \
+              --older-than 1h \
+              --force \
+              --exclude-resource-type s3
           no_output_timeout: 1h
 
   nuke_sandbox:
@@ -61,7 +67,13 @@ jobs:
       - attach_workspace:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
       - run:
-          command: export AWS_ACCESS_KEY_ID=$SANDBOX_AWS_ACCESS_KEY_ID && export AWS_SECRET_ACCESS_KEY=$SANDBOX_AWS_SECRET_ACCESS_KEY && go run main.go aws --older-than 24h --force
+          command: |
+            export AWS_ACCESS_KEY_ID=$SANDBOX_AWS_ACCESS_KEY_ID
+            export AWS_SECRET_ACCESS_KEY=$SANDBOX_AWS_SECRET_ACCESS_KEY
+            go run main.go aws \
+              --older-than 24h \
+              --force \
+              --exclude-resource-type s3
           no_output_timeout: 1h
 
   deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
       - run:
           command: |
-            # We explicitly list the resource types we want to nuke, as we are not ready to nuke some resource types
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -51,7 +51,13 @@ jobs:
       - attach_workspace:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
       - run:
-          command: go run main.go aws --older-than 1h --force
+          command: |
+            # We explicitly list the resource types we want to nuke, as we are not ready to nuke some resource types
```
brikis98, line 55:
```suggestion
            # We explicitly list the resource types we want to nuke, as we are not ready to nuke some resource types in the AWS account we use at Gruntwork for testing (Phx DevOps)
```

.circleci/config.yml
```diff
@@ -61,7 +67,13 @@ jobs:
       - attach_workspace:
           at: /go/src/github.com/gruntwork-io/cloud-nuke
       - run:
-          command: export AWS_ACCESS_KEY_ID=$SANDBOX_AWS_ACCESS_KEY_ID && export AWS_SECRET_ACCESS_KEY=$SANDBOX_AWS_SECRET_ACCESS_KEY && go run main.go aws --older-than 24h --force
+          command: |
+            export AWS_ACCESS_KEY_ID=$SANDBOX_AWS_ACCESS_KEY_ID
+            export AWS_SECRET_ACCESS_KEY=$SANDBOX_AWS_SECRET_ACCESS_KEY
+            go run main.go aws \
```
brikis98, line 73: Add same comment here?

README.md
```diff
@@ -19,6 +19,7 @@ The currently supported functionality includes:
 - Deleting all ECS services in an AWS account
 - Deleting all EKS clusters in an AWS account
 - Deleting all RDS DB instances in an AWS account
+- Deleting all S3 buckets in an AWS account - except for buckets tagged with Key=cloud-nuke-excluded Value=true
```
brikis98, line 22: Seems like a `cloud-nuke-excluded` tag is a good general-purpose mechanism we should support globally across all resource types. Perhaps file an issue to track that and see if anyone wants to submit a PR to support it?

brikis98: Alternatively... Perhaps we could support an `--exclude-tag key=value` mechanism that excludes specific tags for all resource types that support tags? Either one makes sense to track in a GitHub issue.

aws/s3.go
```diff
@@ -0,0 +1,308 @@
...
+
+// hasValidTags checks if bucket tags permit it to be in the deletion list.
+func hasValidTags(svc *s3.S3, bucket *s3.Bucket) (bool, string) {
+	// Exclude deletion of any buckets with cloud-nuke-excluded tags
+	bucketTags, err := getS3BucketTags(svc, bucket.Name)
+	if len(bucketTags) > 0 {
+		for _, tagSet := range bucketTags {
+			key := strings.ToLower(tagSet["Key"])
+			value := strings.ToLower(tagSet["Value"])
+			if key == "cloud-nuke-excluded" && value == "true" {
```
brikis98, line 66: NIT: extract `cloud-nuke-excluded` into a `const`.

README.md
yorinasub17: Filed https://github.com/gruntwork-io/cloud-nuke/issues/103

aws/s3.go
yorinasub17: Done: 5486236

.circleci/config.yml
yorinasub17: Done: 5486236

yorinasub17:
Ok addressed all comments, and build is passing now, so going to merge this in. Thanks for your review!

Issue #88, opened 2020-01-22
Adding support for non-dedicated accounts


aaronsteers:
**Functional Description**

Is there any way to support nuking environments that do not have fully dedicated accounts? Similar to #65 which added support for specific resource types, I'd like to delete resources: 
 (A) when the resources have a matching tag, and/or
 (B) when the resource has an ID (ARN on AWS) which is contained within a specific list or tracked resource IDs

The latter case (B) would require that somewhere we are logging the IDs of resources which were created (not yet solved for) and would be needed in order to give an option for resources that do not support tags needed to implement (A). (For example: ECS clusters have an ARN but do not yet support tagging.)

Has this been considered or is there a way to implement this today? And if not, is it something you would consider in the roadmap and/or accept a PR for?

**Sample Workflow**

One option for an adapted workflow would look like:

1. During the terraform design process, we apply a uniquely generated tag to apply to all (taggable) resources.
2. For resources known to not be taggable (such as ECS) we log the ARNs into a text file or into state variables (or both)
3. Nuke called twice, once with an arg like `--nuke_tag=terraform-autotag-8fsCd` and once with an arg like `--resource_list=created_resources.txt`

**Additional Background**

The specific use case is that (mostly for training labs and POCs), we don't always have the luxury of being able to create new AWS accounts, but we still want to have confidence that we can ALWAYS successfully destroy an environment after we are finished with it.

Thanks!

brikis98:
This is not currently supported. Adding support for nuking resources with specific tags (for resources that support tags) makes sense, and a PR for that is welcome.

As for supporting nuking resources with specific IDs, that seems like a more esoteric use case. `cloud-nuke` is not a Terraform-specific tool, so I'm not a fan of building something into it to track IDs specific to Terraform functionality. Moreover, I'm not certain the approach would work in the general case. For example, to delete some resources, you must first delete all dependent resources: e.g., to delete a VPC, you must remove subnets; to remove subnets, you must remove attached security groups; and so on. So if you passed in the ID of just a VPC, would we automatically clean up the attached subnets and security groups, or only if you passed in the IDs for ALL of those? The former goes against the "delete only these IDs" concept, but the latter seems practically very hard to do, and unlikely to be used.

aaronsteers:
@brikis98 - Those are great points regarding the resource ID list, and I have to agree with you that the dependency-hierarchy problem would be a significant challenge.

[...]

Issue #38, opened 2018-11-14
Feature Request: protect resources from deletion based on a tag


kjhosein:
_Example use case:_
Let's say you have a sandbox/dev account and want to use `cloud-nuke` to keep it clean of old artifacts, but on the other hand have resources that you're actively using and that may also be even older than your time threshold. 

_Solution:_ 
Tag those resources with something like `'cloud-nuke'='protect'`

SeanFarrow:
@brikis98,

I'd be happy to work on this, as it's something that would be really useful and I have a bit of spare time.

A couple of questions though:
Should we have a separate method to nuke included/excluded tagged resources rather than using the existing one?
What sorts of tests would you want for this? I know you are running builds using circle CI so wonder what you do about creating/tearing down resources.

brikis98:
> Should we have a separate method to nuke included/excluded tagged resources rather than using the existing one?

Not sure I follow what you're referring to by "method." Do you mean a separate top-level CLI command? Or separate Go method? 

> What sorts of tests would you want for this? I know you are running builds using circle CI so wonder what you do about creating/tearing down resources.

Probably a test that launches a few resources, some with tags, some without, and makes sure cloud-nuke with various tag parameters cleans up the right ones.

SeanFarrow:
I mean a separate go method for nuking tagged resources.

brikis98:
> I mean a separate go method for nuking tagged resources.

I'd be tempted to say it makes sense to update the existing methods to take tags into account as well, as it should be possible to specify tags *and* other params (e.g., creation date).

SeanFarrow:
I agree, let me get a PR done in the next few days.

brikis98:
Thank you!

mastertinner:
Has there been any progress on this? I'd love to see the tag whitelist.

SeanFarrow:
Nothing has happened currently, I am in the process of a big house move, but will get to this when settled.

rmarable:
Has there been any update to this enhancement request?

SeanFarrow:
Not as yet, but there will be after Christmas.
I now have some time and need it for a project.
Thanks,
Sean.

aaronsteers:
I started a PR which would allow inclusion and exclusion by a tag. 

Contributions and feedback welcome: #109

hongil0316:
[...]

Issue #108, opened 2020-04-09
Add support for more complex matching rules


brikis98:
We have many clean-up use cases where we want to nuke all resources that match a specific pattern: e.g., nuke S3 buckets, IAM roles, and IAM profiles where the name matches one of a long list of specific regular expressions (i.e., those that come from our automated tests).

We should extend cloud-nuke with a way to filter resources using these match rules. Specifying them via CLI arguments is likely going to be inconvenient, so we may need some config format for this.

aaronsteers:
I totally agree this would be valuable. 

Related issue (not started): #88

rhoboat:
The first version of this is merged: #113 **Add config option for complex matching against s3 buckets**

marinalimeira:
@brikis98 was this addressed when adding the config file? Can we close this one?

brikis98:
Yup, closing.

PR #109, opened 2020-04-17
Filters by name and/or resource tag (WIP)

### LINKED ISSUE #103: Allow specifying exclusion tags
The new s3 nuking capability has a built in mechanism to ignore any buckets that have the tag `cloud-nuke-excluded`. We should consider extending this to other resources.

See https://github.com/gruntwork-io/cloud-nuke/pull/101#discussion_r402139387 for more context and suggestions.

aaronsteers:
Note: this is still a work in progress and I would invite others to contribute as well.

The PR would add four new CLI arguments:

* `resource-name-pattern`
* `exclude-resource-name-pattern`
* `require-resource-tag`
* `exclude-resource-tag`

Notes:
* Each pair of arguments is mutually exclusive.
* The name pattern filters would accept regex.
* The resource tag filters would accept either of the following:
    * a tag name passed by itself
        * _(i.e. `my-special-project-tag` would match `my-special-project-tag=1` as well as `my-special-project-tag=true`)_
    * a tag name followed by a colon and a regex pattern for matching values
        * _(i.e. `my-project-name:happ.*` would match `my-project-name=happiness` as well as `my-project-name=happenstance`)_

Resolves #103, resolves: #38, resolves: #88

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -176,7 +176,9 @@ func GetTargetRegions(enabledRegions []string, selectedRegions []string, exclude
 }
 
 // GetAllResources - Lists all aws resources
-func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTypes []string) (*AwsAccountResources, error) {
+func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTypes []string,
+	resourceNamePattern string, excludeResourceNamePattern string,
+	resourceTag string, excludeResourceTag string) (*AwsAccountResources, error) {
 	account := AwsAccountResources{
 		Resources: make(map[string]AwsRegionResource),
 	}
@@ -204,7 +206,11 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		// ASG Names
 		asGroups := ASGroups{}
 		if IsNukeable(asGroups.ResourceName(), resourceTypes) {
-			groupNames, err := getAllAutoScalingGroups(session, region, excludeAfter)
+			groupNames, err := getAllAutoScalingGroups(
+				session, region, excludeAfter,
+				resourceNamePattern, excludeResourceNamePattern,
+				resourceTag, excludeResourceTag
+			)
 			if err != nil {
 				return nil, errors.WithStackTrace(err)
 			}
@@ -218,7 +224,11 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		// Launch Configuration Names
 		configs := LaunchConfigs{}
 		if IsNukeable(configs.ResourceName(), resourceTypes) {
-			configNames, err := getAllLaunchConfigurations(session, region, excludeAfter)
+			configNames, err := getAllLaunchConfigurations(
+				session, region, excludeAfter,
+				resourceNamePattern, excludeResourceNamePattern,
+				resourceTag, excludeResourceTag
+			)
 			if err != nil {
 				return nil, errors.WithStackTrace(err)
 			}
@@ -232,7 +242,11 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		// LoadBalancer Names
 		loadBalancers := LoadBalancers{}
 		if IsNukeable(loadBalancers.ResourceName(), resourceTypes) {
-			elbNames, err := getAllElbInstances(session, region, excludeAfter)
+			elbNames, err := getAllElbInstances(
+				session, region, excludeAfter,
+				resourceNamePattern, excludeResourceNamePattern,
+				resourceTag, excludeResourceTag
+			)
 			if err != nil {
 				return nil, errors.WithStackTrace(err)
 			}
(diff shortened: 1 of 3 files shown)


aaronsteers:
@yorinasub17 and @brikis98 - You were both auto-added as reviewers. Before we continue with the implementation, could you kindly take a look at the spec and the CLI interface proposed? Any omissions/issues you can foresee with this approach?

Thanks!

aaronsteers:
I believe I am close on S3 buckets but if I understand the codebase correctly, we may need to write a separate function similar to `getS3BucketTags` [here](https://github.com/gruntwork-io/cloud-nuke/pull/109/files#diff-eb6be04d37bbd846d321a3a4d44dcff8R40) for all different resource types which support tags. Does that sound correct?

brikis98:
@aaronsteers Apologies for missing this. I think the `WIP` label made me ignore it entirely! I'm super booked this week, but will try to take a look in the next few days.

yorinasub17:
> Would this approach work for cleaning up, for example, test IAM roles in our accounts, without deleting the IAM roles we need to actually access those accounts?

I don't quite have the time to do a deep dive, but after a cursory look into the UX here, I think this will work for that use case, but it will be tedious to use to replace our internal script. The main issue is that we have different patterns for each of the resources, so a single global filter isn't really what we want, so we will end up calling `cloud-nuke` multiple times with different filters restricted to the relevant resources.

brikis98:
Ah, OK, yea, that's the key: this PR assumes we want all resources to be filtered using the same set of names/tags, whereas I imagine most real-world cases, you'd want one sets of names/tags for EC2 instances, another set of names/tags for IAM roles, another set for S3 buckets, and so on.

@aaronsteers Any thoughts on whether the CLI args could be accommodated to support that? Or is the only sane way to to move to a config file as described in #108?

aaronsteers:
@brikis98 - For my part, and for my intended use case, I would like this to be feasible with the CLI, even without a config file.

My own use case is such that I'm using Terraform for deployments, and I already have added the means to run the deployment with a single project-wide tag and also a single project-wide name prefix. I would recommend that for users of the tag-based or name-based filtering should adopt a similar strategy of adding a single tag to ALL resources they are deploying. This can be a "project_name=my-project-A" tag or similar, or it can be that all deployed resources have a common name prefix. (And ideally both are true.)

Other Details:
 - For resources which don't support a "name" but _do_ support "tags", we can check for a tag named "Name" and follow the rules according to the "Name" tag.
 - For resources which don't support tags, those would be reliant on a name-based match. I don't see any way around this but it's one of the resources I wanted to also support name matching.
   - **Q:**  Do we have a list of resources which don't support tags? I had thought ECS clusters were in this bucket of "untaggables" but on recent check, I see that they (now?) do support tags.

aaronsteers:
In regards to supporting multiple tags or name patterns, I think this approach does expect that the initial deployment is keeping in mind also the afterwards needed cleanup. In other words, because we know we need to be able to isolate and (confidently!) nuke the resources we are creating, we therefor are exercising an upfront naming strategy AND/OR tagging strategy. 

In an ideal world, as cloud-nuke usage would otherwise require, we'd be nuking the full account, but for due to constraints we are deploying side-by-side with other resources that should _not_ be nuked at the same time. (In my case the constraints are: limit to number of AWS accounts, and desire to reuse one account for multiple projects.) The only way I can see to have high level of confidence is to establish a naming or tagging strategy upfront which supports this later cleanup.

Does this sound reasonable for your use cases as well or would we need a more ad-hoc strategy with additional flexibility?

aaronsteers:
@yorinasub17 - what do you think about the above in regards to your use case? If not feasible to apply a single tag (or small number of tags) globally to everything deployed, would it be feasible to leverage an exclude-tag pattern instead?

yorinasub17:
My 2 cents are that I don't see anything wrong with supporting these global filters configured by the CLI in cloud-nuke, but I am not sure we will be using it internally, and would prefer to have the config file based approach proposed in https://github.com/gruntwork-io/cloud-nuke/pull/113 along the side.

I unfortunately don't have the time at the moment to explain why your approach doesn't quite work for us, but the tl;dr is that we have a mix of legitimate terraform generated resources, legitimate manually created resources over the years, and cruft resources we don't need created by test failures in our automated test suite for terraform modules we maintain spread across 100s of repos. We're passed the point where we can use tagging to distinguish all three without an expensive effort to audit all our resources, so what @brikis98 is referring to is our short term stop gap that is selecting just a small subset of resources that we know for sure are test related, which unfortunately translates to dozens of regex patterns for each resource as they come from different modules.

brikis98:
@aaronsteers Perhaps  the way to make this a bit more general is to allow specifying the resource type(s) for each name/tag pattern.

Examples:

```bash
# Delete all resources that have a name that starts with `test-`
cloud-nuke --resource-name-pattern 'test-*'

# Delete all EC2, IAM, and S3 resources that have a name that starts with `test-`
cloud-nuke --resource-name-pattern 'test-*':ec2,iam,s3
```

I'm not sure that's the prettiest syntax, but something of that sort would support the use case we have internally and still work from the CLI. In the future, if we add a config file, the exact same underlying logic would work with the config file too.

ina-stoyanova:
Just bumping this PR to see if it makes more sense to close it at this time? It seems like we can't proceed forward. Thoughts?

aaronsteers:
@ina-stoyanova and @brikis98 - I've closed this PR since it's stale at this point and should be reevaluated in the context of the config files mentioned above. I've moved onto other projects now but will reopen/revise if I see a value on future projects to do so.

Thanks for all your help!

ina-stoyanova:
thanks @aaronsteers!

PR #110, opened 2020-04-18
Refactored S3 test cases to remove repetition and use table drive tests

saurabh-hirani:
The current code for `s3_test.go` has the following issues:

- Test cases which should be subtests are added as individual tests.
- Testing var `t` is passed throughout the code and should be consolidated to be used only with tests.
- There is no support for table drive tests to make it easier to add more test cases.
- While we are testing only **list** and **nuke** operations for buckets, it is not apparent from the test cases. 

Refactored the code to fix the above issues. This has no impact on the current functionality as only the `s3_test.go` file has changed. This activity is also to ensure that adding cases for tagged and named deletions in the future is easier.


saurabh-hirani:
@yorinasub17 @brikis98 - can you please check and let me know if this looks okay? Thanks.

aws/s3_test.go
```diff
@@ -33,39 +33,69 @@ func TestMain(m *testing.M) {
...
 }
 
-// TestNukeS3BucketArgs represents arguments for testNukeS3BucketWrapper
-type TestNukeS3BucketArgs struct {
-	isVersioned       bool
-	checkDeleteMarker bool
-	objectCount       int
-	objectBatchsize   int
+// setAWSParams sets up common operations for nuke S3 tests.
+func setAWSParams() (AWSParams, error) {
```
brikis98, line 67: NIT: perhaps `newAWSParams`?

aws/s3_test.go
```diff
@@ -132,332 +163,277 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
+	return func(t *testing.T) {
+		// Please note that we are passing the same session that was used to create the bucket
+		// This is required so that the defer cleanup call always gets the right bucket region
+		// to delete
+		defer nukeAllS3Buckets(awsParams.awsSession, []*string{aws.String(bucket.name)}, 1000)
+
+		// Verify that - before creating bucket - it should not exist
+		//
+		// Please note that we are not reusing awsParams.awsSession and creating a random session in a region other
+		// than the one in which the bucket is created - this is useful to test the sceanrio where the user has
```
brikis98, line 205:
```suggestion
		// than the one in which the bucket is created - this is useful to test the scenario where the user has
```

aws/s3_test.go
```diff
@@ -132,332 +163,277 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
+		// AWS_DEFAULT_REGION set to region x but the bucket is in region y.
+		bucketNamesPerRegion, err := getAllS3Buckets(awsSession, time.Now().Add(1*time.Hour*-1), targetRegions, bucket.name, args.batchSize)
+		if args.shouldError {
+			if err == nil {
+				assert.Fail(t, "Did not fail for invalid batch size")
+			}
+			t.Logf("SUCCESS: Did not list buckets due to invalid batch size - %s", bucket.name)
```
brikis98, line 212: We should not be using `t.Logf` for logging.

aws/s3_test.go
```diff
@@ -132,332 +163,277 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
-}
 
-func TestList_EmptyS3Bucket_NoTags(t *testing.T) {
-	testListS3BucketsWrapper(t, []map[string]string{}, 10, true)
-}
+	var allTestCases []testCaseStruct
 
-func TestList_EmptyS3Bucket_WithoutFilterTag(t *testing.T) {
-	testListS3BucketsWrapper(t, []map[string]string{
+	testCases := []testCaseStruct{
```
brikis98, line 256: NIT: The Go idiom typically used for table-driven tests declares the struct and the tests all in one:

```go
var testCases = []struct {
	param1 string
	param2 string
}{
	{"aaa", "bbb"},
	{"ccc", "ddd"},
}
```

aws/s3_test.go
```diff
@@ -132,332 +163,277 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
 	awsSession, err := createSession("")
 	if err != nil {
-		assert.Failf(t, "Failed to create random session", errors.WithStackTrace(err).Error())
-	}
+		return nil, fmt.Errorf("Failed to create random session - %s", err.Error())
+	}
+
+	targetRegions := []string{awsParams.region}
+
+	return func(t *testing.T) {
```
brikis98, line 196: `t.Parallel()`?

aws/s3_test.go
```diff
@@ -132,332 +163,277 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
+			t.Logf("SUCCESS: Did not match bucket - %s", bucket.name)
 		}
-	}
+	}, nil
+}
 
-	if shouldMatch {
-		assert.Contains(t, bucketNamesPerRegion[s.region], aws.String(s.bucketName))
+// TestListS3Bucket tests listing S3 bucket operation
+func TestListS3Bucket(t *testing.T) {
```
brikis98, line 248: `t.Parallel()`? Each test should have this, unless we have a good reason not to run it in parallel?

aws/s3_test.go
```diff
@@ -33,39 +33,69 @@ func TestMain(m *testing.M) {
...
 
-// TestS3Bucket represents a test S3 bucket.
-type TestS3Bucket struct {
-	name        string
-	tags        []map[string]string
-	isVersioned bool
+// genTestsBucketName generates a test bucket name.
+func genTestBucketName() string {
+	// Call UniqueID twice because even if the nth test tries to reuse a name of the first test
+	// AWS S3 deletion operation might be in progress after nuking.
```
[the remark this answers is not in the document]
saurabh-hirani: Wanted to have a little bit more randomness but yes the comment isn't clear and not needed. Removed.

aws/s3_test.go
saurabh-hirani: Done.

aws/s3_test.go
saurabh-hirani: Done.

aws/s3_test.go
saurabh-hirani: Done.

aws/s3_test.go
saurabh-hirani: Done for `TestListS3Bucket` . For `TestNukeS3Bucket` as per - https://github.com/gruntwork-io/cloud-nuke/blob/7a99393a574793e7a55e16876bf0923ddc6c9845/aws/s3_test.go#L428 - as we need to generate tests both for versioned and non versioned buckets - we have to create list of tests first and then execute them.

aws/s3_test.go
saurabh-hirani: Done.

aws/s3_test.go
saurabh-hirani: Done.

saurabh-hirani:
@brikis98  @yorinasub17 - updated as per feedback.

aws/s3_test.go
```diff
@@ -132,332 +161,338 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
 	}
 
 	awsSession, err := createSession("")
 	if err != nil {
-		assert.Failf(t, "Failed to create random session", errors.WithStackTrace(err).Error())
+		return nil, fmt.Errorf("Failed to create random session - %s", err.Error())
 	}
 
-	defer nukeAllS3Buckets(s.awsSession, []*string{aws.String(s.bucketName)}, 1000)
+	if args.objectCount > 0 {
```
yorinasub17, line 349: This routine to setup the bucket should happen in the subtest function so that it can be done in parallel. Otherwise, we will be serially constructing each bucket for each test which will be slow.

aws/s3_test.go
```diff
@@ -132,332 +161,338 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
-		checkDeleteMarker: false,
-		objectCount:       10,
-		objectBatchsize:   3,
-	}, true)
-}
+	for _, bucketType := range []string{"NoVersioning", "Versioning"} {
+		isVersioned := false
+		if bucketType == "Versioning" {
+			isVersioned = true
+		}
```
yorinasub17, line 432: NIT:

```go
isVersioned := bucketType == "Versioning"
```

aws/s3_test.go
yorinasub17: Also, the error handling is a bit tricky in this style as the deletion happens in the subtest. Ideally, we want to always run the cleanup function if we get far enough to create the bucket. This style makes it tricky to implement that because we might be incentivized to cut the test short if this function returns an error, but the cleanup function is in the subtest function.

saurabh-hirani:
> This removes one layer of indirection and should help make it much easier to follow what is going on. What do you think?

Thanks - that would lead to cleaner code. Fixed it.

aws/s3_test.go
saurabh-hirani: This should be resolved as per the latest commit.

aws/s3_test.go
saurabh-hirani: Done.

saurabh-hirani:
@yorinasub17 - updated as per feedback.

saurabh-hirani:
@yorinasub17 - just wanted to check if the above looks good - I have updated the code as per your feedback.

yorinasub17:
Hi @saurabh-hirani apologies for the delay on this. I've been super booked this month and haven't had the time to look at this. I am hoping to be able to review this again next week.

aws/s3_test.go
```diff
@@ -33,39 +33,67 @@ func TestMain(m *testing.M) {
...
+		if err != nil {
+			return nil, err
+		}
+		logging.Logger.Debugf("Creating session in region - %s", region)
+	}
+	session, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String(region)},
+	)
+	return session, err
+}
```
yorinasub17, line 55: NIT: since the namespace is shared across the entire package, this should probably be put in a different place or renamed to something specific to s3 tests.

aws/s3_test.go
```diff
@@ -33,39 +33,67 @@ func TestMain(m *testing.M) {
...
+
+// AWSParams has AWS params info.
+type AWSParams struct {
 	region     string
 	awsSession *session.Session
 	svc        *s3.S3
-	bucketName string
 }
```
yorinasub17, line 62: NIT: rename to something that is specific to S3.

aws/s3_test.go
```diff
@@ -33,39 +33,67 @@ func TestMain(m *testing.M) {
...
 
-// genTestsBucketName generates a test bucket name.
-func genTestBucketName() string {
-	// Call UniqueID twice because even if the nth test tries to reuse a name of the first test
-	// AWS S3 deletion operation might be in progress after nuking.
-	return strings.ToLower("cloud-nuke-test-" + util.UniqueID() + util.UniqueID())
+// TestS3Bucket represents a test S3 bucket
+type TestS3Bucket struct {
+	name        string
+	tags        []map[string]string
```
yorinasub17, line 90: NIT: consider using the structs from aws sdk for type safety.

aws/s3_test.go
```diff
@@ -132,332 +161,331 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
-func setupNukeTests(t *testing.T) SetupInfo {
-	var s SetupInfo
+// testList3Bucket - helper function for TestListS3Bucket
+func testListS3Bucket(t *testing.T, args TestListS3BucketArgs) {
+	awsParams, err := newAWSParams()
 
-	region, err := getRandomRegion()
 	if err != nil {
-		assert.Fail(t, errors.WithStackTrace(err).Error())
+		assert.Failf(t, "Failed to setup AWS params", errors.WithStackTrace(err).Error())
```
yorinasub17, line 180: Use `require.Fail` or `require.NoError` to cut the test short. Otherwise, execution will continue.

aws/s3_test.go
```diff
@@ -132,332 +161,331 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
-		isVersioned:       true,
-		checkDeleteMarker: false,
-		objectCount:       10,
-		objectBatchsize:   1000,
-	}, true)
-}
+	allTestCases = append(allTestCases, testCaseStruct{
+		"EmptyBucket",
+		TestNukeS3BucketArgs{isVersioned: false, checkDeleteMarker: false, objectCount: 0, objectBatchsize: 0, shouldNuke: true},
+	})
```
yorinasub17, line 421: NIT: statically define the list like you did in the other test instead of using `append`, for consistency.

aws/s3_test.go
```diff
@@ -132,332 +161,331 @@ func (b TestS3Bucket) addObject(s SetupInfo, fileName string, fileBody string) e
...
-func TestNuke_S3Bucket_WithVersioning_BatchObjects(t *testing.T) {
-	testNukeS3BucketWrapper(t, &TestNukeS3BucketArgs{
-		isVersioned:       true,
-		checkDeleteMarker: false,
-		objectCount:       10,
-		objectBatchsize:   3,
-	}, true)
-}
+	for _, bucketType := range []string{"NoVersioning", "Versioning"} {
+		isVersioned := bucketType == "Versioning"
```
yorinasub17, line 424: NIT: use `true` and `false` directly in the range var.

aws/s3_test.go
yorinasub17: Thanks for resolving this!

saurabh-hirani:
> Apologies for the long review cycle this time. I had to put this off multiple times, but finally had the time to look. This is very close, but had a few style nits.
> 
> I am committed to moving this across the finish line in the next two weeks. Let me know if you are not able to handle the updates, and I can try to find time to apply them myself.

Thanks @yorinasub17 - will check out the changes today and reply back in the next 1-2 days.

aws/s3_test.go
saurabh-hirani: Done - added S3Test prefix where possible for consistency.

aws/s3_test.go
saurabh-hirani: This has been removed so not applicable.

aws/s3_test.go
saurabh-hirani: Checked and saw that there is no need for TestS3Bucket struct - easier to have functions which take args and instead of using types and getting into potential conflicts with AWS. `create` and `addObject` have been replaced by `S3TestCreateBucket` and `S3TestBucketAddObject`

aws/s3_test.go
saurabh-hirani: Thanks - wasn't aware of that - have used `require.NoError` - removes the if/else check also.

aws/s3_test.go
saurabh-hirani: Did that - had optimised in this case as EmptyBucket test does not need to deal with versioned/non-versioned flags but moved it in.

aws/s3_test.go
saurabh-hirani: Cannot do so as the tests need the Versioning v/s NonVersioning prefix 

```
--- PASS: TestNukeS3Bucket/NoVersioning_EmptyBucket (9.39s)
    --- PASS: TestNukeS3Bucket/NoVersioning_BatchObjects_InvalidBatchSize_Over (9.97s)
    --- PASS: TestNukeS3Bucket/Versioning_EmptyBucket (9.60s)
    --- PASS: TestNukeS3Bucket/Versioning_AllObjects (24.06s)
   ```

Having true/false in range will lead to reinterpreting true false and setting Versioning/nonVersioning

**Comment by saurabh-hirani** [comment]
@yorinasub17 updated as per feedback and added comments. Please check and let me know if this looks good.

**Comment by saurabh-hirani** [comment]
@yorinasub17 - can you please review the above changes and provide feedback if any more changes are required?

File: aws/s3_test.go (hunk 15 above)
Comment (yorinasub17) [reply aws/s3_test.go:383]: You could do:

```go
fmt.Sprintf("Versioning_%v_EmptyBucket", isVersioning)
```

But it's not a big deal so can keep this as is.

yorinasub17:
Apologies for the delay. I had been snoozing this and lost track of it amongst all the other things on my plate. I just took a pass and the updates look reasonable, so kicked off a build.

If that passes, we can go ahead and merge this in. Thanks!

yorinasub17:
Ok build passed! Will merge this in. Thanks for your contribution!

saurabh-hirani:
Thanks @yorinasub17 and @brikis98 for the feedback!

PR #113, opened 2020-05-05
Add config option for complex matching against s3 buckets

rhoboat:
Addresses https://github.com/gruntwork-io/cloud-nuke/issues/108

Purpose
-------
Allow terminating AWS resources via more complex matching rules via a config file. 

Scope
------
This PR is scoped to support nuking s3 buckets based on matching by name.

CLI changes
-------------
I propose we allow something like this:

```sh
cloud-nuke aws --config path/to/file.yaml
```

Yaml config format
-----------------
The structure of the config should reflect the way AWS describes resources. S3 buckets have names and regions, so for the scope of this project:

```yaml
s3:
  include:
    names_regex:
      - ^alb-alb-.*-access-logs$
      - ...
```

The proposed format at a meta level is this:
```
[RESOURCE_TYPE]:
  [include | exclude]:
    [FIELD_PLURAL][_regex]:
      - [ list of matchers ]
         OR
     [key]: [value]
```
where
  * `resource_type` can be `s3`, `iam_role`, etc. _We are punting on supporting `region` for now._
  * `include` has rules that a resource must match to be nuked
  * `exclude` has rules that a resource must not match to be nuked
  * matching key looks like:
    * `tags_regex`: would accept a list of regular expressions
    * `names_regex`: would accept a list of regular expressions
    * `regions` (assumes plaintext, exact matching)
  * for resources like `ec2_instance`, we need to support key-value maps
    * e.g. `Name: ^test_.*`

Looking Ahead
---------------
<details><summary>[expand]</summary>

Eventually, we might support something like:

```yaml
ec2_instance:
  include:
    tags_regex:
      Name: ^test_.*

iam_role:
  include:
    names_regex:
      - ^test_.*
      - ^cloud-nuke-TestListEksClusters-.*

s3:
  include:
    names_regex:
      - ^alb-alb-.*-access-logs$
```

The following would nuke _only_ s3 buckets matching the `names_regex` but that are not in `us-west-1`, if called with `cloud-nuke aws --resource-type s3 --config path/to/config.yaml`:

```yaml
s3:
  include:
    names_regex:
      - ^alb-alb-.*-access-logs$
  exclude:
    regions:
      - us-west-1
```

</details>

How to test locally
-------------------

<details><summary>[expand]</summary>

1. Clone this repo in your regular development directory (no need to use $GO_DIR)
1. Log into nuclear wasteland (if you need access we need to grant it to you)
1. Run some commands like:
    * `go run main.go aws --resource-type s3 --config config/mocks/s3_cleanup.yaml`
    * try the command with/out `resource-type` option
    * try the command with/out `config` option
1. Run tests against sandbox environment, locally:
    * ```sh
      cd {repo}/aws
      houston exec sandbox -- go test -v -run TestFilterS3
      ```

</details>



Summary of trial project
--------------------------

<details><summary>[expand]</summary>

Notably completed today:
- [x] Add `--config path/to/file.yaml` support to the command line.
- [x] Add support for a simple config yaml including s3 buckets by name, using regexp.
- [x] Confirm via manual testing that these following commands work:
    - `go run main.go aws`: nuke all s3 buckets
    - `go run main.go aws --config ./config/mocks/s3_include_names.yaml`: nuke only s3 buckets in all regions that match any of the regexps.
    - `go run main.go aws --config ./config/mocks/s3_include_names.yaml --region us-east-1`: nuke only s3 buckets in us-east-1 that match any of the regexps.

Notably not done:
- [x] Add test coverage in `config/config_test.go`
- [ ] Support other kinds of keys in the yaml config, like `exclude_names_regex`, `exclude_names`, and `include_names`.

Of course this is still far from complete, but might set some ground work for the future. There are still a few TODOs in the code that signal what might be worth checking. One caveat to mention: I haven't written `go` in years and was never an expert, so some of the conventions might be lost on me. (But that doesn't make it any less fun for me!)
</details>


yorinasub17:
> I'm not sure if the aws specifier is important to include, so could it be this?

IIRC, `urfave/cli` (the CLI framework we are using) is not ergonomic enough in the argument parsing such that if you do the latter option, you have to always provide it there before the subcommand. This may or may not be intuitive. E.g.:

```
cloud-nuke --config /path/to/config aws --region us-east-1
```

Given that, I think it would probably be better to repeat the args for each subcommand for CLI consistency.

archanid:
@yorinasub17 Updated the description to reflect that. Thanks for explaining why that's important.

yorinasub17:
> I propose we use a config format that is universal, consistent with whatever we might already be using elsewhere, and something users use a lot. Does yaml fit the bill?

Yaml makes sense to me!

> I propose we use the _filter to be clear, but I'm open to thoughts here.

I prefer being more explicit, e.g. `include_names_regex`. This makes it clear that:
- Each entry is a regex.
- The regex is for inclusion, and not exclusion. This also allows us to expand to include `exclude_names_regex` in the future.
- The regex matches on names.

yorinasub17:
NIT: `tag_filter` should be a map instead of a list of maps -

```
ec2_instance:
  tag_filter:
    Name: ^test_.*
```

README.md
```diff
@@ -124,6 +124,39 @@ If you want to check what resources are going to be targeted without actually te
 cloud-nuke aws --resource-type ec2 --dry-run
 ```
 
+### Specify optional config file
+
+For more granularity, you can pass in a config file to specify which resources to terminate using regular expressions.
+
+```shell
+cloud-nuke --config path/to/file.yaml
```
yorinasub17, line 132: NIT: make sure this is updated to reflect what you now have in the description.

README.md
```diff
@@ -124,6 +124,39 @@ If you want to check what resources are going to be targeted without actually te
...
+```
+
+For example, if you want to terminate S3 buckets whose name matches the given patterns, here's what your config file might look like:
+```yaml
+s3:
+  name_filter:
+    - ^alb-alb-.*-access-logs$
+    - .*-prod-alb-.*
+```
```
yorinasub17, line 141: This could use a bit more explanation. For example, is it an `and` or `or` expression? That is, does the name need to match all the regexes or only one?

Showing an example would probably help. Something like:

> Given this config, `cloud-nuke` will only nuke S3 buckets that match one of the provided regular expressions. So a bucket named `alb-alb-app-access-logs` would be deleted, but not a bucket named `my-s3-bucket`.

README.md
```diff
@@ -124,6 +124,39 @@ If you want to check what resources are going to be targeted without actually te
...
+
+If you want to terminate S3 buckets by region:
+```yaml
+s3:
+  region_filter:
+    - us-east-1
+    - us-west-1
+```
```
yorinasub17, line 149: Same as above. Having one extra sentence to clarify what this exactly means ("nuke s3 buckets in the regions `us-east-1` and `us-west-1`") would help here.

yorinasub17: Also, how does this work if you provide `--exclude-region` or `--region` in the CLI? Does it override? What if the CLI arg is a region that is not in the config (e.g. `--region eu-west-1` with this config)?

archanid:
> I prefer being more explicit, e.g. `include_names_regex`. This makes it clear that:
> 
>     * Each entry is a regex.
> 
>     * The regex is for inclusion, and not exclusion. This also allows us to expand to include `exclude_names_regex` in the future.
> 
>     * The regex matches on names.

So, the yaml format could be something like this:
```
[resource_type]:
  [include|exclude]_[field_plural]_[format]:
    - [ list of matchers ]
```

Maybe we assume plaintext matching if they only specify `include_names`? Or we require `include_names_plaintext`?

About the `ec2_instance` one, I don't think the `Name: ^test` makes sense, how about this?
```yaml
ec2_instance:
  include_tags_regex:
    - ^test_.*
```

yorinasub17:
> About the ec2_instance one, I don't think the Name: ^test makes sense, how about this?

Well, tags are key-value pairs so what you want to do is filter by tags that match keys and values. So you either need to be specific about matching the tag keys (which I think is what you were going for in the alternative proposal), or matching values of specific tags (which is what I was going for in my suggestion). Does that make sense?

yorinasub17:
> Maybe we assume plaintext matching if they only specify include_names? Or we require include_names_plaintext?

I think assuming `include_names` as plain text, exact matches is a reasonable assumption to make. 👍

archanid:
Okay, I get it now about the key-value pair matching, for tags. Updating the description now.

archanid:
Supporting `regions` as a high level "resource type" is a little bit inconsistent with the rest of the config format. I think we could hash that out later? Maybe regions are just special.

I think everything else feels consistent enough. Let me know if we want to move forward with implementing something for nuking s3 buckets by name.

yorinasub17:
+1 to focusing on just the s3 names filter and punting on supporting regions!

archanid:
@yorinasub17 Okay! Great. Updated the description with what we've discussed.

yorinasub17:
The updated description LGTM!

aws/aws.go
```diff
@@ -422,14 +423,29 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
...
+                matchedBuckets = append(matchedBuckets, aws.String(bucketName))
+              }
+            }
+          }
+
+          if len(configObj.S3FilterRule.IncludeNamesRE) > 0 {
+            resourcesCache["S3"][bucketRegion] = matchedBuckets
+          } else {
+            resourcesCache["S3"][bucketRegion] = bucketNamesPerRegion[bucketRegion]
+          }
```
yorinasub17, line 441: I think it would be cleaner to have this logic embedded in `getAllS3Buckets`, considering that there is already some filtering happening there (the logic with `excludeAfter` timestamp)

README.md
archanid: Good catch!

config/config.go
```diff
@@ -0,0 +1,60 @@
...
+  MatchingRule []string `yaml:"include_names_regex"`
+}
+
+type ConfigObj struct {
+  S3FilterRule filterRule `yaml:"s3"`
+}
+
+type filterRule struct {
+  IncludeNamesRE []*regexp.Regexp `yaml:"include_names_regex"`
+}
```
yorinasub17, line 27: NIT: the `yaml` tags are unnecessary if we aren't directly marshalling/unmarshalling from this object.

README.md
archanid: I propose that the options provided in as CLI options should take precedence over those provided in the config file. This is simlar to how passing in environment variables by prepending their assignment to the CLI command would override those variables in the CLI function.

I'll add this note in the README to clarify this! I think we still might want to discuss those interactions in more detail, though.

archanid:
@yorinasub17 Thanks so much for your comments! I don't have time to take care of those concerns at the moment. But they will be helpful for the future.

I've updated the PR description with wrap-up comments (collapsed in a Summary). :tada:

Thanks, all!

config/config.go
rhoboat: Does it not count as directly unmarshalling if we're unmarshalling into `rawConfig` [here](https://github.com/gruntwork-io/cloud-nuke/blob/2bd89a46ae9ffffd793fc406a42d2dfccd5e3a88/config/config.go#L57)?

aws/aws.go
rhoboat: Are you thinking it would make sense within `getBucketInfo` [here](https://github.com/rhoboat/cloud-nuke/blob/matching-s3-108/aws/s3.go#L208)?

yorinasub17: Ah yup that is a better place!

config/config.go
yorinasub17: That counts if `filterRule` was somehow nested into the `RawConfig` struct, but looking at the code, `ConfigObj` looks disjoint from the `RawConfig` struct. So you need the tags on `RawConfig` and `resourceType` (which are being unmarshalled from the yaml), but I don't think you need it for `ConfigObj` and `filterRule` (which are being manually constructed in the routine). Does that make sense?

rhoboat: Ah, yes, that makes sense. Removed the tags from the ConfigObj and descendents.

README.md
```diff
@@ -124,6 +124,62 @@ If you want to check what resources are going to be targeted without actually te
...
+Given this config, `cloud-nuke` will only nuke S3 buckets that match one of the provided regular expressions. So a bucket named `alb-app-access-logs` would be deleted, but not a bucket named `my-s3-bucket`.
+```yaml
+s3:
+  include:
+    names_regex:
+      - ^alb-.*-access-logs$
+      - .*-prod-alb-.*
+```
+
+[[Unimplemented!!]]
```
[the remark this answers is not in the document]
yorinasub17: I was imagining we can comment out this whole section and the next one on exclude regions:

```markdown
<!-- The following features are not implemented yet, and thus are commented out from the README. These should be restored when we implement the features.

[[Regions are not implemented!!]]

... content omitted for brevity ...

[[Unimplemented!!]]

Given this config, `cloud-nuke` will nuke  .....

-->
```

aws/aws.go
```diff
@@ -422,14 +423,55 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
...
+										// Some exclude rule wants to keep this bucket
+										excluded = true
+									}
+								}
+
+								// Only nuke the bucket if included and not excluded
+								if !excluded {
+									matchedBuckets = append(matchedBuckets, aws.String(bucketName))
+								}
+							}
```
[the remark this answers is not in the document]
rhoboat: - [ ] Explain in README how include and exclude work together
- [ ] Refactor

rhoboat:
- [ ] Add an integration test with the `getAllS3Buckets` function to make sure the filtering logic works

README.md
```diff
@@ -124,6 +124,97 @@ If you want to check what resources are going to be targeted without actually te
 cloud-nuke aws --resource-type ec2 --dry-run
 ```
 
+### Config file
+
+For more granularity, you can pass in a configuration file to specify which resources to terminate using regular expressions.
```
brikis98, line 129: NIT to fix minor ambiguity in sentence:

```suggestion
For more granularity, such as being able to specify which resources to terminate using regular expressions, you can pass in a configuration file.
```

brikis98: Also, consider adding a sentence that the config file is a new feature, and only a few types of more granular filters are supported, but we'll be adding more in the future, and PRs to help with this endeavor are very welcome!

aws/s3.go
```diff
@@ -253,10 +255,54 @@ func getBucketInfo(svc *s3.S3, bucket *s3.Bucket, excludeAfter time.Time, region
...
+		if includeBucketByREList(bucketName, includeNamesREList) {
+			shouldInclude = excludeBucketByREList(bucketName, excludeNamesREList)
+		}
+	} else if len(excludeNamesREList) > 0 {
+		shouldInclude = excludeBucketByREList(bucketName, excludeNamesREList)
+	} else {
+		shouldInclude = true
+	}
+
+	return shouldInclude
```
brikis98, line 282: NIT: I think this logic can be simplified down to:

```go
if len(includeNamesREList) > 0 && includeBucketByREList(bucketName, includeNamesREList) {
  return excludeBucketByREList(bucketName, excludeNamesREList)
} else if len(excludeNamesREList) > 0 {
  return excludeBucketByREList(bucketName, excludeNamesREList)
} else {
  return true
}

File: aws/s3.go [hunk 10]
```diff
@@ -253,10 +255,54 @@ func getBucketInfo(svc *s3.S3, bucket *s3.Bucket, excludeAfter time.Time, region
...
+	}
+	return false
+}
+
+func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
+	shouldInclude := true
+
+	for _, re := range reList {
+		if re.MatchString(bucketName) {
+			shouldInclude = false
```
brikis98, line 299: NIT: Could `return false` right here?

commands/cli.go
```diff
@@ -201,7 +217,7 @@ func awsNuke(c *cli.Context) error {
 	}
 
 	logging.Logger.Infof("Retrieving active AWS resources in [%s]", strings.Join(targetRegions[:], ", "))
-	account, err := aws.GetAllResources(targetRegions, *excludeAfter, resourceTypes)
+	account, err := aws.GetAllResources(targetRegions, *excludeAfter, resourceTypes, *configObj)
```
brikis98, line 220: Couldn't `configObj` be `nil` at this point if the user didn't specify one? If so, will this be a nil pointer dereference?

config/config.go
```diff
@@ -0,0 +1,84 @@
...
+
+// GetConfig - unmarshall the raw config file
+// and parse it into a config object.
+func GetConfig(filePath string) (*Config, error) {
+	var configObj Config
+
+	absolutePath, err := filepath.Abs(filePath)
+	if err != nil {
+		return &Config{}, err
```
brikis98, line 50: NIT: We typically do `return nil, err` in these cases.

config/config_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	return
+}
+
+func TestConfigS3_EmptyFilters(t *testing.T) {
+	configFilePath := "./mocks/s3_empty_filters.yaml"
+	configObj, err := GetConfig(configFilePath)
+
+	if err != nil {
```
brikis98, line 77: NIT: don't need this if statement. Can just call `require.NoError` directly.

commands/cli.go
yorinasub17: Suggestion for fix in the initial assignment:

```
	configObj := config.Config{}
	configFilePath := c.String("config")

	if configFilePath != "" {
		configObjPtr, err = config.GetConfig(configFilePath)

		if err != nil {
			return fmt.Errorf("Error reading config - %s - %s", configFilePath, err)
		}
                configObj = *configObjPtr
	}
```

aws/s3_test.go
```diff
@@ -458,3 +462,62 @@ func TestNukeS3Bucket(t *testing.T) {
...
+	configObj, err = config.GetConfig(testCase.args.configFilePath)
+
+	// Verify that only filtered buckets are listed
+	bucketNamesPerRegion, err := getAllS3Buckets(awsParams.awsSession, time.Now().Add(1*time.Hour), []string{awsParams.region}, "", 100, *configObj)
+	require.NoError(t, err, "Failed to list S3 Buckets")
+	require.Equal(t, len(testCase.args.matches), len(bucketNamesPerRegion[awsParams.region]))
+	require.Subset(t, aws.StringValueSlice(bucketNamesPerRegion[awsParams.region]), testCase.args.matches)
+
+	// TODO: this is happening too early
+	defer nukeAllS3Buckets(awsParams.awsSession, aws.StringSlice(bucketNames), 1000)
```
brikis98, line 522: This actually needs to be much earlier in the test. As it is now, if any of the `require.XXX` checks above fail, then it'll never even get to the `defer`, and this cleanup will never happen.

As for it happening too early, what makes you think so? What does the log output look like for the test?

rhoboat: I'm just rerunning this again, and it's nuking just fine! Disregard--maybe something else fixed it.

aws/s3_test.go
```diff
@@ -458,3 +462,57 @@ func TestNukeS3Bucket(t *testing.T) {
...
+		"alb-alb-234567-access-logs-" + S3TestGenBucketName(),
+		"tonico-prod-alb-access-logs-" + S3TestGenBucketName(),
+		"prod-alb-public-access-logs-" + S3TestGenBucketName(),
+		"stage-alb-internal-access-logs-" + S3TestGenBucketName(),
+		"stage-alb-public-access-logs-" + S3TestGenBucketName(),
+	}
+}
+
+// TestFilterS3Bucket_Config tests listing only S3 buckets that match config file
+func TestFilterS3Bucket_Config(t *testing.T) {
```
rhoboat, line 484: I originally wrote several integration tests against a set of buckets, testing different sets of filters (include, exclude, both together), but it will be a nightmare to maintain those tests. We know the filtering is parsing correctly based on the unit tests (config_test.go), so it makes sense to limit the integration test to the "both together" (include and exclude) case, which Is what I've done here.

Would like some feedback on the effectiveness of this kind of test.

aws/s3.go
rhoboat: This introduces a bug, actually, so not implementing the change. I will explain with data why this doesn't work.

rhoboat: So the reason the simplification is inaccurate is that, when `includeBucketByREList` returns false, we shouldn't ever check that same bucket with `excludeBucketByREList`. (I need to rename these methods, both of them are including buckets, but by either include regex lists or exclude regex lists.)

If the exclude rule ends up including a bucket that the include rule already excluded, that's a bug. So this expanded logic is for that purpose.

aws/s3_test.go
rhoboat: I'm going to add the exclude-only and include-only integration tests back in because it will help with any changes in code review. I've concluded that we probably need them for confidence.

aws/s3_test.go
```diff
@@ -495,26 +497,67 @@ func TestFilterS3Bucket_Config(t *testing.T) {
 		require.NoErrorf(t, err, "Failed to create test bucket - %s", bucketName)
 	}
 
+	awsSession, err := S3TestCreateNewAWSSession("")
+	require.NoError(t, err, "Failed to create random session")
+
 	// Clean up test buckets
-	defer nukeAllS3Buckets(awsParams.awsSession, aws.StringSlice(bucketNames), 1000)
+	// defer nukeAllS3Buckets(awsParams.awsSession, aws.StringSlice(bucketNames), 1000)
```
rhoboat, line 504: @brikis98 With this line commented, no buckets are deleted, the test cases in here all pass.

With it uncommented, the buckets are deleted, here's the output:

image

README.md
rhoboat: - [ ] Also, consider adding a sentence that the config file is a new feature, and only a few types of more granular filters are supported, but we'll be adding more in the future, and PRs to help with this endeavor are very welcome!

commands/cli.go
rhoboat: - [ ] TODO

config/config.go
rhoboat: - [ ] TODO

config/config_test.go
rhoboat: - [ ] TODO

aws/s3_test.go
rhoboat: Moving defer into the `t.Run` block seems to work.

rhoboat:
@brikis98 @yorinasub17 Could use another review, possibly someone to pull and run this, too. I noticed we can't delete s3 buckets that have no region, such as the two buckets that are currently stuck in our nuclear wasteland account. Even the AWS UI doesn't let me delete them.

zackproser:
> @brikis98 @yorinasub17 Could use another review, possibly someone to pull and run this, too. I noticed we can't delete s3 buckets that have no region, such as the two buckets that are currently stuck in our nuclear wasteland account. Even the AWS UI doesn't let me delete them.

Happy to guinea pig this at some point. I haven't used cloud-nuke yet so it would be good learning for me. LMK if there are specific configs / accounts you'd like flexed!

rhoboat:
@zackproser I'm using the nuclear wasteland account, and if you look at s3_test.go, search for `Filter` you'll find a test that runs through 3 different types of mock config files. The config_test.go file also references a bunch of mock config files in the config/mocks dir. Look through there to see what kinds of things I'm testing.

I'd love if we found a config format that actually breaks functionality.

README.md
```diff
@@ -124,6 +124,99 @@ If you want to check what resources are going to be targeted without actually te
...
+
+##### s3 resource type:
+_Note: the fields without `_regex` suffixes refer to support for plain-text matching against those fields._
+
+| field       | include | exclude |
+|-------------|---------|---------|
+| names       | none    | none    |
+| names_regex | ✅      | ✅      |
+| tags        | none    | none    |
+| tags_regex  | none    | none    |
```
yorinasub17, line 218: Nice, clear docs and table! I like the use of the checkmarks here! Will be shamelessly stealing that in the next table I make.

aws/s3_test.go
```diff
@@ -361,8 +362,11 @@ func testNukeS3Bucket(t *testing.T, args TestNukeS3BucketArgs) {
 		return
 	}
 
+	var configObj *config.Config
+	configObj, err = config.GetConfig("../config/mocks/s3_include_names.yaml")
```
yorinasub17, line 366: NIT: for good test hygiene, there should be a `require.NoError(t, err)`.

aws/s3_test.go
```diff
@@ -458,3 +462,103 @@ func TestNukeS3Bucket(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		// Capture the range variable as per https://blog.golang.org/subtests
+		// Not doing this will lead to tc being set to the last entry in the testCases
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			// Clean up test buckets
+			defer nukeAllS3Buckets(awsParams.awsSession, aws.StringSlice(bucketNames), 1000)
```
yorinasub17, line 550: Doesn't this introduce a race condition where depending on how fast the tests go, one test could nuke buckets needed for another test? Also, could this be "wasteful" in that we call nuke multiple times?

I'm guessing that this works fine when running this test in isolation, but I wonder if our CI build will encounter the race condition when it is running all the tests and has many things going on, causing more context switching.

Consider using the test group pattern to run `nukeAllS3Buckets` once when all the parallel tests finish:

```go
defer nukeAllS3Buckets()
t.Run("group", func(t *testing.T) {
  for _, tc := range testCases {
    tc := tc
    t.Run(tc.name, func(t *testing.T) {
      t.Parallel()
      // test code
    })
  }
})
```

config/config_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+func TestConfig_Garbage(t *testing.T) {
+	configFilePath := "./mocks/garbage.yaml"
+	configObj, err := GetConfig(configFilePath)
+
+	require.NoError(t, err)
+
+	if !reflect.DeepEqual(configObj, emptyConfig()) {
+		assert.Fail(t, "Config should be empty, %+v\n", configObj)
+	}
```
yorinasub17, line 23: Not for this PR, but perhaps in the next iteration we should look into seeing if this can be thrown as an error? It would be useful to catch typos (e.g., if I mistyped `include`, I would want it to give me an error than move along as if it didn't exist!)

yorinasub17:
Looks like the CI build failed on `TestFilterS3Bucket_Config`, with error:

```
--- FAIL: TestFilterS3Bucket_Config (17.52s)
    --- FAIL: TestFilterS3Bucket_Config/Exclude (10.54s)
        s3_test.go:560: 
            	Error Trace:	s3_test.go:560
            	Error:      	Not equal: 
            	            	expected: 6
            	            	actual  : 7
            	Test:       	TestFilterS3Bucket_Config/Exclude
```

This runs in phxdevops, so perhaps there's a bucket in there that is matching?

aws/s3_test.go
yorinasub17: > I wonder if our CI build will encounter the race condition when it is running all the tests and has many things going on, causing more context switching.

It looks like it is not surefire to run into this given the most recent test run, but I think it is still in the realm of possibility that it might be worth adjusting this, either to the group model or adjusting the names list so that it only affects the buckets for that test.

config/config_test.go
rhoboat: I think this would be awesome. Any unparseable key in the config file should raise an informative error.

aws/s3_test.go
rhoboat: Should that `defer` be inside of the outer `t.Run`? Isn't the problem that it will nuke the buckets before the tests run?

rhoboat: What you have does work, but how? How is it that when I had a t.Run, just a single one in the for loop, with defer further up (just after the bucketNames were defined), it ran defer before the tests? But here it is still outside of a t.Run (two, in fact) and somehow runs after the tests as it should?

I guess the key is that `defer` somehow knows to wait for `t.Run` to return. When `t.Run` executes within a `for` loop, defer doesn't wait, unless the `for` loop is wrapped in a `t.Run`. Is each `t.Run` in the loop run in parallel, is that how `t.Parallel` works? The way the test-case code inside the innermost `t.Run` is written, it looks lexically like it's just running a single test-case in parallel (which doesn't make sense). But maybe `t.Parallel` somehow knows that the outer `for` loop it's in defines all the stuff that should run in parallel? If that's the case, then maybe `defer` can't know that those multiple `t.Run`s in the `for` loop are done, and _that's_ why we need to wrap the loop in another `t.Run`, which is easy for `defer` to wait for.... ???

I'll research myself, but those are my thoughts trying to make sense of how this works.

rhoboat:
When I run this locally against phxdevops there are 191 buckets matching `config/mocks/s3_exclude_names.yaml` so that can't be what's happening, right?

rhoboat:
Just fyi, we are running against the sandbox account, I believe. Unless these labels are wrong..

```
Using environment variables from project settings and/or contexts:
  ...
  SANDBOX_AWS_ACCESS_KEY_ID=**REDACTED**
  SANDBOX_AWS_SECRET_ACCESS_KEY=**REDACTED**
```

rhoboat:
This last test run passed. Rerunning.

rhoboat:
Running these tests locally against both phxdevops and sandbox, they all keep passing. There's probably some "random region" choice that fails, or something else I missed.

e.g.:
```sh
cd cloud-nuke/aws
houston exec sandbox -- go test -v -run TestFilterS3
houston exec phx -- go test -v -run TestFilterS3
```

zackproser:
> @zackproser I'm using the nuclear wasteland account, and if you look at s3_test.go, search for `Filter` you'll find a test that runs through 3 different types of mock config files. The config_test.go file also references a bunch of mock config files in the config/mocks dir. Look through there to see what kinds of things I'm testing.
> 
> I'd love if we found a config file that actually breaks functionality.
> 
> I have updated the description to describe how to pull and run and test this out!

Thanks for updating with testing instructions! Just a quick update to say the happy paths you specified are working for me and I had zero issues running this as described. I'll try to flex it a bit more as you suggest.

aws/s3_test.go
```diff
@@ -488,19 +493,34 @@ func bucketNamesForConfigTests() []string {
 func TestFilterS3Bucket_Config(t *testing.T) {
 	t.Parallel()
 
-	// Create test buckets
-	awsParams, err := newS3TestAWSParams()
+	// Create AWS session in ca-central-1
+	awsParams, err := newS3TestAWSParams("ca-central-1")
```
rhoboat, line 496: Hardcode the region for this test only.

aws/s3_test.go
```diff
@@ -488,19 +493,34 @@ func bucketNamesForConfigTests() []string {
...
+	// Nuke all buckets in ca-central-1 first
+	// passing in a config that matches all buckets
+	var configObj *config.Config
+	configObj, err = config.GetConfig("../config/mocks/s3_all.yaml")
+
+	// Verify that only filtered buckets are listed
+	cleanupBuckets, err := getAllS3Buckets(awsParams.awsSession, time.Now().Add(1*time.Hour), []string{awsParams.region}, "", 100, *configObj)
+	require.NoError(t, err, "Failed to list S3 Buckets in ca-central-1")
+
+	nukeAllS3Buckets(awsParams.awsSession, cleanupBuckets[awsParams.region], 1000)
```
rhoboat, line 508: Is there another way to do this other than 1. listing all the buckets, 2. nuke the listed buckets?

yorinasub17: AFAIK, this is the model of how cloud nuke works, so unfortunately, no this is the only way for now.

Issue #112, opened 2020-05-04
Incomplete Execution Due to Missing Exception Handling


int0x80:
Please add exception handling to continue execution when an existing control prevents successful deletion of a resource.

```
$ ./cloud-nuke aws --log-level debug
INFO[2020-05-04T14:29:33-05:00] The following resources types will be nuked:
INFO[2020-05-04T14:29:33-05:00] - ami
INFO[2020-05-04T14:29:33-05:00] - asg
INFO[2020-05-04T14:29:33-05:00] - ebs
INFO[2020-05-04T14:29:33-05:00] - ec2
INFO[2020-05-04T14:29:33-05:00] - ecsserv
INFO[2020-05-04T14:29:33-05:00] - eip
INFO[2020-05-04T14:29:33-05:00] - ekscluster
INFO[2020-05-04T14:29:33-05:00] - elb
INFO[2020-05-04T14:29:33-05:00] - elbv2
INFO[2020-05-04T14:29:33-05:00] - lc
INFO[2020-05-04T14:29:33-05:00] - rds
INFO[2020-05-04T14:29:33-05:00] - s3
INFO[2020-05-04T14:29:33-05:00] - snap
INFO[2020-05-04T14:29:37-05:00] Retrieving active AWS resources in [eu-north-1, ap-south-1, eu-west-3, eu-west-2, eu-west-1, ap-northeast-2, ap-northeast-1, sa-east-1, ca-central-1, ap-southeast-1, ap-southeast-2, eu-central-1, us-east-1, us-east-2, us-west-1, us-west-2]
INFO[2020-05-04T14:29:37-05:00] Checking region [1/16]: eu-north-1
ERRO[2020-05-04T14:29:37-05:00] *awserr.requestError AccessDenied: User: arn:aws:sts::111122223333:role/admin/cloud-nuke-test is not authorized to perform: autoscaling:DescribeAutoScalingGroups with an explicit deny
	status code: 403, request id: 103e6933-9797-474d-b391-55ebfeb1f88d
/go/src/github.com/gruntwork-io/cloud-nuke/aws/asg.go:18 (0x18138da)
/go/src/github.com/gruntwork-io/cloud-nuke/aws/aws.go:207 (0x1817f83)
/go/src/github.com/gruntwork-io/cloud-nuke/commands/cli.go:204 (0x1831eaa)
[...]

PR #114, opened 2020-05-13
Adding feature to ignore list and nuke errors

saurabh-hirani:
This PR adds the following features:

1. Support for --ignore-errors flag on command line as per the discussion in https://github.com/gruntwork-io/cloud-nuke/issues/112 

2. Updated ami.go to fix a bug which prints wrong number of deletions if ami deletion fails.

**Demo**:

1. Running cloud-nuke with `--ignore-errors all`:

```
cloud-nuke --ignore-errors all
.
.
INFO[2020-05-06T19:31:11+05:30] Checking region [1/1]: us-east-1
WARN[2020-05-06T19:31:12+05:30] Ignoring get resources error - asg - AccessDenied: User: arn:aws:sts::12345678:assumed-role/role-1/session-1 is not authorized to perform: autoscaling:DescribeAutoScalingGroups
        status code: 403, request id: xxxx
WARN[2020-05-06T19:31:12+05:30] Ignoring get resources error - lc - AccessDenied: User: arn:aws:sts::12345678:assumed-role/role-1/session-1 is not authorized to perform: autoscaling:DescribeLaunchConfigurations
        status code: 403, request id: xxxx
.
.
```
Complete command output - https://gist.github.com/saurabh-hirani/eaa9455488af5299ee5d07746a965f88 - does a best effort deletion. 

Detailed description as per the change in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,10 +40,20 @@ When executed as `cloud-nuke defaults-aws`, this tool deletes all DEFAULT VPCs a
 
 ## Install
 
+### Download from releases page
+
 1. Download the latest binary for your OS on the [releases page](https://github.com/gruntwork-io/cloud-nuke/releases).
 2. Move the binary to a folder on your `PATH`. E.g.: `mv cloud-nuke_darwin_amd64 /usr/local/bin/cloud-nuke`.
 3. Add execute permissions to the binary. E.g.: `chmod u+x /usr/local/bin/cloud-nuke`.
 4. Test it installed correctly: `cloud-nuke --help`.
+
+### Install via package manager
+
+Note that package managers are third party. The third party cloud-nuke packages may not be updated with the latest version, but are often close. Please check your version against the latest available on the [releases page](https://github.com/gruntwork-io/cloud-nuke/releases). If you want the latest version, the recommended installation option is to [download from the releases page](https://github.com/gruntwork-io/cloud-nuke/releases).
+
+- **macOS:** You can install cloud-nuke using [Homebrew](https://brew.sh/): `brew install cloud-nuke`. 
+
+- **Linux:** Most Linux users can use [Homebrew](https://docs.brew.sh/Homebrew-on-Linux): `brew install cloud-nuke`.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -217,6 +217,38 @@ _Note: the fields without `_regex` suffixes refer to support for plain-text matc
 | tags        | none    | none    |
 | tags_regex  | none    | none    |
 
+### Ignoring list/nuke errors for certain/all resources
+
+Ideally you should have AWS IAM permissions to list and nuke all target resources. If you do not have permissions to
+nuke resource X, you should exclude it using the `--exclude-resource-type` flag. However, if you don't know which resource
+types you are not permitted to nuke or you have list access for all but nuke access for some, you can ignore their
+errors by specifying the `--ignore-errors` flag:
+
+```shell
+cloud-nuke aws --ignore-errors ec2 --ignore-errors s3
+```
+
+`--ignore-errors` can take any resource type listed out by the `--list-resource-types` flag.
+
+This will ignore any errors encountered while listing/nuking EC2 and S3 resources. For example, if there are 3
+EC2 instances and 2 S3 buckets to nuke and few/all of the EC2 instances and few/all of the S3 bucket deletion fails, cloud-nuke
+will log those errors and proceed with deleting further resources.
+
+Please note that cloud-nuke deletes resource types in a certain order to avoid dependency errors. If there are resources
+that should be deleted before EC2 or S3, and there are errors encountered while listing/nuking those resources, then cloud-nuke
+will error out and fail before reaching to EC2 or S3. For example, cloud-nuke deletes autoscaling groups before EC2 instances
+and if your role does not have permissions to delete autoscaling groups and you added `ec2` to the `--ignore-errors` list but
+not added `asg`, then cloud-nuke will still end up exiting with an error around deleting those autoscaling groups. To avoid
+this and do an actual best effort deletion, do this:
+
+```shell
+cloud-nuke aws --ignore-errors all
+```
+
+This will log errors encountered while listing/nuking all resources and keep on marching ahead to see what it can nuke.
+
+**Note**: You should at least have `ec2:DescribeRegions` permissions as that is required to find out regions to nuke.
+
 ### Log level
 
 You can set the log level by specifying the `--log-level` flag as per [logrus](https://github.com/sirupsen/logrus) log levels:
--- a/aws/ami_test.go
+++ b/aws/ami_test.go
@@ -33,7 +33,10 @@ func waitUntilImageAvailable(svc *ec2.EC2, input *ec2.DescribeImagesInput) error
 
 func createTestAMI(t *testing.T, session *session.Session, name string) (*ec2.Image, error) {
 	svc := ec2.New(session)
-	instance := createTestEC2Instance(t, session, name, false)
+	instance, err := CreateTestEC2Instance(session, name, false)
+	if err != nil {
+		assert.Failf(t, "Could not create test EC2 instance", errors.WithStackTrace(err).Error())
+	}
 	output, err := svc.CreateImage(&ec2.CreateImageInput{
 		InstanceId: instance.InstanceId,
 		Name:       awsgo.String(name),
@@ -104,7 +107,11 @@ func TestListAMIs(t *testing.T) {
 
 	// clean up after this test
 	defer nukeAllAMIs(session, []*string{image.ImageId})
-	defer nukeAllEc2Instances(session, findEC2InstancesByNameTag(t, session, uniqueTestID))
+	instances, err := findEC2InstancesByNameTag(session, uniqueTestID)
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+	defer nukeAllEc2Instances(session, instances, true)
 
 	amis, err := getAllAMIs(session, region, time.Now().Add(1*time.Hour*-1))
 	if err != nil {
@@ -152,7 +159,11 @@ func TestNukeAMIs(t *testing.T) {
 	}
 
 	// clean up ec2 instance created by the above call
-	defer nukeAllEc2Instances(session, findEC2InstancesByNameTag(t, session, uniqueTestID))
+	instances, err := findEC2InstancesByNameTag(session, uniqueTestID)
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+	defer nukeAllEc2Instances(session, instances, true)
 
 	_, err = svc.DescribeImages(&ec2.DescribeImagesInput{
 		ImageIds: []*string{image.ImageId},
--- a/aws/ami_test.go
+++ b/aws/ami_test.go
@@ -111,7 +111,7 @@ func TestListAMIs(t *testing.T) {
 	if err != nil {
 		assert.Fail(t, errors.WithStackTrace(err).Error())
 	}
-	defer nukeAllEc2Instances(session, instances, true)
+	defer nukeAllEc2Instances(session, instances)
 
 	amis, err := getAllAMIs(session, region, time.Now().Add(1*time.Hour*-1))
 	if err != nil {
@@ -163,7 +163,7 @@ func TestNukeAMIs(t *testing.T) {
 	if err != nil {
 		assert.Fail(t, errors.WithStackTrace(err).Error())
 	}
-	defer nukeAllEc2Instances(session, instances, true)
+	defer nukeAllEc2Instances(session, instances)
 
 	_, err = svc.DescribeImages(&ec2.DescribeImagesInput{
 		ImageIds: []*string{image.ImageId},
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"math/rand"
+	"os"
 	"sort"
 	"strings"
 	"time"
@@ -15,6 +16,7 @@ import (
 	"github.com/gruntwork-io/cloud-nuke/logging"
 	"github.com/gruntwork-io/gruntwork-cli/collections"
 	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/sirupsen/logrus"
 )
 
 // OptInNotRequiredRegions contains all regions that are enabled by default on new AWS accounts
@@ -37,6 +39,46 @@ var OptInNotRequiredRegions = [...]string{
 	"us-east-2",
 	"us-west-1",
 	"us-west-2",
+}
+
+// CloudNukeAWSParams holds params for creating a new AWS session
+type CloudNukeAWSParams struct {
+	Region     string
+	AWSSession *session.Session
+}
+
+// NewCloudNukeAWSParams - creates CloudNukeAWSParams type
+func NewCloudNukeAWSParams(region string) (CloudNukeAWSParams, error) {
+	var awsParams CloudNukeAWSParams
+	if region == "" {
+		var err error
+		region, err = getRandomRegion()
+		if err != nil {
+			return awsParams, err
+		}
+		awsParams.Region = region
+		logging.Logger.Debugf("Creating session in region - %s", region)
+	} else {
+		awsParams.Region = region
+	}
+	session, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String(region)},
+	)
+	awsParams.AWSSession = session
+	return awsParams, err
+}
+
+// SetEnvLogLevel - sets log level from environment
+func SetEnvLogLevel() error {
+	logLevel := os.Getenv("LOG_LEVEL")
+	if len(logLevel) > 0 {
+		parsedLogLevel, err := logrus.ParseLevel(logLevel)
+		if err != nil {
+			return err
+		}
+		logging.Logger.Level = parsedLogLevel
+	}
+	return nil
 }
 
 func newSession(region string) *session.Session {
@@ -176,8 +218,24 @@ func GetTargetRegions(enabledRegions []string, selectedRegions []string, exclude
 	return targetRegions, nil
 }
 
+// GetAllResourcesArgs encapsulates arguments for GetAllResources
+type GetAllResourcesArgs struct {
+	TargetRegions          []string                    // target regions to nuke
+	ExcludeAfter           time.Time                   // exclude resources created after this time
+	NukeResourceTypes      []string                    // type of resources to nuke
+	IgnoreErrResourceTypes []string                    // ignore errors encountered while nuking resources of this type
+	RegionSessionMap       map[string]*session.Session // use pre-created per region sessions if passed in
+	ConfigObj              config.Config               // config object
+}
+
 // GetAllResources - Lists all aws resources
-func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTypes []string, configObj config.Config) (*AwsAccountResources, error) {
+func GetAllResources(args GetAllResourcesArgs) (*AwsAccountResources, error) {
+	targetRegions := args.TargetRegions
+	excludeAfter := args.ExcludeAfter
+	nukeResourceTypes := args.NukeResourceTypes
+	ignoreErrResourceTypes := args.IgnoreErrResourceTypes
+	regionSessionMap := args.RegionSessionMap
+
 	account := AwsAccountResources{
 		Resources: make(map[string]AwsRegionResource),
 	}
@@ -189,12 +247,15 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 	for _, region := range targetRegions {
 		logging.Logger.Infof("Checking region [%d/%d]: %s", count, totalRegions, region)
 
-		session, err := session.NewSession(&awsgo.Config{
-			Region: awsgo.String(region)},
-		)
-
-		if err != nil {
-			return nil, errors.WithStackTrace(err)
+		sess, ok := regionSessionMap[region]
+		if !ok {
+			var err error
+			sess, err = session.NewSession(&awsgo.Config{
+				Region: awsgo.String(region)},
+			)
+			if err != nil {
+				return nil, errors.WithStackTrace(err)
+			}
 		}
 
 		resourcesInRegion := AwsRegionResource{}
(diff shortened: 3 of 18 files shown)


brikis98:
Thanks for the PR! I'm super booked this week, but will try to take a look in the next few days.

README.md
```diff
@@ -124,6 +124,29 @@ If you want to check what resources are going to be targeted without actually te
 cloud-nuke aws --resource-type ec2 --dry-run
 ```
 
+### Ignoring list/nuke errors for certain/all resources
+
+Ideally you should have AWS IAM permissions to list and nuke all target resources. If you do not have permissions to
+nuke resource X - you should exclude it using the `--exclude-resource-type` flag. However, if you don't know which resource
+types you are not permitted to nuke or you have list acceess for all but nuke excess for some, you can ignore their
```
brikis98, line 131:
```suggestion
types you are not permitted to nuke or you have list access for all but nuke access for some, you can ignore their
```

README.md
```diff
@@ -124,6 +124,29 @@ If you want to check what resources are going to be targeted without actually te
...
+errors by specifying the `--ignore-errors` flag:
+
+```shell
+cloud-nuke aws --ignore-errors ec2
+```
+
+This will ignore any errorrs encountered while listing/nuking EC2 resources - please note that if there are resources that
```
brikis98, line 138: Please specify what other values you can pass here beyond `ec2` and how to handle multiple values.

brikis98: Also, does this ignore an EC2 error and continue nuking EC2 instances... Or does it stop nuking EC2 instances at the first EC2 error, but then keeps nuking other types of resources? We should make the behavior clear in the docs to avoid surprises.

aws/aws.go
```diff
@@ -481,8 +534,17 @@ func IsNukeable(resourceType string, resourceTypes []string) bool {
 	return false
 }
 
+// IsErrIgnorable - Checks if we should ignore errors for resource type or not
+func IsErrIgnorable(resourceType string, resourceTypes []string) bool {
+	if collections.ListContainsElement(resourceTypes, "all") ||
+		collections.ListContainsElement(resourceTypes, resourceType) {
+		return true
+	}
+	return false
```
brikis98, line 543:
```suggestion
	return collections.ListContainsElement(resourceTypes, "all") ||
		collections.ListContainsElement(resourceTypes, resourceType)
```

commands/cli.go
```diff
@@ -135,27 +152,26 @@ func awsNuke(c *cli.Context) error {
...
-	for _, resourceType := range resourceTypes {
-		if resourceType == "all" {
-			continue
-		}
-		if !aws.IsValidResourceType(resourceType, allResourceTypes) {
-			invalidresourceTypes = append(invalidresourceTypes, resourceType)
-		}
-	}
+	// Check command line resource type values
+	invalidResourceTypes := map[string][]string{}
```
brikis98, line 156: Why store these in a map?

README.md
saurabh-hirani: Fixed.

README.md
saurabh-hirani: Thanks for the feedback - clarified with example. It keeps on nuking other resource types - updated the README.

commands/cli.go
saurabh-hirani: Had kept them in a map in case we want to aggregate errors for list, exclude and ignore error types if user specifies invalid values. But there isn't a need of doing so at the moment - fixed it.

saurabh-hirani:
> Thanks for the PR!
> 
> Is there some way we can add automated tests to check this behavior?

I doubt if there is a straightforward way to do so. We are not checking for per resource type deletion errors - we are checking access related errors they return and deciding whether we want to move ahead or not. The way I tested it was to create a role that has permissions to create other roles - created a restricted role with it, applied that role's credentials to the cli, ran cloud-nuke and tried to nuke a resource not covered by the restricted role. 

Doing that in a test case would essentially be equal to retesting the above process. The way to introduce errors will lie not in the creation of the resource or its properties but in the role that tries to delete the resource. And to test if cloud-nuke goes ahead - we would essentially have to test multiple deletions while switching the creating and deleting roles. Does it make sense to do that? 

Honestly speaking - I was hoping this to be a feature which helps first time users get some best effort deletion work done and not fail if they encounter errors. I wouldn't use it if I know I have the right roles in place to run cloud-nuke.

README.md
```diff
@@ -124,6 +124,35 @@ If you want to check what resources are going to be targeted without actually te
 cloud-nuke aws --resource-type ec2 --dry-run
 ```
 
+### Ignoring list/nuke errors for certain/all resources
+
+Ideally you should have AWS IAM permissions to list and nuke all target resources. If you do not have permissions to
+nuke resource X - you should exclude it using the `--exclude-resource-type` flag. However, if you don't know which resource
```
brikis98, line 130:
```suggestion
nuke resource X, you should exclude it using the `--exclude-resource-type` flag. However, if you don't know which resource
```

README.md
```diff
@@ -124,6 +124,35 @@ If you want to check what resources are going to be targeted without actually te
...
+types you are not permitted to nuke or you have list access for all but nuke access for some, you can ignore their
+errors by specifying the `--ignore-errors` flag:
+
+```shell
+cloud-nuke aws --ignore-errors ec2 --ignore-errors s3
+```
+
+This will ignore any errorrs encountered while listing/nuking EC2 and S3 resources. For example, if there are 3
```
brikis98, line 138:
```suggestion
This will ignore any errors encountered while listing/nuking EC2 and S3 resources. For example, if there are 3
```

README.md
```diff
@@ -124,6 +124,35 @@ If you want to check what resources are going to be targeted without actually te
...
+EC2 instances and 2 S3 buckets to nuke and few/all of the EC2 instances and few/all of the S3 bucket deletion fails - cloud-nuke
```
brikis98, line 139:
```suggestion
EC2 instances and 2 S3 buckets to nuke and few/all of the EC2 instances and few/all of the S3 bucket deletion fails, cloud-nuke
```

README.md
```diff
@@ -124,6 +124,35 @@ If you want to check what resources are going to be targeted without actually te
...
+will log those errors and proceed with deleting further resources.
+
+Please note that cloud-nuke deletes resource types in a certain order to avoid dependency errors. If there are resources
+that should be deleted before EC2 or S3 - and there are errors encountered while listing/nuking those resources e.g.
+cloud-nuke deletes autoscaling groups before EC2 and if your role does not have permissions to delete autoscaling groups
+and you have not added `asg` to the `--ignore-errors` list, then cloud-nuke will error out and fail before reaching to EC2.
```
brikis98, line 145:
```suggestion
Please note that cloud-nuke deletes resource types in a certain order to avoid dependency errors. If there are resources
that should be deleted before EC2 or S3, and there are errors encountered while listing/nuking those resources, then cloud-nuke will error out and fail before reaching to EC2 or S3. For example,
cloud-nuke deletes autoscaling groups before EC2 instances, and if your role does not have permissions to delete autoscaling groups
and you added `ec2` to the `--ignore-errors` list but not `asg`, then cloud-nuke will still end up exiting with an error around deleting those autoscaling groups.
```

README.md
```diff
@@ -124,6 +124,35 @@ If you want to check what resources are going to be targeted without actually te
...
 
+### Ignoring list/nuke errors for certain/all resources
+
+Ideally you should have AWS IAM permissions to list and nuke all target resources. If you do not have permissions to
+nuke resource X - you should exclude it using the `--exclude-resource-type` flag. However, if you don't know which resource
+types you are not permitted to nuke or you have list access for all but nuke access for some, you can ignore their
+errors by specifying the `--ignore-errors` flag:
+
+```shell
+cloud-nuke aws --ignore-errors ec2 --ignore-errors s3
```
brikis98, line 135: We still need to provide users with the a pointer to what resource types can be specified via `--ignore-errors`. i.e., It's more than just `ec2` and `s3`, right?

brikis98:
> Doing that in a test case would essentially be equal to retesting the above process. The way to introduce errors will lie not in the creation of the resource or its properties but in the role that tries to delete the resource. And to test if cloud-nuke goes ahead - we would essentially have to test multiple deletions while switching the creating and deleting roles. Seems like a bigger overhead than the value it provides. Does it make sense to do that?

I actually think it does make sense to test this! In particular, I'd create automated tests for two scenarios:

1. Create an IAM role that _only_ has read permissions for EC2 and S3. Create an S3 bucket and an EC2 instance, assume the role, run `cloud-nuke` telling it to ignore errors, and make sure the run completes successfully, and that the S3 bucket and EC2 instance are still around (which the test can then clean up at the end).
1. Same as above, except this time, the IAM role also has write permissions for S3, so at the end of the test, the S3 bucket should be deleted, but the EC2 instance should still be around.

saurabh-hirani:
> 1. Create an IAM role that _only_ has read permissions for EC2 and S3. Create an S3 bucket and an EC2 instance, assume the role, run `cloud-nuke` telling it to ignore errors, and make sure the run completes successfully, and that the S3 bucket and EC2 instance are still around (which the test can then clean up at the end).
> 2. Same as above, except this time, the IAM role also has write permissions for S3, so at the end of the test, the S3 bucket should be deleted, but the EC2 instance should still be around.

Okay - will give it a shot and update this thread with findings.

saurabh-hirani:
@brikis98 - am able to try out the individual functions for creating/assuming role and 
creating/deleting policies for the same. I had the following queries before proceeding:

1. Having a file called `cli_nukeallresources_test.go` (more apparent in later queries)
besdies `cli_test.go`. But as we have to create test EC2 and S3 instances - can we reuse the
corresponding `createTestEC2Instance` in `aws/ec2_test.go` - that would mean uppercasing that func to export it and upper casing all its dependent funcs. `getRandomRegion` cannot be renamed as it is heavily used outside also. Or does it make sense to repeat the test EC2 creation code in the test case to avoid any dependencies? 

Same query for creating test S3 buckets but S3 is more complicated as its PR is still 
open - https://github.com/gruntwork-io/cloud-nuke/pull/110 

2. Having the following test functions in `cli_nukeallresources_test.go` - does this coverage
make sense?

- TestNukeAllResources_AllPerms() 
    - Creates test EC2, S3
    - No assume role
    - Call GetAllRresources - for resource type - EC2, S3 - returns list with test EC2 and S3
    - Call NukeAllResources with test EC2 and S3 - deletes them
    - Test EC2, S3 do not exist

- TestNukeAllResources_NoPerms()
    - Creates test EC2, S3
    - Creates test role with no read/write for EC2, S3 
    - Assume role
    - Call GetAllRresources - for resource type - EC2, S3 - returns errors
    - Test EC2, S3 still exist

- TestNukeAllResources_NoPerms_IgnoreErrors()
    - Same as above but GetAllResurces call has ignore errors flag set - so it doesn't fail  and returns empty list.
    - Test EC2, S3 still exist

- TestNukeAllResources_EC2S3_RORole()
    - Creates test EC2, S3
    - Creates test role with read for EC2, S3 
    - Assume role
    - Call GetAllRresources - for resource type - EC2, S3 - no errors - returns test EC2 and S3
    - Call NukeAllResources with test EC2 and S3 - returns errors
    - Test EC2, S3 still exist

- TestNukeAllResources_EC2S3_RORole_IgnoreErorors()
    - Same as above but NukeAllResources has ignore errors flag set - so it doesn't fail  and returns nil.
    - Test EC2, S3 still exist

- TestNukeAllResources_EC2_RORole_S3_RWRole()
    - Creates test EC2, S3
    - Creates test role with read for EC2, write for S3 
    - Assume role
    - Call GetAllRresources - for resource type - EC2, S3 - no errors - returns test EC2 and S3
    - Call NukeAllResources with test EC2 and S3 - fails at EC2 - does not go upto S3
    - Test EC2, S3 still exist

- TestNukeAllResources_EC2_RORole_S3_RWRole_IgnoreErrors()
    - Same as above but NukeAllResources has ignore errors flag set - so it doesn't fail 
        for EC2 and continues S3 deletion and returns nil.
    - Test S3 exists. Test EC2 deleted.

brikis98:
> Having a file called cli_nukeallresources_test.go (content fleshed out in the next query)
besides cli_test.go. But as we have to create test EC2 and S3 instances - can we reuse the
corresponding createTestEC2Instance in aws/ec2_test.go - that would mean uppercasing that func to export it. getRandomRegion cannot be renamed as it is heavily used outside also. Or does it make sense to repeat the test EC2 creation code in the test case to avoid any dependencies?

Feel free to move these functions to some package (e.g.,`test_util`) so they can be imported wherever needed.

> Having the following test functions in cli_nukeallresources_test.go - does this coverage
make sense?

That test coverage sounds great 👍 

One NIT: Go naming convention is CamelCase and not snake_case or Upper_Snake_case. So the function names should be `TestNukeAllResourcesEC2S3ReadOnlyRoleIgnoreErorors`... But otherwise, LGTM!

saurabh-hirani:
@brikis98 Updated as per feedback. As the config file PR also touched some common files - `cli.go`, `aws.go`, `s3_test.go` - had to rebase and the diffs didn't seem reviewable. So squashed the commits to review as one unit. Please review and let me know if this looks good. 

Summary:

- Added the following new files in `aws/`:
  - `ec2_utils.go` - moving out common test CRUD functions for ec2 - led to updating test files - `ec2_test.go` and dependent test files - `ebs_test.go, asg_test.go, ami_test.go, launch_config_test.go` .
  - `s3_utils.go` - same as above but for S3 - led to updating test files - `s3_test.go`
  - `iam_utils.go` - new file for CRUD for IAM roles/policies - doubles up as a base for future addition of IAM role/policy nuking

- Added the following new files in `commands/`:
  - `cli_nukeallresources_test.go` - contains test cases as described above - basically uses ec2/s3 utils for creating/reading/deleting resources and iam utils for creating/assuming/deleting roles. 

- Updated `aws.GetAllResources` and `aws.NukeAllResources` to accept session param because in order to assume role you have to pass in session param while deleting resources - which wasn't there. The argument list of these 2 functions was getting longer and hence used a struct to make it simpler + support optional args.

- Updated `ec2.go` to support flag which decides whether to wait or not for instance deletion after triggering it - https://github.com/gruntwork-io/cloud-nuke/pull/114/files#diff-6c5db3fd88049ee01a36ba713dfcdbdeR76 - Ec2 nuke to not wait for instance termination by default - default case is trigger and return. This is because when testing deletion repeatedly with around 8-10 EC2 instances - (encountered when deferred nukes run for assume role tests) - there were random timeouts + long delay sometimes for EC2 deletion. For dependent tests like `ebs_test.go` they wait for EC2 deletion e.g https://github.com/gruntwork-io/cloud-nuke/pull/114/files#diff-c57ef596e24478400f32b6312d13256eR183

- There is no support to pass in trigger-delete-and-return/wait-for-delete-complete of cloud-nuke resources. Can be taken up as separate feature request. 

- When running `s3_test.go` - the following tests failed for me - because as per https://github.com/gruntwork-io/cloud-nuke/pull/113#issuecomment-643485861 some test run in Gruntwork specific accounts e.g. phxdevops - and I am running them in my own account. Wanted to check if the corresponding tests should be rewritten to create test buckets so that they can run on non-Gruntwork AWS accounts also. 

```
--- FAIL: TestFilterS3Bucket_Config (30.65s)
    --- FAIL: TestFilterS3Bucket_Config/config_tests (0.00s)
        --- FAIL: TestFilterS3Bucket_Config/config_tests/Include (0.20s)
            s3_test.go:441:
                        Error Trace:    s3_test.go:441
                        Error:          Not equal:
                                        expected: 4
                                        actual  : 0
                        Test:           TestFilterS3Bucket_Config/config_tests/Include
        --- FAIL: TestFilterS3Bucket_Config/config_tests/IncludeAndExclude (0.22s)
            s3_test.go:441:
                        Error Trace:    s3_test.go:441
                        Error:          Not equal:
                                        expected: 3
                                        actual  : 0
                        Test:           TestFilterS3Bucket_Config/config_tests/IncludeAndExclude
        --- FAIL: TestFilterS3Bucket_Config/config_tests/Exclude (0.27s)
            s3_test.go:441:
                        Error Trace:    s3_test.go:441
                        Error:          Not equal:
                                        expected: 6
                                        actual  : 0
                        Test:           TestFilterS3Bucket_Config/config_tests/Exclude
```

saurabh-hirani:
@brikis98 @yorinasub17  This change has more testing code than the actual functionality :) - maybe the tests required very methodical series of steps (create role, assume role, create resources, try deletion, delete resources, delete role) to verify ignoring errors.

saurabh-hirani:
@brikis98 - can you please review this and check if this looks good? Thanks.

brikis98:
Sorry for the delay. It's a huge PR, so I keep snoozing it every time I see it 😁 

I'll try to get to it in the next week or so!

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+	}
+
+	// Assume role
+	var roleSession *session.Session
+	if assumeRoleArn == "" {
+		roleSession = awsParams.AWSSession
+	} else {
+		roleSession, err = aws.AssumeIAMRole(assumeRoleArn, awsParams.Region)
+		require.NoErrorf(t, err, "Failed to assume role - %s - %s", assumeRoleArn, err)
+	}
```
brikis98, line 199: Slightly simpler:

```suggestion
	roleSession := awsParams.AWSSession
	if assumeRoleArn != "" {
		roleSession, err = aws.AssumeIAMRole(assumeRoleArn, awsParams.Region)
		require.NoErrorf(t, err, "Failed to assume role - %s - %s", assumeRoleArn, err)
	}
```

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+		)
+		logging.Logger.Debug("Expected: Post get check: ec2 exists")
+	}
+	if collections.ListContains(args.getResourcesOutputTypes, "s3") {
+		require.Truef(
+			t, resourceExists(awsParams.Region, "s3", bucketName, account),
+			"Failed to find test S3 bucket - %s", bucketName,
+		)
+		logging.Logger.Debug("Expected: Post get check: s3 exists")
+	}
```
brikis98, line 237: Seems like you could loop over `args.getResourceOutputTypes` and do this check dynamically, rather than hard-coding ec2 and s3 checks?

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+			TargetRegions:          []string{awsParams.Region},
+			ExcludeAfter:           *excludeAfter,
+			NukeResourceTypes:      resourceTypes,
+			IgnoreErrResourceTypes: args.ignoreErrResourceTypes,
+			RegionSessionMap:       map[string]*session.Session{awsParams.Region: roleSession},
+		},
+	)
+	// Validate that get call should fail if specified
+	if args.getResourcesShouldFail {
+		require.Error(t, err, "GetAllResources did not fail")
```
brikis98, line 216: Hm, prob a good idea to check _which_ error you get here!

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+		)
+		logging.Logger.Debug("Expected: Post get check: s3 exists")
+	}
+
+	// Do assumed role nuke for target test resources
+	err = aws.NukeAllResources(buildNukeAllResourcesArgs(roleSession, resourceTypeIdentifierMap, awsParams.Region, args.ignoreErrResourceTypes))
+
+	// Validate that nuke call should fail if specified
+	if args.nukeResourcesShouldFail {
+		require.Error(t, err, "assumedRole NukeAllResources did not fail")
```
brikis98, line 244: Same here

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+		)
+		logging.Logger.Debug("Expected: Post nuke check - ec2 deleted")
+	}
+	if collections.ListContains(args.postNukeResources, "s3") {
+		require.Truef(t, s3Exists, "Test S3 bucket deleted when it should not have - %s", bucketName)
+		logging.Logger.Debug("Expected: Post nuke check - s3 exists")
+	} else {
+		require.Falsef(t, s3Exists, "Test S3 bucket not deleted - %s", bucketName)
+		logging.Logger.Debug("Expected: Post nuke check - s3 deleted")
+	}
```
brikis98, line 289: This could prob be done dynamically in a loop too?

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+					"s3:*"
+				],
+				"Resource": "*"
+			}
+		}`),
+	}
+}
+
+// TestNukeAllResourcesNoPerms tests deletion of test S3 and EC2 instance with an assumed
+// role which does not have S3/EC2 read/write permissions
```
brikis98, line 341:
```suggestion
// TestNukeAllResourcesNoPerms tests deletion of test S3 buckets and EC2 instances with 
// various IAM permissions: no access, read-only access, read/write access for some but 
// read-only for others, full read/write access.
```

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+func TestNukeAllResources(t *testing.T) {
+	t.Parallel()
+	// Create a top level AWS session object which will be used to create/destroy roles
+	// and resources. Specifically use us-east-1 as we are running >= 8 tests and vCpu limits
```
brikis98, line 345: Are the default limits for `us-east-1` any higher? Or is it just that you requested limited increases already for that region?

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+				},
+				"Action": "sts:AssumeRole"
+			}
+		]
+	}`)
+	assumeRolePolicyDocument := strings.Replace(
+		assumeRolePolicyDocumentTmpl, "$callerIdentityArn", callerIdentityArn, -1,
+	)
+
+	var testCases = []struct {
```
brikis98, line 371: Very nice table-driven tests 👍

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+				getResourcesShouldFail:  false,
+				getResourcesOutputTypes: []string{"ec2", "s3"},
+				nukeResourcesShouldFail: false,
+				postNukeResources:       []string{},
+			},
+		},
+		{
+			"NoPerms",
+			createIAMRoleArgs{
+				roleName:                 "cloud-nuke-test-1-noperms",
```
brikis98, line 393: Does this name get used exactly as-is? If so, this test may fail due to a name conflict if we have two commits that kick off the test suite in parallel. Could you ensure that the role name has a unique ID appended to it using `util.UniqueId()`.

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+				roleDescription:          "cloud-nuke-test role for NoPerms test",
+				assumeRolePolicyDocument: assumeRolePolicyDocument,
+			},
+			createIAMRolePolicyArgs{
+				roleName:       "cloud-nuke-test-1-noperms",
```
brikis98, line 398: Why does the policy need a role name?

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+	assumeRoleArn           string           // ARN of role to assume
+	ignoreErrResourceTypes  []string         // list of resource types to ignore AWS errors
+	getResourcesShouldFail  bool             // whether GetAllResources should return err
+	getResourcesOutputTypes []string         // what resourcetypes from test resources should GetAllResources return
+	nukeResourcesShouldFail bool             // flag which decides whether NukeAllResources should return err
+	postNukeResources       []string         // what resourcetypes from test resources should exist after NukeAllResources
+}
+
+// getIAMRolePolicyMamp maps a role type to a corresponding policy
+func getTestRolePolicyMap() map[string]string {
```
brikis98, line 304: Why use a map with string-based key lookup instead of a few named constants?

aws/ami_test.go
```diff
@@ -33,7 +33,10 @@ func waitUntilImageAvailable(svc *ec2.EC2, input *ec2.DescribeImagesInput) error
 
 func createTestAMI(t *testing.T, session *session.Session, name string) (*ec2.Image, error) {
 	svc := ec2.New(session)
-	instance := createTestEC2Instance(t, session, name, false)
+	instance, err := CreateTestEC2Instance(session, name, false)
+	if err != nil {
+		assert.Failf(t, "Could not create test EC2 instance", errors.WithStackTrace(err).Error())
```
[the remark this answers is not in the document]
saurabh-hirani: Have done that in the new files I added, but didn't do so in the old ones as they all were using `assert.Fail` and wanted to keep in sync. Ideally they should be using `require` where we want the test to not proceed further. But didn't want to do that level of yak shaving. Is that ok or should I use `require` for all the new code I added?

aws/ami_test.go
```diff
@@ -104,7 +107,11 @@ func TestListAMIs(t *testing.T) {
 
 	// clean up after this test
 	defer nukeAllAMIs(session, []*string{image.ImageId})
-	defer nukeAllEc2Instances(session, findEC2InstancesByNameTag(t, session, uniqueTestID))
+	instances, err := findEC2InstancesByNameTag(session, uniqueTestID)
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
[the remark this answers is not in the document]
saurabh-hirani: Used to checking `err` immediately after the call but I see your point. Would changing

```

	// clean up after this test
	defer nukeAllAMIs(session, []*string{image.ImageId})
	instances, err := findEC2InstancesByNameTag(session, uniqueTestID)
	if err != nil {
		assert.Fail(t, errors.WithStackTrace(err).Error())
	}
	defer nukeAllEc2Instances(session, instances, true)
```

to 

```

	// clean up after this test
	defer nukeAllAMIs(session, []*string{image.ImageId})
	instances, err := findEC2InstancesByNameTag(session, uniqueTestID)
	defer nukeAllEc2Instances(session, instances, true)
         if err != nil {
		assert.Fail(t, errors.WithStackTrace(err).Error())
	}
```

make sense?

aws/ec2_utils.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return ec2.Instance{}, gruntworkerrors.WithStackTrace(err)
+	}
+
+	return *runResult.Instances[0], nil
+
+}
+
+// CreateTestEC2Instance - creates a test EC2 instance
+func CreateTestEC2Instance(session *session.Session, name string, protected bool) (ec2.Instance, error) {
```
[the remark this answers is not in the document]
saurabh-hirani: This is no longer in `ec2_test.go` as it is moved to `ec2_utils.go` because `cli_nukeallresources_test.go` creates test S3 and test EC2 instance and we needed that logic to be moved out of test files as per. earlier discssion. So I moved the corresponding utility functions in `ec2_test.go` and `s3_test.go`. Going by the normal convention of creating an AWS resource or returning error if there is a failure. Cannot do an `assert` or a `require` in this method. Does that make sense or is there a better way around this?

aws/ec2.go
```diff
@@ -92,6 +93,11 @@ func nukeAllEc2Instances(session *session.Session, instanceIds []*string) error
 		return errors.WithStackTrace(err)
 	}
 
+	if !waitForTermination {
```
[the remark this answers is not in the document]
saurabh-hirani: Had added it to support async deletion. Reverted it back as it might be better tackled in a separate issue. Sometimes, we just want to trigger deletion and not wait for resources that take a long time to nuke.

aws/iam_utils.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+)
+
+func createSessionIfNotExists(awsSession *session.Session) *session.Session {
+	if awsSession == nil {
```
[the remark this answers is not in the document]
saurabh-hirani: Not required - this was to prevent creating session in the `cli_testnukeresources.go` but it is better done there and this doesn't read well.

aws/iam_utils.go
```diff
@@ -0,0 +1,236 @@
...
+		PolicyName:     aws.String(policyName),
+		RoleName:       aws.String(roleName),
+	}
+
+	_, err = svc.PutRolePolicy(putInput)
+	return err
+}
+
+// ValidateIAMRolePolicy validates an IAM role inline policy
+func ValidateIAMRolePolicy(awsSession *session.Session, roleName string, policyName string) error {
```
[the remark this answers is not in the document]
saurabh-hirani: Thanks for catching this - it was added earlier - not required as it is not being called. Will fix.

aws/iam_utils.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if !policyExists {
+		return fmt.Errorf("Role - %s - does not have policy - %s", roleName, policyName)
+	}
+
+	return nil
+}
+
+// DeleteIAMRolePolicy deletes IAM role policy
+func DeleteIAMRolePolicy(awsSession *session.Session, roleName string, policyName string) error {
```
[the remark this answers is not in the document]
saurabh-hirani: It should. Will fix.

aws/iam_utils.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return sess, err
+	}
+	creds = stscreds.NewCredentials(sess, roleArn)
+	sess.Config.Credentials = creds
+	return sess, nil
+}
+
+// TrimPolicyDocument converts newlines and spaces into a policyDocument format
+// required while creating role
```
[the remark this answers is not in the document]
saurabh-hirani: Yes. Will fix.

commands/cli_nukeallresources_test.go
```diff
@@ -0,0 +1,547 @@
...
+	os.Exit(exitVal)
+}
+
+func resourceExists(targetRegion string, targetResourceType string, targetResourceName string, account *aws.AwsAccountResources) bool {
+	for region, accountResources := range account.Resources {
+		if region != targetRegion {
+			continue
+		}
+		for _, resourceCollectionObj := range accountResources.Resources {
+			if resourceCollectionObj.ResourceName() == targetResourceType {
```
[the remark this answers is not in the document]
saurabh-hirani: Short answer: `ResourceName()` is a misnomer - it actually points to a resource's type. 


Long answer:

`accountResources` is of type 

```
type AwsAccountResources struct {
	Resources map[string]AwsRegionResource
}
```

where AwsRegionResource is 

```
type AwsRegionResource struct {
	Resources []AwsResources
}
```

and AwsResources is 

```
type AwsResources interface {
	ResourceName() string
	ResourceIdentifiers() []string
	MaxBatchSize() int
	Nuke(session *session.Session, identifiers []string) error
}
```

The `ResourceName` key in this type should ideally be called `ResourceType` because ResourceNames key in this is actually `ResourceIdentifiers`. All resources use `ResourceName` when they should mean `ResourceType` e.g. 

```

// ResourceName - the simple name of the aws resource
func (instance EC2Instances) ResourceName() string {
        return "ec2"
}
```

```

// ResourceName - the simple name of the aws resource
func (address EIPAddresses) ResourceName() string {
        return "eip"
}
```

commands/cli_nukeallresources_test.go
saurabh-hirani: Given the current codebase, I don't think we can check. For example `GetAllResources` calls get for all resource types, each of which returns `errors.WithStackTrace(err)` - how do we check specific error type in this case?

commands/cli_nukeallresources_test.go
saurabh-hirani: `NukeAllResources` has the same situation as `GetAllResources`

commands/cli_nukeallresources_test.go
saurabh-hirani: These are the default limits. I did not request an increase.

commands/cli_nukeallresources_test.go
saurabh-hirani: Because it is an inline policy attached to a role -https://docs.aws.amazon.com/IAM/latest/APIReference/API_PutRolePolicy.html

commands/cli_nukeallresources_test.go
saurabh-hirani: Makes sense - will do.

saurabh-hirani:
Thanks for the detailed feedback @brikis98 - I appreciate that. I have closed most of them and have put in queries where I needed some guidance. Smoke tests on `cli_nukeallresources_test.go` look good - will be testing other `_test.go` files tomorrow and will update this thread.

saurabh-hirani:
@brikis98 as per above ran tests for the following test files also - looks good 

```
ami_test.go
asg_test.go
ebs_test.go
ec2_test.go
s3_test.go
```

saurabh-hirani:
We can close this PR as it's review has been pending for a few months and it has deviated from the main codebase.

PR #115, opened 2020-05-16
Extend #91

yorinasub17:
This extends #91 and preps it for merge by updating the CircleCI scripts to be compatible with go modules.


yorinasub17:

PR #122, opened 2020-06-22
Optimising the look of the command outputs

priyansh19:
# Adding New Look and Feel of the CLI of the command execution

The new output will now look like :

Screenshot from 2020-06-22 14-30-34

Tests are also passing:

Screenshot from 2020-06-22 19-40-42

### DIFF
--- a/commands/cli.go
+++ b/commands/cli.go
@@ -186,14 +186,14 @@ func awsNuke(c *cli.Context) error {
 	}
 
 	// Log which resource types will be nuked
-	logging.Logger.Info("The following resource types will be nuked:")
+	logging.Logger.Info(" -> The following resource types will be nuked:")
 	if len(resourceTypes) > 0 {
 		for _, resourceType := range resourceTypes {
-			logging.Logger.Infof("- %s", resourceType)
+			fmt.Println("^ ->",resourceType)
 		}
 	} else {
 		for _, resourceType := range allResourceTypes {
-			logging.Logger.Infof("- %s", resourceType)
+			fmt.Println("* ->",resourceType)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


priyansh19:
@yorinasub17 Please review these changes :+1:

commands/cli.go
```diff
@@ -186,14 +186,14 @@ func awsNuke(c *cli.Context) error {
...
-	logging.Logger.Info("The following resource types will be nuked:")
+	logging.Logger.Info(" -> The following resource types will be nuked:")
 	if len(resourceTypes) > 0 {
 		for _, resourceType := range resourceTypes {
-			logging.Logger.Infof("- %s", resourceType)
+			fmt.Println("^ ->",resourceType)
 		}
 	} else {
 		for _, resourceType := range allResourceTypes {
-			logging.Logger.Infof("- %s", resourceType)
```
yorinasub17, line 196: This means that you can't suppress the logs using log leveling, which would be a regression. I think you can get the same view if you buffer the contents into a newline separated string and pass it in with `logging.Logger.Info`.

commands/cli.go
```diff
@@ -217,15 +217,15 @@ func awsNuke(c *cli.Context) error {
 		return errors.WithStackTrace(err)
 	}
 
-	logging.Logger.Infof("Retrieving active AWS resources in [%s]", strings.Join(targetRegions[:], ", "))
+	logging.Logger.Infof(" Retrieving active AWS resources in:\n[%s]", strings.Join(targetRegions[:], ", "))
```
yorinasub17, line 220:
```suggestion
	logging.Logger.Infof("Retrieving active AWS resources in:\n[%s]", strings.Join(targetRegions[:], ", "))
```

PR #123, opened 2020-06-25
First attempt at cloud-nuking ECS clusters

ina-stoyanova:
**Goal of the PR and the task:**
To update the cloud-nuke to delete ECS clusters with "older-than" filtering.

**How this PR will be updated:**
This PR is a **draft** PR at this stage. Please add any comments related to the goal and plans/tasks below as you wish - any feedback is welcome. 

First attempt at nuking ecs-clusters:
- [x] create empty ecs-cluster with `created_at` tag 
- [x] make sure I can tag and retireve tags for ecs cluster 
- [x] select the ecs cluster and if tag is present ~-> nuke~
- [x] get all ECS clusters with tag `first_seen`, or otherwise set it 
- [x] nuke all ECS clusters (from input list of ECS cluster ARNs)

2nd stage - if timestamp provided is before tag (as timestamp value) or after:
- [x] add timestamp to `first_seen` tag 
- [x] update methods to interpret timestamp
- [x] update filtering based on tag to be able to extract `older than input timestamp` ECS clusters

Tasks:
- [x]To re-read & update as makes sense [x] README initial state of change and additional functionality
- [x] Add ecs_cluster.go
- [x] Write tests for ecs_cluster.go
- [x] add a type def for clusters


.circleci/config.yml
```diff
@@ -55,7 +55,8 @@ jobs:
   deploy:
     <<: *defaults
     steps:
-      - checkout
+      - attach_workspace:
+          at: .
```
brikis98, line 59: Is this change intentional?

README.md
```diff
@@ -17,20 +17,33 @@ The currently supported functionality includes:
...
 - Deleting all ECS services in an AWS account
+- Deleting all ECS clusters in an AWS account
 - Deleting all EKS clusters in an AWS account
 - Deleting all RDS DB instances in an AWS account
 - Deleting all S3 buckets in an AWS account - except for buckets tagged with Key=cloud-nuke-excluded Value=true
 - Deleting all default VPCs in an AWS account
 - Revoking the default rules in the un-deletable default security group of a VPC
 
 ### Caveats
+- As of 2018-2019 AWS APIs seem to support tagging of ECS clusters, and therefore we've reviewed and re-focused on the support of nuking ECS clusters.
```
brikis98, line 28: Thx for updating! Once this PR is ready to merge, all these ECS cluster caveats can be removed!

aws/ecs_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// tagging an ECS cluster with a given tag key-value pair and a cluster ARN
```
brikis98, line 14: NIT: might be a bit clearer to say:

```suggestion
// Tag an ECS cluster identified by the given cluster ARN with a tag that has the given key and value
```

aws/ecs_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+		// if firstSeenTag is present
+		// check if cluster is older than t
+		// then include in filteredEcsClusters
+		//	else
+		// set first seen tag to time now
+
+		firstSeen := getClusterTag(awsSession, clusterArn, "first_seen")
+
+		if firstSeen == nil {
+			tagEcsCluster(awsSession, clusterArn, "first_seen", "tomorrow")
```
brikis98, line 58: Always handle the `error` returned by a method. Otherwise, it get silently swallowed, which will make debugging very tough. Note, most IDEs support warning you about this.

```suggestion
			_, err := tagEcsCluster(awsSession, clusterArn, "first_seen", "tomorrow")
			if err != nil {
			    return nil, err
			}
```

aws/ecs_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// tagging an ECS cluster with a given tag key-value pair and a cluster ARN
+func tagEcsCluster(awsSession *session.Session, clusterArn *string, tagKey string, tagValue string) (*ecs.TagResourceOutput, error) {
```
brikis98, line 15: NIT: Unless you needed the returned data from creating tags (rare), you can probably return _just_ the `error`.

aws/ecs_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+		//	else
+		// set first seen tag to time now
+
+		firstSeen := getClusterTag(awsSession, clusterArn, "first_seen")
+
+		if firstSeen == nil {
+			tagEcsCluster(awsSession, clusterArn, "first_seen", "tomorrow")
+		}
+	}
+	return nil, nil
```
brikis98, line 61: The for-loop above tags clusters that haven't been tagged before, but doesn't return any data?

aws/ecs_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+func getClusterTag(awsSession *session.Session, clusterArn *string, tagKey string) *string {
+	svc := ecs.New(awsSession)
+	input := &ecs.ListTagsForResourceInput{
+		ResourceArn: clusterArn,
+	}
+
+	clusterTags, err := svc.ListTagsForResource(input)
+	if err != nil {
+		logging.Logger.Errorf("Error getting the tags")
+		return nil
```
brikis98, line 94: Shouldn't you return the error here?

aws/ecs_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	for _, tag := range clusterTags.Tags {
+		if *tag.Key == tagKey {
```
brikis98, line 98: The AWS SDK uses a lot of pointers, and sometimes they are `nil`, which makes the risk of a `nil` dereference quite high. Therefore, they provide methods to "safely" read out values. I recommend using them when possible:

```suggestion
		if aws.StringValue(tag.Key) == tagKey {
```

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+func TestCanCreateAndListEcsCluster(t *testing.T) {
+	t.Parallel()
+
+	region := "eu-west-1"
+	awsSession, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String(region),
+	})
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
brikis98, line 23: NIT: the slightly cleaner way to do this:

```suggestion
	require.NoError(t, err)
```

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	clusterName := "test-ina-1"
```
brikis98, line 25: Since we run tests in parallel—both within a single test suite running on one computer, and many tests in various branches running in CircleCi—we need to ensure that names are unique. We have a convenient [`UniqueId`](https://github.com/gruntwork-io/cloud-nuke/blob/master/util/unique_id.go#L12) method for this:

```suggestion
	clusterName := fmt.Sprintf("test-ecs-cluster-%s", util.UniqueId())
```

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+	cluster := createEcsFargateCluster(t, awsSession, clusterName)
+	defer deleteEcsCluster(awsSession, cluster)
+
+	clusterArns, err := getAllEcsClusters(awsSession)
+	if err != nil {
+		assert.Failf(t, "Unable to fetch clusters: %s", err.Error())
+	}
```
brikis98, line 32:
```suggestion
	require.NoError(t, err)
```

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+	awsSession, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String(region),
+	})
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
+
+	// create a cluster without tag
+	cluster1 := createEcsFargateCluster(t, awsSession, "test-ina-3-1")
+	cluster2 := createEcsFargateCluster(t, awsSession, "test-ina-3-2")
```
brikis98, line 51: See comments above about using `UniqueId` in naming.

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+	defer deleteEcsCluster(awsSession, cluster1)
+	defer deleteEcsCluster(awsSession, cluster2)
```
brikis98, line 53: Put each `defer` immediately after the corresponding `createXXX`. Otherwise, if, for example, the first `create` worked, but the second failed, the code will never get to the `defer` methods for cleanup.

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// get cluster
+	clusterArns, err := getAllEcsClusters(awsSession)
+	if err != nil {
+		assert.Failf(t, "Unable to fetch clusters: %s", err.Error())
+	}
```
brikis98, line 59:
```suggestion
	require.NoError(t, err)
```

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	//tag with first_seen tag
+	for _, clusterArn := range clusterArns {
+		tagEcsCluster(awsSession, clusterArn, "first_seen", "ina")
```
brikis98, line 63: Same note here about handling the returned errors!

brikis98, line 63: NIT: extract `"first_seen"` and `"ina"` into variables so they can be used here and below w/o copy/paste.

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+func TestCanCreateWithTagAndFilterEcsClustersByTag(t *testing.T) {
+	t.Parallel()
+
+	region := "eu-west-1"
+	awsSession, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String(region),
+	})
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
brikis98, line 110:
```suggestion
	require.NoError(t, err)
```

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	clusterName := "test-ina-sep-2-with-tag"
```
brikis98, line 112: Same note on naming and `UniqueId`

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,130 @@
...
+	cluster := createEcsFargateCluster(t, awsSession, clusterName) //TODO - also have a separate function to do this so I can test it.
+	tagKey := "created_at"
+	tagValue := "now-inastime"
+	tag, err := tagEcsCluster(awsSession, cluster.ClusterArn, tagKey, tagValue) //should return
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
brikis98, line 119:
```suggestion
	require.NoError(t, err)
```

brikis98 (commented):
Looks like a promising start 👍 

Left a bunch of comments on idiomatic Go.

aws/ecs_cluster.go
ina-stoyanova: awesome - I did spend some time looking at this today, so was one of my questions to bring up at some point!

.circleci/config.yml
ina-stoyanova: no. not at all 👀

aws/ecs_cluster.go
ina-stoyanova: good point, thank you :)

aws/ecs_cluster.go
ina-stoyanova: about to expand on this method, so coming up soon!

zackproser (commented):
Looks like a great start, Ina :+1:

.circleci/config.yml
ina-stoyanova: This looks like it's been added by @yorinasub17 [in this commit](https://github.com/ina-stoyanova/cloud-nuke/commit/9b294286c6e55f436e18ed9064ab909a2e75ab7f)

I assume it's after a rebase on `uspstream/master, will revert anyway, as it's not related.

zackproser:
Hey @ina-stoyanova, 

I was playing around with this locally and ran into an error: 

```$ aws-vault exec zack-test-aws -- ./cloud-nuke aws --region us-east-1 --resource-type ecscluster
INFO[2020-10-07T16:39:30-04:00] The following resource types will be nuked:  
INFO[2020-10-07T16:39:30-04:00] - ecscluster                                 
INFO[2020-10-07T16:39:33-04:00] Retrieving active AWS resources in [us-east-1] 
INFO[2020-10-07T16:39:33-04:00] Checking region [1/1]: us-east-1             
tagKey: cloud-nuke-first-seen
tag.Key: cloud-nuke-first-seen
ERRO[2020-10-07T16:39:33-04:00] Error tagging the ECS cluster with ARN arn:aws:ecs:us-east-1:675304494746:cluster/zack-test-cluster 
ERRO[2020-10-07T16:39:33-04:00] Error getting the `first_seen` tag for ECS cluster with ARN arn:aws:ecs:us-east-1:675304494746:cluster/zack-test-cluster 
ERRO[2020-10-07T16:39:33-04:00] *time.ParseError parsing time "2020-10-07 20:13:29.422211915 +0000 UTC" as "2006-01-02T15:04:05Z07:00": cannot parse " 20:13:29.422211915 +0000 UTC" as "T"
/home/zachary/Gruntwork/cloud-nuke/aws/ecs_cluster.go:129 (0xd042e0)
	getClusterTag: return firstSeenTime, errors.WithStackTrace(err)
/home/zachary/Gruntwork/cloud-nuke/aws/ecs_cluster.go:51 (0xd03481)
	getAllEcsClustersOlderThan: firstSeenTime, err := getClusterTag(awsSession, clusterArn, firstSeenTagKey)
/home/zachary/Gruntwork/cloud-nuke/aws/aws.go:351 (0xcfb47f)
	GetAllResources: ecsClusterArns, err := getAllEcsClustersOlderThan(session, region, excludeAfter)
/home/zachary/Gruntwork/cloud-nuke/commands/cli.go:221 (0xd164e6)
	awsNuke: account, err := aws.GetAllResources(targetRegions, *excludeAfter, resourceTypes, configObj)
/home/zachary/go/pkg/mod/github.com/gruntwork-io/gruntwork-cli@v0.1.2/errors/errors.go:93 (0xa7a4de)
	WithPanicHandling.func1: return action(context)
/home/zachary/go/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:490 (0xa6aac8)
	HandleAction: return a(context)
/home/zachary/go/pkg/mod/github.com/urfave/cli@v1.20.0/command.go:210 (0xa6bcd8)
	Command.Run: err = HandleAction(c.Action, context)
/home/zachary/go/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:255 (0xa68d51)
	(*App).Run: return c.Run(context)
/home/zachary/go/pkg/mod/github.com/gruntwork-io/gruntwork-cli@v0.1.2/entrypoint/entrypoint.go:21 (0xd18ac8)
	RunApp: err := app.Run(os.Args)
/home/zachary/Gruntwork/cloud-nuke/main.go:13 (0xd18cb7)
	main: entrypoint.RunApp(app)
/usr/local/go/src/runtime/proc.go:203 (0x436aa2)
	main: fn()
/usr/local/go/src/runtime/asm_amd64.s:1373 (0x464141)
	goexit: BYTE	$0x90	// NOP
  error="parsing time \"2020-10-07 20:13:29.422211915 +0000 UTC\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \" 20:13:29.422211915 +0000 UTC\" as \"T\""
```
Looks like a time parsing / formatting issue.

ina-stoyanova:
Ah interesting - I hit that issue too, just under different circumstances (when I was writing the tests).

I’m curious to see if I can replicate your issue.

The problem for me before was that the RFC3339 format did not understand these +0000 UTC at the end.

The solution I found earlier (also got my tests passing for this issue) was to set the tag as “RFC3339” too & then parse it explicitly in that format too:

time.Parse(time.RFC3339, aws.StringValue(tag.Value))  to retrieve the tag value
time...Format(time.RFC3339) to set the tag in the tests

README.md
```diff
@@ -17,20 +17,23 @@ The currently supported functionality includes:
...
```
brikis98, line 28: This entire Caveats section can probably be removed as ECS cluster deletion works... So there are no caveats for users to be aware of.

aws/ecs_cluster.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+const firstSeenTagKey = "cloud-nuke-first-seen"
```
brikis98, line 14: NIT: don't we already have a variable like this? Could we use the same/consistent var everywhere?

aws/ecs_cluster.go
```diff
@@ -0,0 +1,132 @@
...
+
+		if firstSeenTime.IsZero() {
+			err := tagEcsCluster(awsSession, clusterArn, firstSeenTagKey, time.Now().UTC().String())
+			if err != nil {
+				logging.Logger.Errorf("Error tagigng the ECS cluster with ARN %s", *clusterArn)
+				return nil, err
+			}
+		}
+
+		if excludeAfter.After(firstSeenTime) {
```
brikis98, line 64: Should this be in an `else`? After all, if the cluster was not tagged (the `IsZero()` check above), we'd never want to delete it, would we?

aws/ecs_cluster.go
```diff
@@ -0,0 +1,132 @@
...
+	svc := ecs.New(awsSession)
+
+	numNuking := len(ecsClusterArns)
+
+	if numNuking == 0 {
+		logging.Logger.Infof("No ECS clusters to nuke in region %s", *awsSession.Config.Region)
+		return nil
+	}
+
+	logging.Logger.Infof("Deleting %d ECS clusters in region %s", numNuking, *awsSession.Config.Region)
```
brikis98, line 82: Use `aws.StringValue` to more safely dereference pointers in the AWS SDK.

aws/ecs_cluster.go
```diff
@@ -0,0 +1,132 @@
...
+
+	var nukedEcsClusters []*string
+	for _, clusterArn := range ecsClusterArns {
+		params := &ecs.DeleteClusterInput{
+			Cluster: clusterArn,
+		}
+		_, err := svc.DeleteCluster(params)
+
+		if err != nil {
+			logging.Logger.Errorf("Error, failed to delete cluster with ARN %s", *clusterArn)
```
brikis98, line 92: Do we need to return the error at this point?

aws/ecs_cluster.go
```diff
@@ -0,0 +1,132 @@
...
+		logging.Logger.Errorf("Error getting the tags for ECS cluster with ARN %s", *clusterArn)
+		return firstSeenTime, nil
+	}
+
+	for _, tag := range clusterTags.Tags {
+		if aws.StringValue(tag.Key) == tagKey {
+
+			firstSeenTime, err := time.Parse(time.RFC3339, *tag.Value)
+			if err != nil {
+				logging.Logger.Errorf("Error tagging the ECS cluster with ARN %s", *clusterArn)
```
brikis98, line 124: NIT: wrong error message

aws/ecs_cluster.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, clusterArn := range clusterArns {
+
+		firstSeenTime, err := getClusterTag(awsSession, clusterArn, firstSeenTagKey)
+		if err != nil {
+			logging.Logger.Errorf("Error getting the `first_seen` tag for ECS cluster with ARN %s", *clusterArn)
+			return nil, err
+		}
+
+		if firstSeenTime.IsZero() {
+			err := tagEcsCluster(awsSession, clusterArn, firstSeenTagKey, time.Now().UTC().String())
```
brikis98, line 57: Don't you need to explicitly format the timestamp here to `time.RFC3339` format?

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	awsSession, err := session.NewSession(&awsgo.Config{
+		Region: awsgo.String(region),
+	})
+	require.NoError(t, err)
+
+	cluster := createEcsFargateCluster(t, awsSession, util.UniqueID())
+	defer deleteEcsCluster(awsSession, cluster)
+
+	tagValue := time.Now().UTC().Format(time.RFC3339)
```
brikis98, line 28: I notice you're tagging the cluster with an explicitly formatted timestamp here in the test, but not the module code above... I suspect you'll want to do it the same way in both places. You should prob create `formatTimestampTag` and `parseTimestampTag` methods that both the implementation code and test code use.

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	tagEcsCluster(awsSession, cluster.ClusterArn, firstSeenTagKey, tagValue)
+	require.NoError(t, err)
```
brikis98, line 31:
```suggestion
	tagErr := tagEcsCluster(awsSession, cluster.ClusterArn, firstSeenTagKey, tagValue)
	require.NoError(t, tagErr)
```

brikis98: BTW, your IDE should warn you loudly about unhandled errors in Go, as they are almost always bugs. Make sure to turn that on!

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,104 @@
...
+	cluster2 := createEcsFargateCluster(t, awsSession, util.UniqueID())
+	defer deleteEcsCluster(awsSession, cluster2)
+
+	now := time.Now().UTC()
+	var olderClusterTagValue = now.Add(time.Hour * time.Duration(-48)).Format(time.RFC3339)
+	var youngerClusterTagValue = now.Add(time.Hour * time.Duration(-23)).Format(time.RFC3339)
+
+	tagEcsCluster(awsSession, cluster1.ClusterArn, firstSeenTagKey, olderClusterTagValue)
+	tagEcsCluster(awsSession, cluster2.ClusterArn, firstSeenTagKey, youngerClusterTagValue)
+	require.NoError(t, err)
```
brikis98, line 59:
```suggestion
	err1 := tagEcsCluster(awsSession, cluster1.ClusterArn, firstSeenTagKey, olderClusterTagValue)
	require.NoError(t, err1)
	err2 := tagEcsCluster(awsSession, cluster2.ClusterArn, firstSeenTagKey, youngerClusterTagValue)
	require.NoError(t, err2)
```

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	now := time.Now().UTC()
+	var oldClusterTagValue1 = now.Add(time.Hour * time.Duration(-48)).Format(time.RFC3339)
+	var youngClusterTagValue = now.Format(time.RFC3339)
+	var oldClusterTagValue2 = now.Add(time.Hour * time.Duration(-27)).Format(time.RFC3339)
+
+	tagEcsCluster(awsSession, cluster1.ClusterArn, firstSeenTagKey, oldClusterTagValue1)
+	tagEcsCluster(awsSession, cluster2.ClusterArn, firstSeenTagKey, youngClusterTagValue)
+	tagEcsCluster(awsSession, cluster3.ClusterArn, firstSeenTagKey, oldClusterTagValue2)
+	require.NoError(t, err)
```
brikis98, line 92:
```suggestion
	err1 := tagEcsCluster(awsSession, cluster1.ClusterArn, firstSeenTagKey, oldClusterTagValue1)
	require.NoError(t, err1)	
	err2 := tagEcsCluster(awsSession, cluster2.ClusterArn, firstSeenTagKey, youngClusterTagValue)
	require.NoError(t, err2)	
	err3 := tagEcsCluster(awsSession, cluster3.ClusterArn, firstSeenTagKey, oldClusterTagValue2)
	require.NoError(t, err3)
```

aws/ecs_cluster_types.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// ResourceIdentifiers - the collected ECS clusters
+func (clusters ECSClusters) ResourceIdentifiers() []string {
+	return clusters.ClusterArns
+}
+
+// MaxBatchSize - the maximum number of ECS clusters for a single request
+func (clusters ECSClusters) MaxBatchSize() int {
+	return 200
```
brikis98, line 26: Is this a number you found in the API or SDK docs or something?

brikis98: (Might want to clarify in comment)

brikis98 (commented):
Almost there!

aws/ecs_cluster.go
ina-stoyanova: Thanks! I mistakenly hard-resetted 😱 that yesterday and forgot to re-add. I have now updated all pointers to be `aws.StringValue()`. Could you point me to some guidance why it's better than just using the pointer? I found this so far, which is somewhat explanatory: https://github.com/aws/aws-sdk-go/issues/114

aws/ecs_cluster.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+
+	numNuked := len(nukedEcsClusters)
+	logging.Logger.Infof("[OK] %d of %d ECS cluster(s) deleted in %s", numNuked, numNuking, aws.StringValue(awsSession.Config.Region))
+
+	return nil
+}
+
+func getClusterTag(awsSession *session.Session, clusterArn *string, tagKey string) (time.Time, error) {
```
infraredgirl, line 105: A question about this function. Based on the function name and signature, it looks like a generic "get tag" function that does not make an assumption about what the tag key/value is. However, the implementation is specific to the "first seen" tag - e.g. in the line 122, we are assuming that the `tag.Value` can be parsed into a `time` structure. Is this a concern? Should we refactor, so that there is a generic `getClusterTag` function, and there a more specific one that parses the tag value? (Or at least rename this existing function to suggest that it is opinionated about what kind of tag it is getting?)

aws/ecs_cluster.go
ina-stoyanova: Good point - we do use it just in one other place - the `eip.go` module. I've added a `globals.go` file with the constant declared there

aws/ecs_cluster.go
brikis98: The AWS Go SDK uses pointers all over the place as a way to handle optional values: that is, data that might or might not be in the JSON/XML data that's exchanged during API calls. That means that any of those pointers could be `nil`. If you dereference a `nil` pointer in Go, you get a panic, and your app exits, with no chance for clean up code or anything else to run. You can add things to your code to catch (or more accurately, "recover from") panics, but it's a messy andd error prone business. You can also read the docs and try to deeply understand exactly which values are guaranteed to be non-`nil`, but again, it's a messy and error prone business (and may change as the API evolves).

The `XXXValue` methods (e.g., `StringValue`) in the SDK do a `nil`-safe dereference: e.g., if the value is `nil`, `StringValue` returns an empty string.

So, in general, we:

1. Never use `panic` ourselves in code. Instead, we always return an `error` all the way up the stack, and only the main entry method may choose to `panic` or exit in some other way.
1. Handle every single `error` that is returned. We never ignore them.
1. Use `XXXValue` methods whenever derferencing `nil` pointers in the AWS Go SDK.
1. Check for `nil` and empty arrays whenever possible.

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"time"
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/gruntwork-io/cloud-nuke/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const region = "eu-west-1"
```
infraredgirl, line 14: Nit: The convention is to use all caps to name constants to make it more obvious when reading the code that it's constant. E.g. `REGION` instead of `region`.

aws/ecs_cluster_types.go
infraredgirl: Also would be good to extract magic numbers like this into a const. I.E. at the top of the file, you can have `const MAX_BATCH_SIZE` = 200` (preceded with a comment that explains where that comes from).

infraredgirl (commented):
Looks really good @ina-stoyanova! I left some comments inline.

aws/ecs_cluster.go
infraredgirl: I just want to say that I really appreciate in-depth and well thought-out answers like this. As a new contributor, it helps _so_ much to ramp up more quickly and easily. Thank you, @brikis98!

aws/ecs_cluster_test.go
ina-stoyanova: Ah yes - I'm looking to see what extension would catch such sort of an overlook, or what is the VSCode setting i need switched on. I know we chatted briefly about this, so I'll ask a question on Slack about what people use.

aws/ecs_cluster.go
brikis98: Glad to hear it!

Actually, something @yorinasub17 started recently that I think is great is creating style guides. We have one here for Terraform: https://www.notion.so/gruntwork/Style-guides-0756577007fc41d9916ad34e9940d202. Perhaps we should start building a page for Go? For the most part, we follow Go's official conventions (incl what's enforced by `go fmt`), but some of the items above our uniquely our own. I'm a bit buried today, but if someone wants to start that page on our wiki, copy some of the above, and what's in this comment, I'd be happy to contribute (and I'm sure many others at the company would too).

aws/ecs_cluster.go
brikis98: Ah, big +1 on that. Renaming the function prob makes the most sense for now!

aws/ecs_cluster_test.go
brikis98: In general, that's spot on, but in Go specifically, it's a bit different. Capitalized variables/functions in Go are public, and that may not be what you want. So generally, in Go, you use [mixed caps (AKA CamelCase)](https://golang.org/doc/effective_go.html#mixed-caps), even for constants. So `region` or `testRegion` for private constants and `Region` or `TestRegion` for public ones.

Note that we got this convention wrong in the past (from all prev languages, I was using all caps just as you suggested Ana!), but we've been incrementally switching to the official Go convention.

aws/ecs_cluster.go
infraredgirl: A style guide for Go is a fantastic idea. I'd love to start this. Will post a comment back here once I have something up.

aws/ecs_cluster_types.go
ina-stoyanova: @brikis98 Very good question - it is something I've been meaning to follow up on. I looked up [here](https://docs.aws.amazon.com/cli/latest/reference/ecs/delete-cluster.html) & [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-quotas.html) & it doesn't seem like there's any mention of the need to rate-limit how many ECS Clusters we delete at one time. 

The rate-limiting here has been added to follow pretty much the same pattern as all other modules in the `aws` package - but I'm more than happy to take guidance on this in terms of reducing or increasing the number. 

@infraredgirl Now that we have a `globals.go` file we could extract this there too! I'd suggest we do that separately, though, as that would mean changing all `*_types.go` files in this PR and in one go, which is probably not ideal. Thoughts?

aws/ecs_cluster_test.go
infraredgirl: Ah, good point about the capitalization making things public. I stand correct here - thanks for pointing this out!

ina-stoyanova: awesome - thanks for pointing this out!

aws/ecs_cluster.go
ina-stoyanova: This is very helpful @brikis98 Thank you! Will keep this in mind in future :)

aws/ecs_cluster_types.go
brikis98: Well, if there's no obvious limit, then I'd prob just take a guess at something reasonable, and add a comment explaining just that. Many AWS APIs are limited to 100 "things," so that might be a good starting point.

infraredgirl: > @infraredgirl Now that we have a `globals.go` file we could extract this there too! I'd suggest we do that separately, though, as that would mean changing all `*_types.go` files in this PR and in one go, which is probably not ideal. Thoughts?

I don't see a `globals.go` file - did you perhaps forget to `git add` it? At any rate, I think it's totally fine to change it in a follow up PR - just put a `TODO` comment so we don't forget about it.

aws/ecs_cluster.go
ina-stoyanova: yeah - totally missed this even with the tests.

aws/ecs_cluster_test.go
ina-stoyanova: that's also a good point, thank you 👍

aws/ecs_cluster.go
ina-stoyanova: Absolutely - I can see it going wrong in future, so will rename it 👍 Thanks @infraredgirl :)

aws/ecs_cluster_types.go
ina-stoyanova: Try pulling the latest - you should see [the file in this commit ](https://github.com/ina-stoyanova/cloud-nuke/blob/1af68c1ef81ebc24def9761405115b2253fadd4a/aws/globals.go#L3)

I will extract this out in the file for now, and will follow up on understanding if the MaxBatchSize for all other AWS modules is arbitrary, and to extract it to also in the `globals.go` file

infraredgirl: Yep, it's there now! Cool - sounds good. 👍

aws/ecs_cluster.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+
+		if firstSeenTime.IsZero() {
+			err := tagEcsClusterWhenFirstSeen(awsSession, clusterArn, time.Now().UTC())
+			if err != nil {
+				logging.Logger.Errorf("Error tagigng the ECS cluster with ARN %s", aws.StringValue(clusterArn))
+				return nil, errors.WithStackTrace(err)
+			}
+		} else {
+			if excludeAfter.After(firstSeenTime) {
```
brikis98, line 41: NIT: this could be an `else if`

aws/ecs_cluster_test.go
```diff
@@ -0,0 +1,114 @@
...
+	require.NoError(t, err2)
+	err3 := tagEcsClusterWhenFirstSeen(awsSession, cluster3.ClusterArn, oldClusterTagValue2)
+	require.NoError(t, err3)
+
+	last24Hours := now.Add(time.Hour * time.Duration(-24))
+	filteredClusterArns, err := getAllEcsClustersOlderThan(awsSession, region, last24Hours)
+	require.NoError(t, err)
+
+	nukeEcsClusters(awsSession, filteredClusterArns)
+	require.NoError(t, err)
```
brikis98, line 109:
```suggestion
	nukeErr := nukeEcsClusters(awsSession, filteredClusterArns)
	require.NoError(t, nukeErr)
```

Check your IDE 😁

aws/ecs_cluster_types.go
```diff
@@ -0,0 +1,41 @@
...
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// The pattern we use for running the `cloud-nuke` tool is to split the AWS API calls
+// into batches when the function `NukeAllResources` is executed.
+// A batch max number has been chosen for most modules.
+// However, for ECS clusters there is no explicit limiting described in the AWS CLI docs.
+// Therefore this `maxBatchSize` here is set to 100 as a safe maximum.
```
brikis98, line 13: This seems reasonable to me 👍

aws/globals.go
```diff
@@ -0,0 +1,3 @@
+package aws
+
+const firstSeenTagKey = "cloud-nuke-first-seen"
```
brikis98, line 3: NIT: please add a comment explaining what this is and how it is used.

go.sum
```diff
@@ -23,6 +23,7 @@ github.com/google/uuid v0.0.0-20161128191214-064e2069ce9c h1:jWtZjFEUE/Bz0IeIhqC
 github.com/google/uuid v0.0.0-20161128191214-064e2069ce9c/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/gruntwork-io/gruntwork-cli v0.1.2 h1:lUaJdPTRJ0HGELW8a3r2QJnHA+Dak+PkvOP+T8ykXF4=
 github.com/gruntwork-io/gruntwork-cli v0.1.2/go.mod h1:IBX21bESC1/LGoV7jhXKUnTQTZgQ6dYRsoj/VqxUSZQ=
+github.com/gruntwork-io/gruntwork-cli v0.7.0 h1:YgSAmfCj9c61H+zuvHwKfYUwlMhu5arnQQLM4RH+CYs=
```
brikis98, line 26: Are these version number bumps intentional?

brikis98 (commented):
OK, this is super close. Just a few minor tweaks left.

aws/ecs_cluster_test.go
ina-stoyanova: hahaha, no way! I installed the extensions I found! can't believe it keeps happening!

go.sum
ina-stoyanova: They happened automatically - so I've left them in.

aws/ecs_cluster.go
ina-stoyanova: Both versions read fine to me, but very happy to change it 👍

brikis98: No strong preference from me, so your call!

aws/ecs_cluster_test.go
brikis98: I may have missed others, so yea, best to let the computer do the hard work 😁

brikis98 (approved):
Code LGTM! Make sure the tests are running and pass and this is good to merge! 🎉

infraredgirl (approved):
Nice, LGTM!

PR #128, opened 2020-07-21
Fix deploy step

yorinasub17:


.circleci/config.yml
```diff
@@ -55,7 +55,8 @@ jobs:
   deploy:
     <<: *defaults
     steps:
-      - checkout
+      - attach_workspace:
+          at: .
```
brikis98, line 59: I was trying to figure out how this ever worked before... For anyone else who is curious, it turns out that this step used to do `attach_workspace`, but it accidentally got dropped: https://github.com/gruntwork-io/cloud-nuke/commit/cc11314d70f2632d2c86e405c1b02dcfaf9c6fb4#diff-1d37e48f9ceff6d8030570cd36286a61L30-L31

yorinasub17:
Thanks for review! I verified this worked on the `v0.0.0-alpha.1` test release, so will go ahead and merge this so that future releases will work correctly.

Issue #117, opened 2020-05-24
Homebrew support


chenrui333:
Just FYI, I have added one formula for this tool in the homebrew side (https://github.com/Homebrew/homebrew-core/pull/55228)

brikis98:
Nice, thanks! Looks like that PR is closed though?

chenrui333:
@brikis98 yeah, I need a new release to kick if off again (not sure if you can make one anytime soon :) )

chenrui333:
Currently, it failed with audit failure.

```
==> brew audit --skip-style cloud-nuke --online --new-formula
==> FAILED
Error: undefined local variable or method `created_pr_comment' for Homebrew:Module
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:148:in `audit'
/usr/local/Homebrew/Library/Homebrew/brew.rb:110:in `<main>'
cloud-nuke:
  * Formulae should not require patches to build. Patches should be submitted and accepted upstream first.
```

chenrui333:
@brikis98 if possible, can I get a new release so that I can proceed with the homebrew formula? This would be a great tool for lots of homebrew users.

brikis98:
Done! Binaries should show up shortly. https://github.com/gruntwork-io/cloud-nuke/releases/tag/v0.1.19

chenrui333:
thanks, I will resume what I did.

chenrui333:
Made a new one https://github.com/Homebrew/homebrew-core/pull/55559

chenrui333:
Have merged homebrew-core PR for a while.

Here is my installation log for the brew formula.
```
==> Downloading https://homebrew.bintray.com/bottles/cloud-nuke-0.1.20.mojave.bottle.tar.gz
Already downloaded: /Users/rchen/Library/Caches/Homebrew/downloads/52c0b8a5380156c4566a6117ac0677d4bfbcb603c98d4d2fbbeb91c976ef1d6d--cloud-nuke-0.1.20.mojave.bottle.tar.gz
==> Pouring cloud-nuke-0.1.20.mojave.bottle.tar.gz
==> Caveats
Before you can use these tools, you must export some variables to your $SHELL.
  export AWS_ACCESS_KEY="<Your AWS Access ID>"
  export AWS_SECRET_KEY="<Your AWS Secret Key>"
  export AWS_REGION="<Your AWS Region>"
==> Summary
[...]

PR #129, opened 2020-07-21
Add homebrew to install instructions on README

marinalimeira:
This option is available per discussion on https://github.com/gruntwork-io/cloud-nuke/issues/117

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ When executed as `cloud-nuke defaults-aws`, this tool deletes all DEFAULT VPCs a
 3. Add execute permissions to the binary. E.g.: `chmod u+x /usr/local/bin/cloud-nuke`.
 4. Test it installed correctly: `cloud-nuke --help`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,8 @@ When executed as `cloud-nuke defaults-aws`, this tool deletes all DEFAULT VPCs a
 3. Add execute permissions to the binary. E.g.: `chmod u+x /usr/local/bin/cloud-nuke`.
 4. Test it installed correctly: `cloud-nuke --help`.
 
+As an alternative, you can install it using [Homebrew](https://brew.sh/): `brew install cloud-nuke`
```
yorinasub17, line 48: NIT: I think we should have our usual caveat about this. See https://terragrunt.gruntwork.io/docs/getting-started/install/#install-via-a-package-manager

marinalimeira: Not sure if I follow, it's to update the text or add more things?

yorinasub17: It's to add the paragraph about the homebrew packages being maintained by a 3rd party and not by us, so versions may drift.

PR #135, opened 2020-09-21
Memory efficient s3 delete

yorinasub17:
This fixes a few bugs in the s3 deletion routine:

- `cloud-nuke` can crash with OOM if there are too many objects in the bucket. This is because it loads all the objects into memory before starting to delete them. **This was fixed by combining the deletion routine with the pager.**

- No error was returned if there was an error emptying the bucket. This led to confusing error messages where it may have failed to empty the bucket, but cloud-nuke exits because it can't delete the bucket itself. **This was fixed by updating the routine to return a combined error for all the buckets**

- Tests were not checking errors in the nuke routine, so it would exit with a pass even when the nuking failed.

### DIFF
--- a/aws/s3.go
+++ b/aws/s3.go
@@ -495,9 +495,22 @@ func nukeEmptyS3Bucket(svc *s3.S3, bucketName *string, verifyBucketDeletion bool
 		return err
 	}
 
-	err = svc.WaitUntilBucketNotExists(&s3.HeadBucketInput{
-		Bucket: bucketName,
-	})
+	// The wait routine will try for up to 100 seconds, but that is not long enough for all circumstances of S3. As
+	// such, we retry this routine up to 3 times for a total of 300 seconds.
+	const maxRetries = 3
+	for i := 0; i < maxRetries; i++ {
+		logging.Logger.Infof("Waiting until bucket (%s) deletion is propagated (attempt %d / %d)", aws.StringValue(bucketName), i+1, maxRetries)
+		err = svc.WaitUntilBucketNotExists(&s3.HeadBucketInput{
+			Bucket: bucketName,
+		})
+		// Exit early if no error
+		if err == nil {
+			logging.Logger.Info("Successfully detected bucket deletion.")
+			return nil
+		}
+		logging.Logger.Warnf("Error waiting for bucket (%s) deletion propagation (attempt %d / %d)", aws.StringValue(bucketName), i+1, maxRetries)
+		logging.Logger.Warnf("Underlying error was: %s", err)
+	}
 	return err
 }
 
--- a/aws/s3.go
+++ b/aws/s3.go
@@ -314,7 +314,6 @@ func emptyBucket(svc *s3.S3, bucketName *string, isVersioned bool, batchSize int
 	// Since the error may happen in the inner function handler for the pager, we need a function scoped variable that
 	// the inner function can set when there is an error.
 	var errOut error
-	errOut = nil
 	pageId := 1
 
 	// Handle versioned buckets.
--- a/aws/s3.go
+++ b/aws/s3.go
@@ -309,7 +309,7 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
 // NOTE: In the progress logs, we deliberately do not report how many pages or objects are left. This is because aws
 // does not provide any API for getting the object count, and the only way to do that is to iterate through all the
 // objects. For memory and time efficiency, we opted to delete the objects as we retrieve each page, which means we
-// don't know how many is left until we complete all the operations.
+// don't know how many are left until we complete all the operations.
 func emptyBucket(svc *s3.S3, bucketName *string, isVersioned bool, batchSize int) error {
 	// Since the error may happen in the inner function handler for the pager, we need a function scoped variable that
 	// the inner function can set when there is an error.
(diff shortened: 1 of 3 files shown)


aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+				if err := deleteDeletionMarkers(svc, bucketName, page.DeleteMarkers); err != nil {
+					logging.Logger.Errorf("Error deleting deletion markers for page %d from bucket %s: %s", pageId, aws.StringValue(bucketName), err)
+					errOut = err
+					return false
 				}
+				logging.Logger.Infof("[OK] - deleted object page %d (%d deletion markers) from bucket %s", pageId, len(page.DeleteMarkers), aws.StringValue(bucketName))
+
+				pageId++
 				return true
 			},
```
yorinasub17, line 346: The key change is here, where instead of getting all the bucket objects and then calling delete on the combined list, we call the delete operation in the page handler so that we delete each page as it is queried.

I was worried if this would cause a race condition since we are modifying the list that is being queried, but empirical testing shows that AWS handles this correctly (see below for annotation in test highlighting the empirical test).

aws/s3_test.go
```diff
@@ -410,16 +418,18 @@ func TestNukeS3Bucket(t *testing.T) {
...
 				},
 			},
 			{
 				bucketType + "_BatchObjects_ValidBatchSize",
 				TestNukeS3BucketArgs{
 					isVersioned:       isVersioned,
 					checkDeleteMarker: false,
-					objectCount:       10,
+					objectCount:       30,
 					objectBatchsize:   5,
```
yorinasub17, line 430: Bumped the object count here to make sure the delete in page routine works even when there are many pages (in this case at least 6).

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
 
-// getS3BucketObjects returns S3 bucket objects struct for a given bucket name
-func getS3BucketObjects(svc *s3.S3, bucketName *string, isVersioned bool) ([]*s3.ObjectIdentifier, error) {
-	identifiers := []*s3.ObjectIdentifier{}
+// emptyBucket will empty the given S3 bucket by deleting all the objects that are in the bucket. For versioned buckets,
+// this includes all the versions and deletion markers in the bucket.
+// NOTE: In the progress logs, we deliberately do not report how many pages or objects are left. This is because aws
+// does not provide any API for getting the object count, and the only way to do that is to iterate through all the
+// objects. For memory and time efficiency, we opted to delete the objects as we retrieve each page, which means we
+// don't know how many is left until we complete all the operations.
```
brikis98, line 312:
```suggestion
// don't know how many are left until we complete all the operations.
```

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+func emptyBucket(svc *s3.S3, bucketName *string, isVersioned bool, batchSize int) error {
+	// Since the error may happen in the inner function handler for the pager, we need a function scoped variable that
+	// the inner function can set when there is an error.
+	var errOut error
+	errOut = nil
```
brikis98, line 317: Seems unnecessary to set to `nil`?

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+				MaxKeys: aws.Int64(int64(batchSize)),
 			},
 			func(page *s3.ListObjectVersionsOutput, lastPage bool) (shouldContinue bool) {
-				for _, obj := range page.Versions {
-					logging.Logger.Debugf("Bucket %s object %s version %s", *bucketName, *obj.Key, *obj.VersionId)
-					identifiers = append(identifiers, &s3.ObjectIdentifier{
-						Key:       obj.Key,
-						VersionId: obj.VersionId,
-					})
+				logging.Logger.Debugf("Deleting object page %d (%d objects) from bucket %s", pageId, len(page.Versions), aws.StringValue(bucketName))
```
brikis98, line 328:
```suggestion
				logging.Logger.Debugf("Deleting page %d of object versions (%d objects) from bucket %s", pageId, len(page.Versions), aws.StringValue(bucketName))
```

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+				if err := deleteObjectVersions(svc, bucketName, page.Versions); err != nil {
+					logging.Logger.Errorf("Error deleting objects for page %d from bucket %s: %s", pageId, aws.StringValue(bucketName), err)
```
brikis98, line 330:
```suggestion
					logging.Logger.Errorf("Error deleting objects versions for page %d from bucket %s: %s", pageId, aws.StringValue(bucketName), err)
```

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+					errOut = err
+					return false
 				}
-				for _, obj := range page.DeleteMarkers {
-					logging.Logger.Debugf("Bucket %s object %s DeleteMarker %s", *bucketName, *obj.Key, *obj.VersionId)
-					identifiers = append(identifiers, &s3.ObjectIdentifier{
-						Key:       obj.Key,
-						VersionId: obj.VersionId,
-					})
+				logging.Logger.Infof("[OK] - deleted object page %d (%d objects) from bucket %s", pageId, len(page.Versions), aws.StringValue(bucketName))
```
brikis98, line 334:
```suggestion
				logging.Logger.Infof("[OK] - deleted page %d of object versions (%d objects) from bucket %s", pageId, len(page.Versions), aws.StringValue(bucketName))
```

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+
+				logging.Logger.Debugf("Deleting object page %d (%d deletion markers) from bucket %s", pageId, len(page.DeleteMarkers), aws.StringValue(bucketName))
```
brikis98, line 336:
```suggestion
				logging.Logger.Debugf("Deleting page %d of deletion markers (%d deletion markers) from bucket %s", pageId, len(page.DeleteMarkers), aws.StringValue(bucketName))
```

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+				if err := deleteDeletionMarkers(svc, bucketName, page.DeleteMarkers); err != nil {
+					logging.Logger.Errorf("Error deleting deletion markers for page %d from bucket %s: %s", pageId, aws.StringValue(bucketName), err)
+					errOut = err
+					return false
 				}
+				logging.Logger.Infof("[OK] - deleted object page %d (%d deletion markers) from bucket %s", pageId, len(page.DeleteMarkers), aws.StringValue(bucketName))
```
brikis98, line 342:
```suggestion
				logging.Logger.Infof("[OK] - deleted page %d of deletion markers (%d deletion markers) from bucket %s", pageId, len(page.DeleteMarkers), aws.StringValue(bucketName))
```

aws/s3.go
```diff
@@ -304,119 +304,181 @@ func excludeBucketByREList(bucketName string, reList []*regexp.Regexp) bool {
...
+func nukeAllS3BucketObjects(svc *s3.S3, bucketName *string, batchSize int) error {
+	versioningResult, err := svc.GetBucketVersioning(&s3.GetBucketVersioningInput{
+		Bucket: bucketName,
+	})
+	if err != nil {
+		return err
+	}
 
-		logging.Logger.Infof("[OK] - %d-%d objects of batch %d/%d - Bucket: %s - deleted", i+1, j, batchCount, totalBatches, *bucketName)
+	isVersioned := versioningResult.Status != nil && *versioningResult.Status == "Enabled"
```
brikis98, line 471: NIT: I know the code was this way before, but a minor improvement:

```suggestion
	isVersioned := aws.StringValue(versioningResult.Status) == "Enabled"
```

aws/s3_test.go
```diff
@@ -185,7 +186,14 @@ func testListS3Bucket(t *testing.T, args TestListS3BucketArgs) {
 	// Please note that we are passing the same session that was used to create the bucket
 	// This is required so that the defer cleanup call always gets the right bucket region
 	// to delete
-	defer nukeAllS3Buckets(awsParams.awsSession, []*string{aws.String(bucketName)}, 1000)
+	defer func() {
+		_, err := nukeAllS3Buckets(awsParams.awsSession, []*string{aws.String(bucketName)}, 1000)
+		if args.shouldError {
+			assert.Error(t, err)
+		} else {
+			assert.NoError(t, err)
```
brikis98, line 194: Does `assert` work in a separate goroutine? Or can you only call these in the original goroutine used by the test?

yorinasub17:
Feedback from slack:

> Did you change the error handling so that the app exits if it fails to delete a resource? IMHO it would be better to report the failure and continue rather than exit (old behavior)

aws/s3_test.go
yorinasub17: I'm a bit confused: which goroutine are you referring?

aws/s3.go
yorinasub17: Removed in 2712702

yorinasub17:
> Did you change the error handling so that the app exits if it fails to delete a resource? IMHO it would be better to report the failure and continue rather than exit (old behavior)

yorinasub17:
Ok I believe I addressed all the comments! This is ready for a rereview!

aws/s3_test.go
brikis98: Oh, scratch that, I think I just had a brain fart and misread `defer func() { ... }` as `go func() { ... }`.

brikis98 (approved):
LGTM!

yorinasub17:

PR #136, opened 2020-09-30
Ignore errors

yorinasub17:
By default `cloud-nuke` will halt processing if it encounters any error with nuking a region or resource, instead of
continuing to process additional resources. This PR introduces a new CLI flag (`--ignore-errrors`) to inform `cloud-nuke` to ignore errors and continue processing, but aggregate the errors at the end so the overall command still exists with a non-zero exit code.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -17,6 +17,16 @@ import (
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
+// Options for configuring the nuking operation.
+type NukeOptions struct {
+	// What regions to nuke resources in.
+	Regions []string
+
+	// Whether or not to ignore errors during processing. When true, cloud-nuke will continue to process additional
+	// resources even if any of them fail with an error.
+	IgnoreErrors bool
+}
+
 // OptInNotRequiredRegions contains all regions that are enabled by default on new AWS accounts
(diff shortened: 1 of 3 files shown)


aws/aws.go
```diff
@@ -511,6 +527,11 @@ func NukeAllResources(account *AwsAccountResources, regions []string) error {
 						continue
 					}
 
+					if options.IgnoreErrors {
+						logging.Logger.Warnf("Encountered error nuking resource type %s in region %s, but --ignore-errors flag is passed in so continuing.", resources.ResourceName(), region)
+						allErrors.Add(RegionResourceInfo{region: region, resource: resources.ResourceName()}, err)
```
brikis98, line 532: Handling the error at this level means that this particular batch will only be partially processed, right? E.g., Perhaps it successfully deletes ELBs, but hits an error half way through deleting ASGs, so for this batch, it does not finish the other half of the ASGs, or any other resources after the ASG, such as EC2s, S3 buckets, etc?

brikis98 (commented):
How does this compare with https://github.com/gruntwork-io/cloud-nuke/pull/114?

yorinasub17:
Ahhhh I somehow missed that PR. Will close this and if necessary, extend that PR.

PR #144, opened 2020-10-14
Nuke gruntwork S3 buckets

yorinasub17:
This updates the CircleCI config to run nuke jobs for S3 buckets. To avoid deleting buckets that we expect to use, this adds a config file with the regex list that we have been using in the `aws-gc.py` script.

Note that I also updated the nuke jobs to not depend on the tests passing. It seems unnecessary to continuously run tests for every nuke job, especially since they are fairly slow. Not to mention that they are flaky and transiently fail, causing us to skip a nuke cycle.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,9 +92,7 @@ workflows:
             branches:
               only: master
     jobs:
-      - nuke_phx_devops:
-          requires:
-            - test
+      - nuke_phx_devops
   nightly:
     triggers:
       - schedule:
@@ -103,6 +101,4 @@ workflows:
             branches:
               only: master
     jobs:
-      - nuke_sandbox:
-          requires:
-            - test
+      - nuke_sandbox
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
             go run main.go aws \
               --older-than 1h \
               --force \
-              --exclude-resource-type s3 \
+              --config ./.circleci/nuke_config.yml \
               --exclude-resource-type iam
           no_output_timeout: 1h
   nuke_sandbox:
@@ -46,7 +46,7 @@ jobs:
             go run main.go aws \
               --older-than 24h \
               --force \
-              --exclude-resource-type s3 \
+              --config ./.circleci/nuke_config.yml \
               --exclude-resource-type iam
           no_output_timeout: 1h
   deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,6 @@ workflows:
             branches:
               only: master
     jobs:
-      - test
       - nuke_phx_devops:
           requires:
             - test
@@ -104,7 +103,6 @@ workflows:
             branches:
               only: master
     jobs:
-      - test
       - nuke_sandbox:
           requires:
             - test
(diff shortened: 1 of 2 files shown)


infraredgirl (dismissed):
LGTM!

yorinasub17:
Thanks for review! Will wait for Jim to weigh in on

> Note that I also updated the nuke jobs to not depend on the tests passing.

to make sure he is ok with that change prior to merge.

.circleci/cloud_nuke_config.yml
```diff
@@ -0,0 +1,19 @@
+s3:
```
brikis98, line 1: NIT: we should probably make it clearer what this file is. At the least, add a comment explaining that this is a config file specifically for the Gruntwork Phx DevOps accounts, where we want to delete certain S3 buckets, but not all. Might be worth renaming the file to make that clearer too.

brikis98: Also, I'm having a brain fart... But with the config file, it still deletes all resource types, and the config here just tells it how to filter `s3`, with the assumption all other resources are unfiltered? That is, by only specifying `s3` here, it doesn't mean _only_ delete S3, right?

brikis98 (commented):
Removing tests before regular nuking makes sense to me!

.circleci/cloud_nuke_config.yml
yorinasub17: > we should probably make it clearer what this file is.

done: 447d363

> But with the config file, it still deletes all resource types, and the config here just tells it how to filter s3, with the assumption all other resources are unfiltered? That is, by only specifying s3 here, it doesn't mean only delete S3, right?

Yes, this config is just for specifying what s3 buckets to delete, not "just delete s3". Here is a part of the run log on my local machine that confirms this:

``` go run main.go aws \
              --older-than 1h \
              --config ./.circleci/cloud_nuke_config.yml

INFO[2020-10-15T09:20:30-05:00] The following resource types will be nuked:
INFO[2020-10-15T09:20:30-05:00] - ami
INFO[2020-10-15T09:20:30-05:00] - asg
INFO[2020-10-15T09:20:30-05:00] - ebs
INFO[2020-10-15T09:20:30-05:00] - ec2
INFO[2020-10-15T09:20:30-05:00] - ecscluster
INFO[2020-10-15T09:20:30-05:00] - ecsserv
INFO[2020-10-15T09:20:30-05:00] - eip
INFO[2020-10-15T09:20:30-05:00] - ekscluster
INFO[2020-10-15T09:20:30-05:00] - elb
INFO[2020-10-15T09:20:30-05:00] - elbv2
INFO[2020-10-15T09:20:30-05:00] - lc
INFO[2020-10-15T09:20:30-05:00] - rds
INFO[2020-10-15T09:20:30-05:00] - s3
INFO[2020-10-15T09:20:30-05:00] - snap
INFO[2020-10-15T09:20:32-05:00] Retrieving active AWS resources in [eu-north-1, ap-south-1, eu-west-3, eu-west-2, eu-west-1, ap-northeast-2, ap-northeast-1, sa-east-1, ca-central-1, ap-southeast-1, ap-southeast-2, eu-central-1, us-east-1, us-east-2,
```

brikis98: Roger, thx

brikis98 (dismissed):
LGTM!

hposca (approved):
LGTM :+1:

yorinasub17:
Thanks for review! Going to merge this in now, given all the past approvals as well.

Issue #145, opened 2020-10-15
TestCanNukeAllEcsClustersOlderThan24Hours failing due to 'INACTIVE' ECS clusters


ina-stoyanova:
**The issue:**
This test fails in some scenarios when there's inactive ECS Clusters - can be seen in these CircleCI test runs:
https://app.circleci.com/pipelines/github/gruntwork-io/cloud-nuke/1679/workflows/b30a43e4-770c-4759-b7b0-8e900d1316ee/jobs/21376
```
--- FAIL: TestCanNukeAllEcsClustersOlderThan24Hours (2.01s)
    ecs_cluster_test.go:106: 
        	Error Trace:	ecs_cluster_test.go:106
        	Error:      	Received unexpected error:
        	            	InvalidParameterException: The specified cluster is inactive. Specify an active cluster and try again.
        	            		status code: 400, request id: 606792e6-68de-44fd-b571-d7aa14381d76
        	Test:       	TestCanNukeAllEcsClustersOlderThan24Hours
```
**Suggestion:**
Could filter out ECS clusters, so that tagging can be only applied to 'ACTIVE' clusters

ina-stoyanova:
Closing as solved with PR above.

PR #146, opened 2020-10-15
Fix for inactive ecs clusters

ina-stoyanova:
Hey, this PR is aiming to add filtering to ECS Clusters. The issue I was seeing is described here: https://github.com/gruntwork-io/cloud-nuke/issues/145

**Overview:** 
In some scenarios, the ECS Clusters TagResource function would be called while the cluster is not in `ACTIVE` state. This meant that the tests we have fail, and also the code itself was flawed. 

**It has now been updated:**
- `cloud-nuke ecscluster` will nuke ECS Clusters which are also set as `ACTIVE` or `PROVISIONING`
- `PROVISIONING` is also part of the condition, as the assumption is the cluster will be discoverable, and its state is going towards being `ACTIVE`
- all other states (`INACTIVE`, `DEPROVISIONING`, `FAILED`) are not relevant and therefore used in this functionality (you can't tag a cluster that's not active)

Additional changes:
- the tests now use a random region 
- the tests also now use a name for the clusters `cloud-nuke-test-XXXXX`

### DIFF
--- a/aws/ecs_cluster.go
+++ b/aws/ecs_cluster.go
@@ -67,7 +67,7 @@ func getAllEcsClustersOlderThan(awsSession *session.Session, region string, excl
 
 	clusterArns, err := getAllActiveEcsClusterArns(awsSession)
 	if err != nil {
-		logging.Logger.Errorf("Error getting all ECS clusters with `ACTIVE` status")
+		logging.Logger.Errorf("Error getting all ECS clusters with `ACTIVE` or `PROVISIONING` status")
 		return nil, errors.WithStackTrace(err)
 	}
 
--- a/aws/ecs_cluster.go
+++ b/aws/ecs_cluster.go
@@ -11,14 +11,63 @@ import (
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
+// Used in this context to determine if the ECS Cluster is ready to be used & tagged
+// For more details on other valid status values: https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#Cluster
+const activeEcsClusterStatus string = "ACTIVE"
+const provisioningEcsClusterStatus = "PROVISIONING"
+
+// filter out all active ecs clusters
+func getAllActiveEcsClusterArns(awsSession *session.Session) ([]*string, error) {
+	svc := ecs.New(awsSession)
+
+	allClusters, err := getAllEcsClusters(awsSession)
+	if err != nil {
+		logging.Logger.Errorf("Error getting all ECS clusters")
+		return nil, errors.WithStackTrace(err)
+	}
+
+	var result []*ecs.Cluster
+
+	for _, cluster := range allClusters {
+		input := &ecs.DescribeClustersInput{
+			Clusters: []*string{cluster},
+		}
+
+		describedClusters, describeErr := svc.DescribeClusters(input)
+		if describeErr != nil {
+			logging.Logger.Errorf("Error describing all ECS clusters")
+			return nil, errors.WithStackTrace(describeErr)
+		}
+
+		result = append(result, describedClusters.Clusters...)
+	}
+
+	var filteredEcsClusterArns []*string
+
+	// Filter out invalid state ECS Clusters (will return only `ACTIVE` state clusters)
+	// `cloud-nuke` needs to tag ECS Clusters it sees for the first time.
+	// Therefore to tag a cluster, that cluster must be in the `ACTIVE` state.
+	for _, cluster := range result {
+		logging.Logger.Infof("Status for ECS CLuster %s is %s", aws.StringValue(cluster.ClusterArn), aws.StringValue(cluster.Status))
+
+		if aws.StringValue(cluster.Status) == activeEcsClusterStatus ||
+			aws.StringValue(cluster.Status) == provisioningEcsClusterStatus {
+
+			filteredEcsClusterArns = append(filteredEcsClusterArns, cluster.ClusterArn)
+		}
+	}
+
+	return filteredEcsClusterArns, nil
+}
+
 func getAllEcsClustersOlderThan(awsSession *session.Session, region string, excludeAfter time.Time) ([]*string, error) {
 	awsSession, err := session.NewSession(&awsgo.Config{
 		Region: awsgo.String(region),
 	})
 
-	clusterArns, err := getAllEcsClusters(awsSession)
+	clusterArns, err := getAllActiveEcsClusterArns(awsSession)
 	if err != nil {
-		logging.Logger.Errorf("Error getting all ECS clusters")
+		logging.Logger.Errorf("Error getting all ECS clusters with `ACTIVE` status")
 		return nil, errors.WithStackTrace(err)
 	}
 
--- a/aws/ecs_cluster.go
+++ b/aws/ecs_cluster.go
@@ -16,7 +16,7 @@ import (
 const activeEcsClusterStatus string = "ACTIVE"
 const provisioningEcsClusterStatus = "PROVISIONING"
 
-// filter out all active ecs clusters
+// Filter all active ecs clusters
 func getAllActiveEcsClusterArns(awsSession *session.Session) ([]*string, error) {
 	svc := ecs.New(awsSession)
 
--- a/aws/ecs_cluster.go
+++ b/aws/ecs_cluster.go
@@ -25,20 +25,10 @@ func getAllActiveEcsClusterArns(awsSession *session.Session) ([]*string, error) 
 		return nil, errors.WithStackTrace(err)
 	}
 
-	var result []*ecs.Cluster
-
-	for _, cluster := range allClusters {
-		input := &ecs.DescribeClustersInput{
-			Clusters: []*string{cluster},
-		}
-
-		describedClusters, describeErr := svc.DescribeClusters(input)
-		if describeErr != nil {
-			logging.Logger.Errorf("Error describing ECS clusters")
-			return nil, errors.WithStackTrace(describeErr)
-		}
-
-		result = append(result, describedClusters.Clusters...)
+	described, describeErr := describeEcsClusters(svc, allClusters)
+	if describeErr != nil {
+		logging.Logger.Errorf("Error describing ECS clusters")
+		return nil, errors.WithStackTrace(describeErr)
 	}
 
 	var filteredEcsClusterArns []*string
@@ -46,7 +36,7 @@ func getAllActiveEcsClusterArns(awsSession *session.Session) ([]*string, error) 
 	// Filter and return only `ACTIVE` state ECS Clusters.
 	// `cloud-nuke` needs to tag ECS Clusters it sees for the first time.
 	// Therefore to tag a cluster, that cluster must be in the `ACTIVE` state.
-	for _, cluster := range result {
+	for _, cluster := range described {
 		logging.Logger.Debugf("Status for ECS Cluster %s is %s", aws.StringValue(cluster.ClusterArn), aws.StringValue(cluster.Status))
 
 		if aws.StringValue(cluster.Status) == activeEcsClusterStatus {
@@ -56,6 +46,29 @@ func getAllActiveEcsClusterArns(awsSession *session.Session) ([]*string, error) 
 
 	return filteredEcsClusterArns, nil
 }
+
+func describeEcsClusters(svc *ecs.ECS, allClusters []*string) ([]*ecs.Cluster, error) {
+	var result []*ecs.Cluster
+	batches := split(aws.StringValueSlice(allClusters), 10)
+
+	for _, batch := range batches {
+		input := &ecs.DescribeClustersInput{
+			Clusters: awsgo.StringSlice(batch),
+		}
+
+		describedClusters, describeErr := svc.DescribeClusters(input)
+		if describeErr != nil {
+			logging.Logger.Errorf("Error describing ECS clusters from input %s: ", input)
+			return nil, errors.WithStackTrace(describeErr)
+		}
+
+		result = append(result, describedClusters.Clusters...)
+	}
+
+	return result, nil
+}
+
+
 
 func getAllEcsClustersOlderThan(awsSession *session.Session, region string, excludeAfter time.Time) ([]*string, error) {
 	awsSession, err := session.NewSession(&awsgo.Config{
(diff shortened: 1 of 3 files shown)


aws/ecs_cluster.go
```diff
@@ -11,14 +11,63 @@ import (
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
+// Used in this context to determine if the ECS Cluster is ready to be used & tagged
+// For more details on other valid status values: https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#Cluster
+const activeEcsClusterStatus string = "ACTIVE"
+const provisioningEcsClusterStatus = "PROVISIONING"
```
brikis98, line 17: Does the AWS SDK not export these values itself?

aws/ecs_cluster.go
```diff
@@ -11,14 +11,63 @@ import (
...
+
+// filter out all active ecs clusters
```
yorinasub17, line 19: NIT: I believe "filter out" means "get rid of", so this would be the opposite of what this function is doing.

aws/ecs_cluster.go
```diff
@@ -11,14 +11,63 @@ import (
...
+	if err != nil {
+		logging.Logger.Errorf("Error getting all ECS clusters")
+		return nil, errors.WithStackTrace(err)
+	}
+
+	var result []*ecs.Cluster
+
+	for _, cluster := range allClusters {
+		input := &ecs.DescribeClustersInput{
+			Clusters: []*string{cluster},
```
brikis98, line 33: Instead of making one API call per ECS cluster, it looks like `Clusters` accepts a slice, so could you describe all of them? Note: there may be some limit on how many `Clusters` you can query at a time, so you may have to batch these a bit, but it would still be faster than one API call at a time.

aws/ecs_cluster.go
```diff
@@ -11,14 +11,63 @@ import (
...
+		}
+
+		describedClusters, describeErr := svc.DescribeClusters(input)
+		if describeErr != nil {
+			logging.Logger.Errorf("Error describing all ECS clusters")
```
brikis98, line 38:
```suggestion
			logging.Logger.Errorf("Error describing ECS clusters")
```

aws/ecs_cluster.go
```diff
@@ -11,14 +11,63 @@ import (
...
+		result = append(result, describedClusters.Clusters...)
+	}
+
+	var filteredEcsClusterArns []*string
+
+	// Filter out invalid state ECS Clusters (will return only `ACTIVE` state clusters)
+	// `cloud-nuke` needs to tag ECS Clusters it sees for the first time.
+	// Therefore to tag a cluster, that cluster must be in the `ACTIVE` state.
+	for _, cluster := range result {
+		logging.Logger.Infof("Status for ECS CLuster %s is %s", aws.StringValue(cluster.ClusterArn), aws.StringValue(cluster.Status))
```
brikis98, line 51: NIT: Should this be `Debugf` instead of `Infof`?

aws/ecs_cluster_test.go
```diff
@@ -11,18 +11,17 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const region = "eu-west-1"
-
 // Test we can create a cluster, tag it, and then find the tag
 func TestCanTagEcsClusters(t *testing.T) {
 	t.Parallel()
+	region, err := getRandomRegion()
```
brikis98, line 17: Don't forget to check `err`!

aws/ecs_cluster.go
yorinasub17, line 51:
```suggestion
		logging.Logger.Infof("Status for ECS Cluster %s is %s", aws.StringValue(cluster.ClusterArn), aws.StringValue(cluster.Status))
```

aws/ecs_cluster_test.go
```diff
@@ -11,18 +11,17 @@ import (
...
 
 	awsSession, err := session.NewSession(&awsgo.Config{
 		Region: awsgo.String(region),
 	})
 	require.NoError(t, err)
 
-	cluster := createEcsFargateCluster(t, awsSession, util.UniqueID())
+	cluster := createEcsFargateCluster(t, awsSession, "cloud-nuke-test-" + util.UniqueID())
```
brikis98, line 24: Sanity check: is Fargate supported in all AWS regions? If not, you might  have to reduce the list.

aws/ecs_cluster_test.go
```diff
@@ -46,6 +45,7 @@ func TestCanTagEcsClusters(t *testing.T) {
 // Test we can get all ECS clusters younger than < X time based on tags
 func TestCanListAllEcsClustersOlderThan24hours(t *testing.T) {
 	t.Parallel()
+	region, err := getRandomRegion()
```
brikis98, line 48: Check `err`

aws/ecs_cluster.go
```diff
@@ -11,14 +11,63 @@ import (
...
+		}
+
+		result = append(result, describedClusters.Clusters...)
+	}
+
+	var filteredEcsClusterArns []*string
+
+	// Filter out invalid state ECS Clusters (will return only `ACTIVE` state clusters)
+	// `cloud-nuke` needs to tag ECS Clusters it sees for the first time.
+	// Therefore to tag a cluster, that cluster must be in the `ACTIVE` state.
```
yorinasub17, line 49: NIT: This comment is out of date with the code, since this also includes clusters in `PROVISIONING` state.

brikis98 (commented):
- Can clusters in `PROVISIONING` state be tagged?
- `INACTIVE` I believe means deleted, so I think that's safe to ignore.
- Does `DEPROVISIONING` meaning "in the process of being deleted" or something else?
- Do `FAILED` clusters need to be deleted?

yorinasub17 (commented):
Fix makes sense to me! Had a few comments, but mostly minor. The one major issue I see is the API call efficiency.

aws/ecs_cluster_test.go
ina-stoyanova: very good point 👍 thanks

aws/ecs_cluster.go
yorinasub17: Just checked and didn't find it.

aws/ecs_cluster.go
yorinasub17: Lol race condition! Had the same comment, and you can see https://github.com/gruntwork-io/cloud-nuke/pull/146#discussion_r505672148 for a suggested fix.

aws/ecs_cluster.go
ina-stoyanova: I've been trying to find them in the awsgo sdk - but they're only described in a comment. I considered having an enum over it, but wasn't sure whether it's needed at this point. Thoughts?

aws/ecs_cluster_test.go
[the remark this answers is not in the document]
ina-stoyanova: yeah - I discovered why I don't get any notifications if I miss this out basically. It's because I keep naming my error `errz which it thinks is handled if I've handled it anywhere else. How do you deal with this? Do you name every error separately? (that's been my way around it now)

aws/ecs_cluster.go
ina-stoyanova: yep, definitely!

aws/ecs_cluster_test.go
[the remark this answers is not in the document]
brikis98: TBH, I mostly just built muscle memory at this point where, any time I do `err := ...`, I immediately check if `err` is `nil`... But yea, different var names is one way. In some cases, where `err` is the only var on the left you  need, you can do:

```go
if err := someThingThatReturnsError(); err != nil {
  // handle the err
}
```

That way, the `err` is only scoped to that one if-statement, so even though it's the same name, it only exists there, and is checked as part of this one-liner.

ina-stoyanova:
@brikis98 

> * Can clusters in `PROVISIONING` state be tagged? _- I haven't actually found docs supporting this. I have left it in, because of an assumption I made (which was that it's progressing the state to an `active` one). However, on second thought now, I think we should be explicit in which ECS Clusters can be used by cloud-nuke, as it removes ambiguity._ 
> * `INACTIVE` I believe means deleted, so I think that's safe to ignore. _= yes_
> * Does `DEPROVISIONING` meaning "in the process of being deleted" or something else? _= yes_
> * Do `FAILED` clusters need to be deleted? = I'm happy to take guidance here - maybe? 🤔 I guess cloud-nuke is going to be used on purpose, so ideally should not leave too many things behind which are note explicitly ignored. (is my view of this correct?)

brikis98:
> > * Can clusters in `PROVISIONING` state be tagged? _- I haven't actually found docs supporting this. I have left it in, because of an assumption I made (which was that it's progressing the state to an `active` one). However, on second thought now, I think we should be explicit in which ECS Clusters can be used by cloud-nuke, as it removes ambiguity._

In theory, `PROVISIONING` is a temporary state... It will eventually resolve to one of the others. So probably safest to skip it from the `cloud-nuke` perspective. If someone needs to clean it up, they can re-run, and by then, it should be in one of the other states.

> > * Do `FAILED` clusters need to be deleted? = I'm happy to take guidance here - maybe? 🤔 I guess cloud-nuke is going to be used on purpose, so ideally should not leave too many things behind which are note explicitly ignored. (is my view of this correct?)

What causes `FAILED`? Is it a temporary state that eventually disappears, a bit like `INACTIVE`? Or is it more permanent? If it's more permanent, then, yes, they should be deleted, but again, the same filters we use elsewhere, such as age, should apply so there are no surprises. So can we tag `FAILED` clusters?

aws/ecs_cluster_test.go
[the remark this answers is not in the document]
ina-stoyanova: yep, I like this! Thank you, Jim!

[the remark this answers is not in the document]
ina-stoyanova: I think I'll just do my own pre-commit hook to make me double-check a few of these things until I get into a habit.

aws/ecs_cluster.go
```diff
@@ -11,14 +11,54 @@ import (
...
+
+	allClusters, err := getAllEcsClusters(awsSession)
+	if err != nil {
+		logging.Logger.Errorf("Error getting all ECS clusters")
+		return nil, errors.WithStackTrace(err)
+	}
+
+	var filteredEcsClusterArns []*string
+
+	batches := split(aws.StringValueSlice(allClusters), 10)
```
brikis98, line 30: NIT: Why batches of 10? Might want to extract that "magic number" into a `const` with a comment explaining the reasoning.

aws/ecs_cluster.go
```diff
@@ -11,14 +11,54 @@ import (
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
+// Used in this context to determine if the ECS Cluster is ready to be used & tagged
+// For more details on other valid status values: https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#Cluster
+const activeEcsClusterStatus string = "ACTIVE"
+
+// Filter all active ecs clusters
```
brikis98, line 18: NIT: Expand the comment a bit to explain why we only use ACTIVE clusters, as it took us a bit of back and forth to figure that out. 

BTW, what did you find on clusters in FAILED status?

brikis98 (commented):
Very close, just two NITs!

aws/ecs_cluster.go
ina-stoyanova: just realised I did not post my answer - started looking at it, and then jumped on a call. Will add shortly what I've found.

aws/ecs_cluster.go
ina-stoyanova: 😆 good question - I followed strictly the example Yori referred me to. What you've suggested makes sense.

ina-stoyanova:
I spent around 10-15 mins manually creating and breaking ECS Clusters in the AWS console. I've updated the answers below after these experiments & trying to interpret better the AWS docs for the ECS Cluster state.
The only piece of docs I found regarding this is: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/clusters.html#clusters-concepts
and unfоrtunately that is not very descriptive in terms of what gets the cluster in these states. 

> > > * Can clusters in `PROVISIONING` state be tagged? _- I haven't actually found docs supporting this. I have left it in, because of an assumption I made (which was that it's progressing the state to an `active` one). However, on second thought now, I think we should be explicit in which ECS Clusters can be used by cloud-nuke, as it removes ambiguity._
> 
> In theory, `PROVISIONING` is a temporary state... It will eventually resolve to one of the others. So probably safest to skip it from the `cloud-nuke` perspective. If someone needs to clean it up, they can re-run, and by then, it should be in one of the other states.

**NEW comment:**
_There's a case where I was able to actually manually add tags in the web console to an ECS cluster in this state. So we could go back and add this as a covered case in the filtering clusters._ 

> 
> > > * Do `FAILED` clusters need to be deleted? = I'm happy to take guidance here - maybe? 🤔 I guess cloud-nuke is going to be used on purpose, so ideally should not leave too many things behind which are note explicitly ignored. (is my view of this correct?)
> 
> What causes `FAILED`? Is it a temporary state that eventually disappears, a bit like `INACTIVE`? Or is it more permanent? If it's more permanent, then, yes, they should be deleted, but again, the same filters we use elsewhere, such as age, should apply so there are no surprises. So can we tag `FAILED` clusters?

**NEW comment:**
_I didn't manage to get a cluster in this state, so can only interpret what the AWS docs say about it. After trying to tag ECS clusters in the `DEPROVISIONING` and `INACTIVE` states, AWS complains & says that the cluster is `inactive` so I can only assume this `FAILED` state will be similar, where ECS cannot talk to the underlying provider/resources/anything else in the mix._

aws/ecs_cluster.go
ina-stoyanova: Oops, I could swear there was a better comment when the actual filtering is done. I think I accidentally got rid of it while refactoring. I'm bringing it back & would leave it closer to the logic, rather than the const declaration, as it makes it a bit clearer in my view.

brikis98 (approved):
Roger, ship it!

brikis98:
> _There's a case where I was able to actually manually add tags in the web console to an ECS cluster in this state. So we could go back and add this as a covered case in the filtering clusters._
> _I didn't manage to get a cluster in this state, so can only interpret what the AWS docs say about it. After trying to tag ECS clusters in the `DEPROVISIONING` and `INACTIVE` states, AWS complains & says that the cluster is `inactive` so I can only assume this `FAILED` state will be similar, where ECS cannot talk to the underlying provider/resources/anything else in the mix._

It's probably to OK to ignore both of these. One is a temporary state, the other seems quite rare and hard to repro. If we need to clean them up some time in the future, we can always add it, but not worth holding up this PR now for these rare cases.

PR #147, opened 2020-10-31
Bring Go to 1.15

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: /go/src/github.com/gruntwork-io/cloud-nuke
   docker:
-    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.13
+    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.15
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: /go/src/github.com/gruntwork-io/cloud-nuke
   docker:
-    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.13
+    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.15
```
brikis98, line 4: I don't believe we've published such an image yet.

brikis98 (commented):
Thanks for the PR! Is there something in Go 1.15 that you were looking for specifically? We haven't published the corresponding CircleCi image yet, so I suspect the build won't pass. Trying to figure out if this is something timely or not?

.circleci/config.yml
gliptak: @brikis98 just keeping up with tooling. I didn't realize that an intermediate image is a build dependency.

PR #150, opened 2020-11-12
Fix CircleCI Contexts

zackproser:
This pull request was programmatically opened by the multi-repo-updater program. It should be adding the 'Gruntwork Admin' context to any Workflows -> Jobs nodes and should also be leaving the rest of the .circleci/config.yml file alone. 

 This PR was opened so that all our repositories' .circleci/config.yml files can be converted to use the same CircleCI context, which will make rotating secrets much easier in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,6 @@ defaults: &defaults
   working_directory: /go/src/github.com/gruntwork-io/cloud-nuke
   docker:
     - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.13
-
 version: 2
 jobs:
   test:
@@ -12,7 +11,6 @@ jobs:
       - run:
           command: run-go-tests --timeout 45m
           no_output_timeout: 45m
-
   build:
     <<: *defaults
     steps:
@@ -21,7 +19,6 @@ jobs:
       - persist_to_workspace:
           root: .
           paths: bin
-
   nuke_phx_devops:
     <<: *defaults
     steps:
@@ -35,7 +32,6 @@ jobs:
               --force \
               --exclude-resource-type s3
           no_output_timeout: 1h
-
   nuke_sandbox:
     <<: *defaults
     steps:
@@ -51,7 +47,6 @@ jobs:
               --force \
               --exclude-resource-type s3
           no_output_timeout: 1h
-
   deploy:
     <<: *defaults
     steps:
@@ -59,7 +54,6 @@ jobs:
           at: .
       - run: cd bin && sha256sum * > SHA256SUMS
       - run: upload-github-release-assets bin/*
-
 workflows:
   version: 2
   build-and-test:
@@ -68,12 +62,16 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+          context:
+            - Gruntwork Admin
       - build:
           requires:
             - test
           filters:
             tags:
               only: /^v.*/
+          context:
+            - Gruntwork Admin
       - deploy:
           requires:
             - build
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -96,7 +94,8 @@ workflows:
       - nuke_phx_devops:
           requires:
             - test
-
+          context:
```
brikis98, line 97: There is a `test` job defined above, on line 95, that also needs the `context`, but it doesn't look like it was added? Perhaps because `test` has no child nodes?

.circleci/config.yml
```diff
@@ -109,3 +108,5 @@ workflows:
       - nuke_sandbox:
           requires:
             - test
+          context:
```
brikis98, line 111: Same issue with `test` on line 108

.circleci/config.yml
zackproser: I see what you mean and I can repro this locally with YQ.

zackproser: Yes - I believe you are correct. After fiddling with this for a while, I think what's happening is that the [path expression](https://mikefarah.gitbook.io/yq/usage/path-expressions) I'm currently using assumes that jobs found under `workflows` will be either an array or a map - and the syntax supports accessing either of those data types and appending to (or creating) the `context` array that should be there. 

I think I may need a separate query that detects these single string jobs and replaces them with maps of the same name and containing a context key - in a second pass.

zackproser: This also *might* fall under the category of small annoyances we'll need to tolerate for the first pass and fix manually - I can't currently see a path to selecting or updating nodes by their type in `yq` docs.

brikis98: Ah, I see. I have two ideas for you to consider:

1. Post an issue in https://github.com/mikefarah/yq/issues to see if they have a suggestion on how to update the non object nodes below `workflows`.
1. Hacky solution: using an expression like ` yq r config.yml --printMode pv 'workflows.*.jobs.[*]', you get output like:
    ```
    workflows.build-and-test.jobs.[0]: test
    workflows.build-and-test.jobs.[1]:
      build:
        requires:
          - test
        filters:
          tags:
            only: /^v.*/
    workflows.build-and-test.jobs.[2]:
      deploy:
        requires:
          - build
        filters:
          tags:
            only: /^v.*/
          branches:
            ignore: /.*/
    workflows.frequently.jobs.[0]: test
    workflows.frequently.jobs.[1]:
      nuke_phx_devops:
        requires:
          - test
    workflows.nightly.jobs.[0]: test
    workflows.nightly.jobs.[1]:
      nuke_sandbox:
        requires:
          - test
    ```
    If you parse that with Go and pull the entries that have no children (e.g., `workflows.frequently.jobs.[0]: test`) into a list, you can loop over each item `x` in that list and do `yq w config.yml 'x.context[+]' 'Gruntwork Admin'` (e.g., `yq w config.yml 'workflows.build-and-test.jobs.[0].test.context[+]' 'Gruntwork Admin').

zackproser: Awesome- thanks, @brikis98  I'll get #1 cooking and in the meanwhile take a stab at playing around with option #2 as well.

zackproser: https://github.com/mikefarah/yq/issues/586

zackproser: Mike got back to us quickly and confirmed that this is not currently supported - the forthcoming `v4` rewrite of `yq` will support this and more - so I will forge ahead with an implementation of option 2.

brikis98: Roger!

zackproser: Fixed in https://github.com/gruntwork-io/prototypes/commit/11ebf9f18eb1b467544ba9831cf5c1f8fa69d510

zackproser:
Preparing to re-run the batch that includes this repo now that the bugs have been fixed.

PR #152, opened 2020-11-19
Fix CircleCI Contexts

zackproser:
This pull request was programmatically opened by the multi-repo-updater program. It should be adding the 'Gruntwork Admin' context to any Workflows -> Jobs nodes and should also be leaving the rest of the .circleci/config.yml file alone. 

 This PR was opened so that all our repositories' .circleci/config.yml files can be converted to use the same CircleCI context, which will make rotating secrets much easier in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,6 @@ defaults: &defaults
   working_directory: /go/src/github.com/gruntwork-io/cloud-nuke
   docker:
     - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.13
-
 version: 2
 jobs:
   test:
@@ -12,7 +11,6 @@ jobs:
       - run:
           command: run-go-tests --timeout 45m
           no_output_timeout: 45m
-
   build:
     <<: *defaults
     steps:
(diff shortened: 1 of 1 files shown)


brikis98 (approved):
The code changes LGTM! 

However, I just realized that `cloud-nuke` has several sets of credentials (one for Phx DevOps, one for sandbox) and the `Gruntwork Admin` context only has the Phx DevOps ones for now... But I can add those in using the same env vars after this is merged.

zackproser:
> The code changes LGTM!
> 
> However, I just realized that `cloud-nuke` has several sets of credentials (one for Phx DevOps, one for sandbox) and the `Gruntwork Admin` context only has the Phx DevOps ones for now... But I can add those in using the same env vars after this is merged.

Thanks @brikis98 - do we need to coordinate on this timing to avoid any failures, issues?

brikis98:
I think it'll work fine to merge this... When, later, I go through and remove the env vars from each project, I'll need too remember to update Gruntwork Admin context accordingly.

zackproser:
> I think it'll work fine to merge this... When, later, I go through and remove the env vars from each project, I'll need too remember to update Gruntwork Admin context accordingly.

Roger - merging!

PR #154, opened 2020-11-30
Remove contexts from scheduled builds

brikis98:
CircleCi builds that run on a schedule run on behalf of some CircleCi machine user, and that machine user does not have access to our CircleCi contexts. As a result, all the builds show up as "unauthorized":

Screen Shot 2020-11-30 at 9 41 18 AM


marinalimeira (approved):

brikis98:
Thanks for the review! Merging now.

Issue #116, opened 2020-05-17
Feature Request: add ability to delete users, groups, and policies


queglay:
For each deployment and test I create new users, with a pipeline id in the user name, since tags can't be used.

It would be awesome if we could use a regex to clean out resources like users if they matched.

brikis98:
A PR to add support for deleting IAM resources (users, groups, policies) is welcome. See also https://github.com/gruntwork-io/cloud-nuke/issues/108 for more ability to match on regex and other more complex patterns.

marinalimeira:
As of today, we have the ability to nuke IAM Users and they can be filtered using the config file. I am leaving this issue open because we still don't support IAM Groups and Policies.

queglay:
A great step nonetheless, thanks!

PR #160, opened 2020-12-08
Nuke IAM Users

hposca:
Attempt to solve https://github.com/gruntwork-io/cloud-nuke/issues/116 

Currently it only tries to delete IAM Users.

Please look at each commit for a better understanding of the steps.

### DIFF
--- a/aws/iam.go
+++ b/aws/iam.go
@@ -9,6 +9,7 @@ import (
 
 // List all IAM users in the AWS account and returns a slice of the UserNames
 // TODO: Implement exclusion by time filter
+// TODO: AWS IAM is global, specifying a region doesn't make sense and creates duplicated output
 func getAllIamUsers(session *session.Session, region string) ([]*string, error) {
 	svc := iam.New(session)
 	input := &iam.ListUsersInput{}
--- a/aws/iam.go
+++ b/aws/iam.go
@@ -6,7 +6,7 @@ import (
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
-// Returns a slice of IAM UserNames
+// List all IAM users in the AWS account and returns a slice of the UserNames
 // TODO: Implement exclusion by time filter
 func getAllIamUsers(session *session.Session, region string) ([]*string, error) {
 	svc := iam.New(session)
--- a/aws/iam.go
+++ b/aws/iam.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/gruntwork-io/cloud-nuke/logging"
 	"github.com/gruntwork-io/gruntwork-cli/errors"
 )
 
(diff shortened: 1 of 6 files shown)


aws/iam_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestListIamUsers(t *testing.T) {
+	t.Parallel()
+
+	region, err := getRandomRegion()
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
yorinasub17, line 19: NOTE: we've been converting these checks to `require.NoError(t, err)`, so that it short circuits the test when there is an error. Also reduces these 3 line checks to 1 line.

aws/iam_test.go
```diff
@@ -0,0 +1,41 @@
...
+	// TODO: Implement exclusion by time filter
+	// userNames, err := getAllIamUsers(session, region, time.Now().Add(1*time.Hour*-1))
+	userNames, err := getAllIamUsers(session, region)
+	if err != nil {
+		assert.Fail(t, "Unable to fetch list of IAM users")
+	}
+
+	// TODO: Remove this, just for temporary visual confirmation
+	for _, name := range userNames {
+		fmt.Printf("this is the *name: %s\n", *name)
```
yorinasub17, line 37: I know this is just debug logs, but a friendly tidbit: you can use `awsgo.StringValue` for a safe string pointer dereference. This function will return `""` for `nil` instead of panicking.

aws/iam.go
```diff
@@ -0,0 +1,28 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+// Returns a slice of IAM UserNames
```
yorinasub17, line 9: NIT:

```suggestion
// List all IAM users in the AWS account and returns a slice of the UserNames
```

yorinasub17 (commented):
This looks like a great start! I only had a few style nits and minor issues so far.

---

MAINTAINER'S NOTE: we will most likely also want to add support for the config file prior to merge - https://github.com/gruntwork-io/cloud-nuke/#config-file

IAM users are one of those resources where every account (even for test) will always have a few important users that we don't want to delete, so the conservative include and exclude listing features of the Config file will be very important.

aws/iam_test.go
```diff
@@ -0,0 +1,37 @@
...
+		Region: awsgo.String(region)},
+	)
+	require.NoError(t, err)
+
+	// TODO: Implement exclusion by time filter
+	// userNames, err := getAllIamUsers(session, region, time.Now().Add(1*time.Hour*-1))
+	userNames, err := getAllIamUsers(session, region)
+	if err != nil {
+		assert.Fail(t, "Unable to fetch list of IAM users")
+	}
```
yorinasub17, line 29: NIT: Looks like there's one more error check to convert!

hposca: oops

aws/iam.go
```diff
@@ -0,0 +1,59 @@
...
+	svc := iam.New(session)
+
+	for _, userName := range userNames {
+		input := &iam.DeleteUserInput{
+			UserName: userName,
+		}
+
+		_, err := svc.DeleteUser(input)
+		if err != nil {
+			logging.Logger.Errorf("[Failed] %s", err)
```
yorinasub17, line 50: We're not consistent about what should happen here, whether to bubble errors or swallow them. I think we are ok with swallowing them for now, but we should address this at some point.

I filed https://github.com/gruntwork-io/cloud-nuke/issues/162 so we have this point in the issue tracker.

yorinasub17 (commented):
Updates LGTM! Solid progress!

Just leaving a reminder here that in addition to the filtering routine, we will also need to hook it up to the main nuking CLI routine before merging the PR.

yorinasub17 (commented):
Solid progress! I didn't have any comments to add here outside of the TODOs you already captured.

Issue #167, opened 2021-01-01
Deleting DynamoDB tables


fredrb:
I've been using `cloud-nuke` to purge all resources created in test AWS accounts. One of the projects creates a bunch of DynamoDB tables, and I realized that DynamoDB deletion is not supported. 

Is this something you plan to implement? Or is not implemented on purpose? 

Thanks for the project, it's pretty useful. :+1:

DMEvanCT:
I think you can do this by calling list tables api and adding each table to a slice to be deleted. Looking into this.

DMEvanCT:
@fredrb  found out how to do this. Working on the feature now.

brikis98:
A PR to add this is very welcome!

lplazas:
@brikis98 @DMEvanCT  is this something that hasn't been implemented? If so I'd gladly like to have a go at it

DMEvanCT:
I hade a PR for this but there were some requested modifications I have not had time to take a look. Go ahead though!

DMEvanCT:
@fredrb This exists now! Let me know if this solves your problem!

infraredgirl:
Fixed in #169.

infraredgirl:
https://github.com/gruntwork-io/cloud-nuke/releases/tag/v0.5.2

PR #169, opened 2021-01-04
Feature/dynamodb nuke

DMEvanCT:


DMEvanCT:
This was in reference to issue #167

DMEvanCT:
INFO[2021-01-03T23:55:54-05:00] Checking region [15/16]: us-west-1           
INFO[2021-01-03T23:55:55-05:00] Checking region [16/16]: us-west-2           
INFO[2021-01-03T23:55:55-05:00] The following 1 AWS resources will be nuked: 
INFO[2021-01-03T23:55:55-05:00] * dynamodb Table1 us-east-1
                 

THE NEXT STEPS ARE DESTRUCTIVE AND COMPLETELY IRREVERSIBLE, PROCEED WITH CAUTION!!!

Are you sure you want to nuke all listed resources? Enter 'nuke' to confirm (or exit with ^C): nuke
INFO[2021-01-03T23:55:58-05:00] Terminating 1 resources in batches           
INFO[2021-01-03T23:55:58-05:00] Deleting all EBS volumes in region us-east-1

DMEvanCT:
@yorinasub17  This worked. Let me know if I missed documentation anywhere.

aws/dynamodb.go
```diff
@@ -0,0 +1,71 @@
...
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Error() {
+				case dynamodb.ErrCodeInternalServerError:
+					return nil, errors.WithStackTrace(aerr)
+				default:
+					return nil, errors.WithStackTrace(aerr)
+				}
+			} else {
+				return nil, errors.WithStackTrace(aerr)
+			}
```
brikis98, line 29: It seems like in all cases here, you return `errors.WithStackTrace(err)`... So is there a need for the if/else statements?

aws/dynamodb.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func getAllDynamoTables(session *session.Session) ([]*string, error) {
+	var TableNames []*string
+	svc := dynamodb.New(session)
+
+	for {
+		result, err := svc.ListTables(&dynamodb.ListTablesInput{})
```
brikis98, line 17: Does this API call need pagination?

aws/aws.go
```diff
@@ -424,7 +424,21 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 			}
 		}
 		// End Lambda Functions
+		// Dynamo DB tables
+		DynamoDB := DynamoDB{}
+		if IsNukeable(DynamoDB.ResourceName(), resourceTypes) {
+			tablenames, err := getAllDynamoTables(session)
```
brikis98, line 430: With `cloud-nuke`, we need to respect the `older-than` parameter... It looks like this nukes all DynamoDB tables, rather than those older than the date requested by the user?

aws/dynamodb.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+)
+
+func getAllDynamoTables(session *session.Session) ([]*string, error) {
+	var TableNames []*string
```
brikis98, line 13: Please use lower case for private variables: e.g., `tableNames`. Go uses upper case for public variables.

aws/dynamodb.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func nukeAllDynamoDBTables(session *session.Session, tables []*string) error {
+	svc := dynamodb.New(session)
+	if len(tables) == 0 {
+		logging.Logger.Infof("No EBS volumes to nuke in region %s", *session.Config.Region)
+		return nil
+	}
+
+	logging.Logger.Infof("Deleting all EBS volumes in region %s", *session.Config.Region)
```
brikis98, line 47: Copy paste issue: replace "EBS volumes" with "DynamoDB tables"

aws/dynamodb.go
```diff
@@ -0,0 +1,71 @@
...
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Error() {
+				case dynamodb.ErrCodeInternalServerError:
+					return errors.WithStackTrace(aerr)
+				default:
+					return errors.WithStackTrace(aerr)
+				}
+			} else {
+				return errors.WithStackTrace(aerr)
+			}
```
brikis98, line 65: Same question here about the need for if/else statements.

aws/dynamodb_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func CreateTestDynamoTables(t *testing.T, session *session.Session) error {
+	svc := dynamodb.New(session)
+	tablestocreate := []string{"Table1", "Table2"}
```
brikis98, line 14: These table names need to include a unique, randomly generated ID to avoid conflicts when there are many tests running in parallel. See some of the other tests for the `UniqueId()` function we use for this.

brikis98, line 14: Also, please use camelCase: `tablesToCreate`

aws/dynamodb_test.go
```diff
@@ -0,0 +1,38 @@
...
+	for _, table := range tablestocreate {
+		input := &dynamodb.CreateTableInput{TableName: aws.String(table)}
+		_, err := svc.CreateTable(input)
+		if err != nil {
+			assert.Fail(t, errors.WithStackTrace(err).Error())
+		}
```
brikis98, line 20:
```suggestion
		require.NoError(t, err)
```

aws/dynamodb_test.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	return nil
+}
+
+func NukeAllDynamoDBTablesTest(t *testing.T, session *session.Session) error {
+	tables, err := getAllDynamoTables(session)
+	t.Log(&tables)
```
brikis98, line 28: Remove

aws/dynamodb_test.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
brikis98, line 31:
```suggestion
	require.NoError(t, err)
```

aws/dynamodb_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	error := nukeAllDynamoDBTables(session, tables)
```
brikis98, line 33: `error` is the name of the interface... Use `err` or, if that's already used, `nukeErr` or similar instead.

aws/dynamodb_test.go
```diff
@@ -0,0 +1,38 @@
...
+	if error != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
brikis98, line 36:
```suggestion
	require.NoError(t, error)
```

aws/dynamodb_test.go
brikis98, line 33: The nuke call needs to be in a `defer` close to the top of the function to ensure it always runs, even if other errors happened along the way.

brikis98 (commented):
Thanks for the PR!

calebtote:
Any chance of this getting reviewed soon?

brikis98:
I can re-review. Could you pull in the latest from `master` to fix the merge conflict?

DMEvanCT:
I sure can!

DMEvanCT:
@brikis98 Master has been pulled back into my branch

aws/dynamodb.go
brikis98: Bump

aws/dynamodb.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/gruntwork-cli/errors"
+	"time"
+)
+
+func getAllDynamoTables(session *session.Session, excludeAfter time.Time) ([]*string, error) {
```
brikis98, line 13: I just realized this method is not filtering the tables based on the `excludeAfter` param. That is a must-have.

aws/dynamodb_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}))
+	tables := 0
+	var tablesToCreate []string
+	// Creates 20 tables to simulate a smaller environment. We can make this dynamic if needed.
+	for tables <= 20 {
+
+		tablesToCreate = append(tablesToCreate, "dynamo-nuke-test-" + random.UniqueId() )
+		tables +=1
+
+	}
```
brikis98, line 28: NIT: Please remove extra whitespace and run `go fmt` on this code.

aws/dynamodb_test.go
```diff
@@ -0,0 +1,101 @@
...
+				},
+			},
+			ProvisionedThroughput: &dynamodb.ProvisionedThroughput{
+				ReadCapacityUnits:  aws.Int64(1),
+				WriteCapacityUnits: aws.Int64(1),
+			},
+
+		}
+		_, err := svc.CreateTable(input)
+		time.Sleep(2 * time.Second)
```
brikis98, line 63: Why `Sleep`? Is this to account for tables being created async? If so, then we should use a loop to explicitly check if the API knows about the table, rather than sleep, as sleep will fail intermittently.

aws/dynamodb_test.go
```diff
@@ -0,0 +1,101 @@
...
+	awsSession := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+
+	createTestDynamoTables(t)
+
+	tables, err := getAllDynamoTables(awsSession, time.Now().Add(1*time.Hour*-1))
+	if err != nil {
+		require.Error(t, err)
+		}
```
brikis98, line 96:
```suggestion
	require.NoError(t, err)
```

aws/dynamodb_test.go
```diff
@@ -0,0 +1,101 @@
...
+		// Dynamo needs a wait of about 2 seconds to create the tables before delete
+		time.Sleep(2 * time.Second)
```
brikis98, line 98: See comment above

aws/dynamodb_test.go
brikis98: You should also check that `tables` contains the 20 tables you expect.

aws/dynamodb_test.go
```diff
@@ -0,0 +1,101 @@
...
+	nukeErr := nukeAllDynamoDBTables(awsSession, tables)
+	require.NoError(t, nukeErr)
```
brikis98, line 100: After calling nuke, we should call `getAllTables` and ensure they no longer contain the 20 you created.

brikis98: Also, there should be a `nukeAllDynamoDBTables` call in a `defer` early in the test to ensure that's always called, even if some other part of the test failed.

aws/dynamodb_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	createTestDynamoTables(t)
+
+	tables, err := getAllDynamoTables(awsSession, time.Now().Add(1*time.Hour*-1))
+	if err != nil {
+		require.Error(t, err)
+		}
+		// Dynamo needs a wait of about 2 seconds to create the tables before delete
+		time.Sleep(2 * time.Second)
+	nukeErr := nukeAllDynamoDBTables(awsSession, tables)
```
brikis98, line 99: In this test, you should be deleting solely the tables you created. So pass in the list of table names you created, rather than those fetched from the API, as the latter may include other tables that happen to be in the AWS account.

ina-stoyanova:
@DMEvanCT bumping this PR. Is there a way we can update with the latest from master and help you get this PR merged in?

DMEvanCT:
@ina-stoyanova @brikis98  I think I got all the changes you requested. Let me know if I need to update anything else.

ina-stoyanova:
Thanks @DMEvanCT I'll get to review this one today!

aws/dynamodb_test.go
```diff
@@ -0,0 +1,126 @@
...
+	return result.Table.TableStatus
+
+}
+
+func TestGetTablesDynamo(t *testing.T) {
+	t.Parallel()
+	region, err := getRandomRegion()
+	if err != nil {
+		assert.Fail(t, errors.WithStackTrace(err).Error())
+	}
```
ina-stoyanova, line 79: could we do `require.NoError()` here?

My reasoning is we want to fail as soon as possible if there's an error (since there's no point sending any calls to AWS if there's any prior error). If my understanding is correct here `assert.Fail` will actually assert that the test is failed at the end. `NoError()` will fail it now.

aws/dynamodb.go
ina-stoyanova: Bumping this one too. It would be best to have pagination in place, since it's already supported by default in the ListTables API method too. The maximum number of DynamoDB table names returned seems to be 100 per query. https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_ListTables.html

Could we add that?

aws/dynamodb.go
```diff
@@ -0,0 +1,83 @@
...
+		if err != nil {
+			//
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Error() {
+				case dynamodb.ErrCodeInternalServerError:
+					return errors.WithStackTrace(aerr)
+				default:
+					return errors.WithStackTrace(aerr)
+				}
+			}
```
ina-stoyanova, line 77: NIT: cleanup empty comment `//` & fix indentation.

aws/dynamodb_types.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func (tables DynamoDB) ResourceIdentifiers() []string {
+	return tables.DynamoTableNames
+}
+
+func (table DynamoDB) MaxBatchSize() int {
+	// Tentative batch size to ensure AWS doesn't throttle
+	return 50
+}
```
ina-stoyanova, line 24: It seems like you've got half of the pagination logic in place too!

DMEvanCT:
> Thanks for picking this one back up! @DMEvanCT! I've left a few comments, but overall it looks really good!
> 
> I think the major thing is to add pagination to the `list-tables` API call.

The API defaults to 100 tables if not specified. We can tell the code if there are = 100 tables run the list tables again until it's under 100 then continue?

DMEvanCT:
@ina-stoyanova I fixed this and added pagination. Please check now!

ina-stoyanova:
Thanks a lot @DMEvanCT Having a look now!

aws/dynamodb.go
```diff
@@ -11,12 +11,18 @@ import (
 	"time"
 )
 
+const maxBatchSizeDynamo = 100
```
ina-stoyanova: nice!

aws/dynamodb_test.go
```diff
@@ -0,0 +1,120 @@
...
+	defer nukeAllDynamoDBTables(awsSession, []*string{&tableName})
+	createTestDynamoTables(t, tableName)
+	COUNTER := 0
+	for COUNTER <= 1 {
+		tableStatus := getTableStatus(tableName)
+		if *tableStatus == "ACTIVE" {
+			COUNTER += 1
+			log.Printf("Created a table: %v\n", tableName)
+		} else {
+			log.Printf("Table no ready yet: %v", tableName)
```
ina-stoyanova, line 103:
```suggestion
			log.Printf("Table not ready yet: %v", tableName)
```

aws/dynamodb.go
```diff
@@ -0,0 +1,97 @@
...
+func getAllDynamoTables(session *session.Session, excludeAfter time.Time, db DynamoDB) ([]*string, error) {
+	var tableNames []*string
+	svc := dynamodb.New(session)
+	// Run count is used for pagination if the list tables exceeds max value
+	// Tells loop to rerun
+	var runCount = 1
+	for runCount > 0 {
+		result, err := svc.ListTables(&dynamodb.ListTablesInput{Limit: aws.Int64(int64(DynamoDB.MaxBatchSize(db)))})
+		if err != nil {
+			//
```
ina-stoyanova, line 23:
```suggestion
```

ina-stoyanova (dismissed):
Overall, I believe we've got a good implementation here. Thanks for all the effort so far! 

I've left a few suggestions to make sure that the spacing is a little bit better - there's still a little bit of cleanup to do, but that's no blocker for the PR. 
- The naming in places is a little bit unintuitive, so I've left a comment there. 
- the use of comments I think helps
- it does make me wonder though if we used a little bit better names, or tried to follow more the logging & the counters such as [here](https://github.com/gruntwork-io/cloud-nuke/blob/a69326ba271f6adaf82debd17d6456991400a12d/aws/s3.go#L95-L146) it would a little bit clearer and without the need to explain why we need to decrement the `runCount` variable. 
All in all, I think we're in a good position here. @marinalimeira @infraredgirl I am curious to see if you think we've missed anything here with @DMEvanCT 

The next steps would be to run the tests against this PR & get it merged in!

DMEvanCT:
@ina-stoyanova cleaned up those recommendations also added a slice clean as the slice will need to be empty per iteration.

ina-stoyanova (approved):

aws/dynamodb_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"log"
+	"testing"
+	"time"
+)
+
+func createTestDynamoTables(t *testing.T, tableName string) {
+	awsSession := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+
```
ina-stoyanova, line 20: I believe you'll need to change thins too.

DMEvanCT:
@ina-stoyanova  I think I fixed it! Thanks for working with me on this!!!

ina-stoyanova:
Hey @DMEvanCT it doesn't actually look like master is failing for me. The problem that I see is the generation of the binaries that happens at the last step of the ci workflow. I haven't had time to dig into why this is happening yet, but hope to find some time this week! 

Thanks for the patience! My goal is to try and get this merged & released asap!

marinalimeira:
@brikis98 do you want to take another look before we merge?

ina-stoyanova (approved):
> Hey @DMEvanCT thanks as always for the prompt updates! We'll test it out with Marina tomorrow morning first thing!

Hey @DMEvanCT thanks a lot for your patience! Let's merge this! 🚢

infraredgirl:
https://github.com/gruntwork-io/cloud-nuke/releases/tag/v0.5.2

Issue #164, opened 2020-12-15
Docker version of cloud-nuke


ghost:
I would like a dockerized version of cloud-nuke.

I'm planning to run it something like this:

    docker run -it --rm -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN gruntwork-io/cloud-nuke:0.1.24 aws

...thus passing the AWS related environment variable from the current shell into the docker container.

I'm happy to contribute the work and send a PR, if you think this is a good idea

brikis98:
This does seem like a handy thing to have. For the time being, a PR to add a simple `Dockerfile` would be great. We'll eventually want to hook up the CI / CD machinery to publish the `Dockerfile` to a registry on release, but that may have to be added later.

sev-obviux:
Cool, I'll make a PR for just that :)

sev-obviux:
I've created a PR for you: https://github.com/gruntwork-io/cloud-nuke/pull/170
I hope you like it.

ina-stoyanova:
This is still a valid issue. The PR was closed due to inability to devote time and fully test the new features. It can be taken forward, so leaving this issue open.

hongil0316:
No update for the past 2 years. Don't expect this to be prioritized so closing this issue for now. We can reopen this later if we need this in the future.

PR #170, opened 2021-01-04
Docker version of cloud-nuke

sev-obviux:
Here is my take on a docker version of cloud-nuke, as per the issue https://github.com/gruntwork-io/cloud-nuke/issues/164

I've added a Dockerfile, a .dockerignore file and a small shell script to build and verify the binary.


DMEvanCT:
@sev-obviux  I would add an args function to allow people to do things like excludes.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ARG VERSION=0.0.0
+RUN go build -o dest/cloud-nuke -ldflags="-X main.VERSION=$VERSION"
+
+####### Runner ########
+#Choosing alpine base as runner, since we don't need the go compiler to run the program
+FROM alpine:3 as runner
+
+#Set working directory
+WORKDIR /wrk
+#Create non-root user, to make the containersafer to run
```
brikis98, line 27:
```suggestion
#Create non-root user, to make the container safer to run
```

README.md
```diff
@@ -301,6 +301,10 @@ cd aws
 go test -v -run TestListAMIs
 ```
 
+### Building and running in Docker
+
+In order to build and run as a docker container, have a look at the instructions at the top of the Dockerfile in this repository.
```
brikis98, line 306: NIT: Any reason to have the docs in the `Dockerfile` rather than the README? Seems like it would be a bit easier to read in a README with Markdown rendering, syntax highlighting, etc.

build_docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export CIRCLE_TAG=1.0.0
+DOCKER_BUILDKIT=1 docker build --compress --build-arg "VERSION=$CIRCLE_TAG" --tag "gruntwork-io/cloud-nuke:$CIRCLE_TAG" --tag "gruntwork-io/cloud-nuke:latest" .
+docker run -it --rm -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN gruntwork-io/cloud-nuke --version | grep "$CIRCLE_TAG" && echo "TEST OK: Version numbers match up"
```
brikis98, line 6: To actually have this test run as part of our test suite, could you convert this to a Go test?

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+#Preparing working directory and copying files
+WORKDIR /wrk
+COPY . . 
+
+#You can overrride by setting the CIRCLE_TAG environment variable. 0.0.0 is just a default, so you'll notice if you forget
+ARG VERSION=0.0.0
+RUN go build -o dest/cloud-nuke -ldflags="-X main.VERSION=$VERSION"
+
+####### Runner ########
+#Choosing alpine base as runner, since we don't need the go compiler to run the program
```
brikis98, line 22: Nice use of multi-stage builds 👍

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+FROM alpine:3 as runner
+
+#Set working directory
+WORKDIR /wrk
+#Create non-root user, to make the containersafer to run
+RUN addgroup -S app && adduser -S -G app app
+#Own the working folder
+RUN chown app:app . 
+#Switch to the new non-root user
+USER app
```
brikis98, line 32: Nice 👍

brikis98 (commented):
Thanks for the PR!

Dockerfile
sev-obviux: Thanks. WIll change

README.md
sev-obviux: I guess i just started with the Dockerfile and only added the read me after. I'm happy to change it, and move docs to the readme.

build_docker.sh
sev-obviux: I'm not familiar with Go. If you point me to a good test to use as a starting point, I'd like to take a stab at it

Dockerfile
sev-obviux: :)

Dockerfile
sev-obviux: Yeah, an important detail. I guess, with a CLI tool like this it's not so critical, but it's a habit.

sev-obviux:
I'll follow your suggestions and update the PR and let you know

sev-obviux:
I've updated the PR with my response to your feedback.
I'm not comfortable with writing a Go test yet, and cannot really justify the time for it. Sorry.

I had an idea while refactoring.... do you want to have all the docker things in a subfolder instead of the root?

brikis98:
> I've updated the PR with my response to your feedback.
> I'm not comfortable with writing a Go test yet, and cannot really justify the time for it. Sorry.

I understand. Unfortunately, if it's not tested in our CI build, it will likely break over time, so I don't think we could merge it.

> I had an idea while refactoring.... do you want to have all the docker things in a subfolder instead of the root?

That seems like a good idea!

sev-obviux:
> I understand. Unfortunately, if it's not tested in our CI build, it will likely break over time, so I don't think we could merge it.

I see. So what do we do here? Do you or anyone on your team want to make the unit test?
I can make it easy, so there's a unit test bash script that you just run and then check that the return code is 0

If you agree to that, then I'll make the unit test script and put it in docker/unit-test.sh

brikis98:
> > I understand. Unfortunately, if it's not tested in our CI build, it will likely break over time, so I don't think we could merge it.
> 
> I see. So what do we do here? Do you or anyone on your team want to make the unit test?
> I can make it easy, so there's a unit test bash script that you just run and then check that the return code is 0
> 
> If you agree to that, then I'll make the unit test script and put it in docker/unit-test.sh

As you can see in the delays in my replies, we are super buried right now, and unfortunately, can't devote as much time as we'd like to `cloud-nuke`. Even adding a simple test can turn into a long yak shave that derails some of our other work. Sorry about that 😞

sev-obviux:
> > > I understand. Unfortunately, if it's not tested in our CI build, it will likely break over time, so I don't think we could merge it.
> > 
> > 
> > I see. So what do we do here? Do you or anyone on your team want to make the unit test?
> > I can make it easy, so there's a unit test bash script that you just run and then check that the return code is 0
> > If you agree to that, then I'll make the unit test script and put it in docker/unit-test.sh
> 
> As you can see in the delays in my replies, we are super buried right now, and unfortunately, can't devote as much time as we'd like to `cloud-nuke`. Even adding a simple test can turn into a long yak shave that derails some of our other work. Sorry about that 😞

I understand, that happens to me all the time too. No hard feelings.

Would you like me to cancel the PR or do you want to keep it around for whenever?

brikis98:
Let's keep it open in case someone in the future has the time to fill in the gaps! If we close now, we'll prob forget it about it otherwise, and we can always close later.

sev-obviux:
Deal :)

sev-obviux:
I'm cleaning up my repos. This PR is so old that it now has conflicts with master and I cannot priorititze fixing it. I'm cancelling this PR and nuking my fork, to clean up.

ina-stoyanova:
Thanks @sev-obviux for your contribution despite not being able to support you fully on it!

PR #181, opened 2021-03-05
Nuke IAM Users

hposca:
Closes #160 , Implements #116 

Now we can delete IAM Users, configuring inclusion and exclusion rules on the configuration file and having a clean output when deleting IAM Users (as they are global and not region specific).

Please read each commit message for more details.

All comments are welcomed.


config/mocks/iam_users_all.yaml
```diff
@@ -0,0 +1,4 @@
+IAMUsers:
+  include:
+    names_regex:
+      - .*
```
ina-stoyanova, line 4: I couldn't see anywhere that this mock was used. Is it safe to have it in here as well - e.g. what's the likelihood that we accidentally delete all our IAM users by using this mock?

README.md
```diff
@@ -190,6 +194,26 @@ s3:
...
+As an example, in the following config file, a user named `some-nice-user-name` would be deleted but a user named `interesting-user-name` would not.
+
+```yaml
+IAMUsers:
+  include:
+    names_regex:
+      - ^some-.*-user-name$
+      - .*-cool-name-.*
+```
+
```
ina-stoyanova, line 216: Good docs - reads very easy!

aws/aws.go
```diff
@@ -478,6 +487,44 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
...
+	)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	globalResources := AwsRegionResource{}
+
+	// IAM Users
+	iamUsers := IAMUsers{}
+	if IsNukeable(iamUsers.ResourceName(), resourceTypes) {
```
ina-stoyanova, line 511: NIT: I have had a look at the whole function, and it is getting very long. It's not for this PR, but was wondering if we could somehow refactor the overall function `GetAllResources`?

aws/iam_types.go
```diff
@@ -0,0 +1,36 @@
...
+	return u.UserNames
+}
+
+// Tentative batch size to ensure AWS doesn't throttle
+func (u IAMUsers) MaxBatchSize() int {
+	return 200
+}
+
+// Nuke - nuke 'em all!!!
+func (u IAMUsers) Nuke(session *session.Session, identifiers []string) error {
```
ina-stoyanova, line 30: could we name this explitily as `users`

ina-stoyanova: same applies to the function params above too!

config/config.go
```diff
@@ -1,47 +1,51 @@
...
-	ExcludeRule FilterRule
-}
+// UnmarshalText - Internally used by yaml.Unmarshal to unmarshall an Expression field
+func (expression *Expression) UnmarshalText(data []byte) error {
+	var pattern string
 
-// FilterRule - contains regular expressions or plain text patterns
-// used to match against a resource type's properties
-type FilterRule struct {
-	NamesRE []*regexp.Regexp
```
ina-stoyanova, line 40: NIT: could we also use a more descriptive name here: `NamesRegExp` would do fine too - I just found myself wondering later on what this was.

aws/iam.go
```diff
@@ -0,0 +1,66 @@
...
+
+// List all IAM users in the AWS account and returns a slice of the UserNames
+// TODO: Implement exclusion by time filter
+func getAllIamUsers(session *session.Session, configObj config.Config) ([]*string, error) {
+	svc := iam.New(session)
+	input := &iam.ListUsersInput{}
+
+	var userNames []*string
+
+	// TODO: Probably use ListUsers together with ListUsersPages in case there are lots of users
```
ina-stoyanova, line 19: Question: is there any number you're thinking of here? You could also probably use `Marker` and `MaxItems` - see [here](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListUsers.html).

ina-stoyanova (commented):
Looks good! 🙌 
I've left a couple of comments/questions, but nothing that should block you hopefully! I've also noticed there's a couple of TODOs that are still around here (the exclusion time filter & listPages), so I imagine this is still WIP 🙂

aws/aws.go
```diff
@@ -194,6 +198,11 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 	var resourcesCache = map[string]map[string][]*string{}
 
 	for _, region := range targetRegions {
+		// The "global" region case is dealt outside this loop
```
zackproser, line 201:
```suggestion
		// The "global" region case is handled outside this loop
```

config/mocks/iam_users_all.yaml
hposca: Thanks for :eyes: that :smile: 
Will remove it now.
Remember kids: Copy & Paste is not the best approach for most things :stuck_out_tongue:

README.md
hposca: Thanks :smile:

aws/aws.go
hposca: Absolutely! 337 lines for a single function is too much :smile: 
And, as you said, maybe not on this PR.

aws/iam_types.go
hposca: > NIT: could we name this explitily as `users`

You mean the `IAMUsers` struct?
(And then change its usage to `users` everywhere in the code/config?)

hposca: > same applies to the function params above too!

Yeap! That makes a lot of sense... doing it now :smile:

config/config.go
hposca: Yeap, naming is hard :disappointed: 
Do you have any more suggestions? `NamesRegExp` is fine for me too, but I don't know if it is clear/explicit enough :thinking:

aws/iam.go
hposca: No, I wasn't thinking about any numbers specifically. Just because of the "mantra/practice/scars" of _"if something have pagination, it will use pagination"_ :stuck_out_tongue:

hposca:
> Looks good! raised_hands
> I've left a couple of comments/questions, but nothing that should block you hopefully! I've also noticed there's a couple of TODOs that are still around here (the exclusion time filter & listPages), so I imagine this is still WIP slightly_smiling_face

Thank you so much for your thorough review :bowing_man:

aws/aws.go
hposca: Thanks :smile:

config/mocks/iam_users_all.yaml
hposca: I removed it on the latest commit :smile:

aws/aws.go
```diff
@@ -478,6 +487,44 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
...
 
+	// Global Resources - These resources are global and do not belong to a specific region
+
+	logging.Logger.Infof("Checking region [%d/%d]: %s", count, totalRegions, GlobalRegion)
+
+	// As there is no actual region named global we have to pick a random one just to create the session
+	randomRegion, err := getRandomRegion()
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
```
yorinasub17, line 498: NIT: Use `us-east-1` instead of random, which is available in every account.

aws/aws.go
```diff
@@ -478,6 +487,44 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
...
+
+		if err != nil {
+			return nil, errors.WithStackTrace(err)
+		}
+		if len(userNames) > 0 {
+			iamUsers.UserNames = awsgo.StringValueSlice(userNames)
+			globalResources.Resources = append(globalResources.Resources, iamUsers)
+		}
+	}
+	// End IAM Users
```
yorinasub17, line 522: Somewhere in here we need a check to make sure the global region is in `targetRegions`. This is to handle the case where someone puts `global` in the excluded regions list.

aws/aws.go
```diff
@@ -519,43 +567,66 @@ func IsNukeable(resourceType string, resourceTypes []string) bool {
...
+		// region that will be used to create a session
+		sessionRegion := region
+		var err error
+
+		// As there is no actual region named global we have to pick a random one just to create the session
+		if region == GlobalRegion {
+			sessionRegion, err = getRandomRegion()
+			if err != nil {
+				return errors.WithStackTrace(err)
+			}
```
yorinasub17, line 615: NIT: Use `us-east-1`

aws/iam.go
```diff
@@ -0,0 +1,66 @@
...
+	output, err := svc.ListUsers(input)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	for _, user := range output.Users {
+		if config.ShouldInclude(*user.UserName, configObj.IAMUsers.IncludeRule.NamesRE, configObj.IAMUsers.ExcludeRule.NamesRE) {
```
yorinasub17, line 26: NIT: Use `aws.StringValue` when dereferencing string ptrs from AWS SDK (this is a safe dereference that returns `''` for `nil` instead of `panic`)


```suggestion
		if config.ShouldInclude(aws.StringValue(user.UserName), configObj.IAMUsers.IncludeRule.NamesRE, configObj.IAMUsers.ExcludeRule.NamesRE) {
```

aws/iam.go
```diff
@@ -0,0 +1,66 @@
...
+			userNames = append(userNames, user.UserName)
+		}
+	}
+
+	return userNames, nil
+}
+
+// TODO: This is only deleting the user but no associated resources
+// According to https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.DeleteUser
+// "you must delete the items attached to the user manually, or the deletion fails"
```
yorinasub17, line 36: Unfortunately, this needs to be handled before we can merge this in as it is going to be super common for users to have stuff associated with it.

We should replicate https://github.com/gruntwork-io/prototypes/blob/master/aws-gc/aws-gc.py#L74 here.

yorinasub17 (commented):
Looks close, but had a few things that need to be resolved prior to merge.

Also, somewhere in [the docs on regions](https://github.com/gruntwork-io/cloud-nuke/#nuke-resources-in-certain-regions) we should mention how the `global` region keyword is handled.

aws/iam_types.go
ina-stoyanova: Yep, that's exactly what I meant - apologies for not being clearer to start with!

config/config.go
ina-stoyanova: Agreed - finding the right name is difficult. I am happy to proceed with whatever makes sense to you most!

aws/iam.go
ina-stoyanova: Hahah, that makes sense :)

hposca:
@ina-stoyanova @zackproser @yorinasub17 mind taking another look, specially at the [last commit](https://github.com/gruntwork-io/cloud-nuke/pull/181/commits/01e5d6e8d7d9fe342fa92f2cd69d5deccbe57730)?
This last one is _"only a small feature"_:tm:  that actually requires a bunch of other stuff and extra validations, which I believe I was able to do.
Thanks :smile:

ina-stoyanova (approved):
I think it looks really good & easy to read now! 👏

aws/aws.go
```diff
@@ -519,43 +567,63 @@ func IsNukeable(resourceType string, resourceTypes []string) bool {
...
+
+	return nil
+}
+
 // NukeAllResources - Nukes all aws resources
 func NukeAllResources(account *AwsAccountResources, regions []string) error {
 	for _, region := range regions {
+		// region that will be used to create a session
+		sessionRegion := region
+		var err error
```
yorinasub17, line 608: NIT: this is not necessary, as it gets defined a few lines below with `session, err := ...`

aws/iam.go
```diff
@@ -0,0 +1,417 @@
...
+		if err != nil {
+			logging.Logger.Errorf("[Failed] %s", err)
+			multierror.Append(multiErr, err)
+			continue
+		}
+
+		// Delete the user
+		// NOTE: The actual user deletion should always be the last step.
+		// This way we can guarantee that it will fail if we forgot to delete/detach an item
+		err = deleteUser(svc, userName)
```
yorinasub17, line 405: NIT: This inner block should be a function. It would greatly simplify this logic, as you can return `err` to shortcircuit and only have to remember to add to multiErr in one place.

yorinasub17 (commented):
Updates LGTM! Had a few nits, but all of them can be addressed in a later PR.

One final thing: can you update our `circleci` config so it doesn't delete users when we run it as part of our regularly scheduled nuke jobs? We have real users we need to keep in our accounts and we need to sanitize those by updating our config.

You can update https://github.com/gruntwork-io/cloud-nuke/blob/master/.circleci/config.yml#L48 to include user to do this.

aws/iam.go
hposca: You mean moving all `detach*/delete*` functions to a new one that simply returns an `error` (as soon as there is one) so we can have something like this?
```go
if err := deleteAll(...); err != nil {
  logging.Logger.Errorf("[Failed] %s", err)
  multierror.Append(multiErr, err)
  continue
}
```

yorinasub17: Yes, although you won't need to do the `continue` at that point and you can use the if block immediately following this line I commented on.

hposca: By the way I just did the following and it worked as expected :smile: 
```go
func nukeUser(svc *iam.IAM, userName *string) error {
	functions := []func(svc *iam.IAM, userName *string) error{
		detachUserPolicies, // TODO: Add CLI option to delete the Policy as policies exist independently of the user
		deleteInlineUserPolicies,
		removeUserFromGroups, // TODO: Add CLI option to delete groups as groups exist independently of the user
		deleteLoginProfile,
		deleteAccessKeys,
		deleteSigningCertificate,
		deleteSSHPublicKeys,
		deleteServiceSpecificCredentials,
		deleteMFADevices,
		deleteUser, // NOTE: The actual user deletion should always be the last step.
		// This way we can guarantee that it will fail if we forgot to delete/detach an item
	}

	for _, fn := range functions {
		err := fn(svc, userName)
		if err != nil {
			return err
		}
	}

	return nil
}

// Delete all IAM Users
func nukeAllIamUsers(session *session.Session, userNames []*string) error {
	if len(userNames) == 0 {
		logging.Logger.Info("No IAM Users to nuke")
		return nil
	}

	logging.Logger.Info("Deleting all IAM Users")

	deletedUsers := 0
	svc := iam.New(session)
	multiErr := new(multierror.Error)

	for _, userName := range userNames {
		err := nukeUser(svc, userName)
		if err != nil {
			logging.Logger.Errorf("[Failed] %s", err)
			multierror.Append(multiErr, err)
		} else {
			deletedUsers++
			logging.Logger.Infof("Deleted IAM User: %s", *userName)
		}
	}

	logging.Logger.Infof("[OK] %d IAM User(s) terminated", deletedUsers)
	return multiErr.ErrorOrNil()
}
```

Was it something like this? :thinking:

aws/iam.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	return nil
+}
+
+func deleteLoginProfile(svc *iam.IAM, userName *string) error {
+	var lastError error
+	maxAttempts := 10
+
+	for maxAttempts > 0 {
```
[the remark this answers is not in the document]
hposca: I was able to use a similar logic but I believe it will be better for another PR :smile:

hposca:
@yorinasub17 mind taking another look into the new commits that implemented the NITs that you gave?
(The retry logic will be in another PR)

aws/iam.go
yorinasub17: Yup this is what I had in mind!

zackproser (approved):
LGTM! Nice work - :ship: it!

PR #187, opened 2021-04-01
Exclude transit gateway resources from nuking job

yorinasub17:
There appears to be a bug in the implementation of the transit gateway nuking job that causes our account nuking CI job to fail. See [this failing build](https://app.circleci.com/pipelines/github/gruntwork-io/cloud-nuke/3272/workflows/466051ad-35fb-47b4-a5e4-5ba89d1db316/jobs/24002) for example.

Note that it does not appear to be IAM permission related, as the transit gateway related IAM permissions are all nested under `ec2`, which our CI user has full access to.

This PR disables nuking those resources until we can find a resolution.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,10 @@ jobs:
               --older-than 1h \
               --force \
               --config ./.circleci/nuke_config.yml \
-              --exclude-resource-type iam
+              --exclude-resource-type iam \
+              --exclude-resource-type transit-gateway \
+              --exclude-resource-type transit-gateway-attachment \
+              --exclude-resource-type transit-gateway-route-table
           no_output_timeout: 1h
   nuke_sandbox:
     <<: *defaults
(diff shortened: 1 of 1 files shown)


bwhaley (approved):
Oh, that's unfortunate. At least we have a quick workaround.

yorinasub17:
Thanks for review! Going to merge this in so we can get nuking going again.

bwhaley:
Looks like it has to do with the availability of the service within the region:

```
$ aws --region us-west-2 ec2 describe-transit-gateway-vpc-attachments                                                                                                                                
{
    "TransitGatewayVpcAttachments": []
}
$ aws --region ap-northeast-3 ec2 describe-transit-gateway-vpc-attachments                                                                                                        

An error occurred (InvalidAction) when calling the DescribeTransitGatewayVpcAttachments operation: The action DescribeTransitGatewayVpcAttachments is not valid for this web service.
```

PR #194, opened 2021-06-24
Add support for nuking secrets manager secrets

yorinasub17:
This includes support for nuking Secrets Manager secrets, including usage of the config file to restrict what secrets should be nuked, and an update to our config file to ensure that only test secrets manager secrets are deleted in our environment.

NOTE: This was prompted from our internal investigation in to the AWS bill (notes available in notion).


hposca (approved):
:+1:

yorinasub17:
Thanks for review! I looked into the test failures, and they are unrelated to this change:

One is about Transit gateway and us hitting limits in our account.
The other one is about a version incompatibility when creating aurora-mysql.

I'll open a new PR to stabilize the build.

yorinasub17:
Thanks for review! Going to merge this in as the build is passing now.

PR #188, opened 2021-04-01
Check if Transit Gateway is available in the given region

bwhaley:
Should fix the issue described in #187.

PR #195, opened 2021-06-25
Attempt to stabilize build

yorinasub17:
- Remove exclusion for transit gateway. Not sure why we had some exclusions there for our nuke jobs, since we don't use transit gateway internally.
- Add logging to find the regions where the RDS cluster test fails (I had a few builds where it failed, but couldn't discern the region).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,10 +30,7 @@ jobs:
               --older-than 1h \
               --force \
               --config ./.circleci/nuke_config.yml \
-              --exclude-resource-type iam \
-              --exclude-resource-type transit-gateway \
-              --exclude-resource-type transit-gateway-attachment \
-              --exclude-resource-type transit-gateway-route-table
+              --exclude-resource-type iam
           no_output_timeout: 1h
   nuke_sandbox:
     <<: *defaults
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -30,10 +30,7 @@ jobs:
               --older-than 1h \
               --force \
               --config ./.circleci/nuke_config.yml \
-              --exclude-resource-type iam \
-              --exclude-resource-type transit-gateway \
-              --exclude-resource-type transit-gateway-attachment \
-              --exclude-resource-type transit-gateway-route-table
```
brikis98, line 36: I think we were excluding transit gateway because it wasn't enabled in some regions and cloud-nuke wasn't handling that correctly. I believe that was fixed in https://github.com/gruntwork-io/cloud-nuke/pull/188, so removing the exclusion is the right thing to do.

brikis98 (approved):
LGTM!

.circleci/config.yml
yorinasub17: Ahhh makes sense! Thanks for the clarification!

yorinasub17:

PR #197, opened 2021-06-25
Nuke NAT Gateways

yorinasub17:
This adds support for nuking NAT gateways. Additionally, this introduces a config file that is compatible with our environments (so we don't accidentally delete real NAT resources).

Note that this required bumping the AWS SDK for Go, which in turn required regenerating the EC2 mocks.


.circleci/nuke_config.yml
```diff
@@ -24,3 +24,13 @@ SecretsManager:
       - "^RDSDBConfig[a-zA-Z0-9]{6}$"
       - "ECRDeployRunnerTestSSHKey-[a-zA-Z0-9]{6}$"
       - "ECRDeployRunnerTestGitPAT-[a-zA-Z0-9]{6}$"
+
+NatGateway:
+  # There is a bug in the cloud-nuke config where exclude does not work by itself, so we add a null include rule that
```
brikis98, line 29: NIT: link to bug?

.circleci/nuke_config.yml
```diff
@@ -24,3 +24,13 @@ SecretsManager:
...
+  # includes everything.
+  include:
+    names_regex:
+      - ".*"
+  exclude:
+    names_regex:
+      - "^ecs-deploy-runner-v2-nat-gateway-0$"
```
brikis98, line 36: NIT: add comment on why these should be excluded?

README.md
```diff
@@ -232,45 +237,16 @@ Be careful when nuking and append the `--dry-run` option if you're unsure. Even
...
-##### secretsmanager resource type:
-_Note: the fields without `_regex` suffixes refer to support for plain-text matching against those fields._
-
-| field       | include | exclude |
-|-------------|---------|---------|
-| names       | none    | none    |
-| names_regex | ✅      | ✅      |
-| tags        | none    | none    |
-| tags_regex  | none    | none    |
+| resource type      | names | names_regex | tags | tags_regex |
```
brikis98, line 240: Nice improvement 👍

aws/nat_gateway.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+func getAllNatGateways(session *session.Session, excludeAfter time.Time, configObj config.Config) ([]*string, error) {
+	svc := ec2.New(session)
+
+	allNatGateways := []*string{}
+	input := &ec2.DescribeNatGatewaysInput{}
+	err := svc.DescribeNatGatewaysPages(
+		input,
+		func(page *ec2.DescribeNatGatewaysOutput, lastPage bool) bool {
```
brikis98, line 26: Oh, interesting, is this some new way to handle pagination in the new AWS Go SDK?

aws/nat_gateway.go
```diff
@@ -0,0 +1,156 @@
...
+	)
+	return allNatGateways, errors.WithStackTrace(err)
+}
+
+func shouldIncludeNatGateway(ngw *ec2.NatGateway, excludeAfter time.Time, configObj config.Config) bool {
+	if ngw == nil {
+		return false
+	}
+
+	if ngw.CreateTime != nil && excludeAfter.Before(*ngw.CreateTime) {
```
brikis98, line 43: What if `CreateTime` is `nil`? What does that mean?

aws/nat_gateway.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	// There is no bulk delete nat gateway API, so we delete the batch of nat gateways concurrently using go routines.
+	logging.Logger.Infof("Deleting Nat Gateways in region %s", region)
+	wg := new(sync.WaitGroup)
+	wg.Add(len(identifiers))
+	errChans := make([]chan error, len(identifiers))
+	for i, ngwID := range identifiers {
+		errChans[i] = make(chan error, 1)
+		go deleteNatGatewayAsync(wg, errChans[i], svc, ngwID)
```
brikis98, line 80: NIT: Does AWS throttle us if we do too many of these concurrently?

aws/nat_gateway.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return nil
+}
+
+// areAllNatGatewaysDeleted returns true if all the requested NAT gateways have been deleted. This is determined by
+// querying for the statuses of all the NAT gateways, and checking if AWS knows about them (if not, the NAT gateway was
+// deleted and rolled off AWS DB) or if the status was updated to deleted.
+func areAllNatGatewaysDeleted(svc *ec2.EC2, identifiers []*string) (bool, error) {
+	// NOTE: we don't need to do pagination here, because the pagination is handled by the caller to this function.
+	resp, err := svc.DescribeNatGateways(&ec2.DescribeNatGatewaysInput{NatGatewayIds: identifiers})
```
brikis98, line 128: NIT: is there a max number of NAT gateway IDs that can be passed in?

aws/nat_gateway.go
yorinasub17: I recall seeing this for a while, so I don't think it's new. However, it's not available in all API functions. It seems to be one of those things that are sometimes supported and othertimes not, depending on the team that manages the service.

brikis98: Roger

.circleci/nuke_config.yml
yorinasub17: Done: 996d7068a0b9da7e77c7d148e344fbd30ad38542

.circleci/nuke_config.yml
yorinasub17: Done: 996d7068a0b9da7e77c7d148e344fbd30ad38542

aws/nat_gateway.go
yorinasub17: I think that is an impossible condition (since it's an autopopulated field), but I added this guard to avoid a panic. If it is `nil`, this routine will consider it in the nuke routine, which seems like the right thing to do given the impossibility of the condition.

aws/nat_gateway.go
yorinasub17: The max is 1000, which I doubt would be hit so I think it's safe to leave this as is.

aws/nat_gateway.go
yorinasub17: Added a guard to be extra clear: 039adb4788bb302f4447c87ce05f6bdb56b1d650

aws/nat_gateway.go
brikis98: Roger

brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Test failure is an unrelated function that appears to be transient, so I'll go ahead and merge this in!

PR #209, opened 2021-09-28
Fix Config file not including when there is a 'exclude' rule

### LINKED ISSUE #198: Config file with only exclude rules do not work as expected
I don't remember if this was intentional or not.

If you create a config file where there is only an exclude rule, intuitively this should mean "include everything but what is requested to be excluded," but in practice this becomes "exclude everything" because `cloud-nuke` treats empty `include` rules as "include nothing."

You can work around this by adding an explicit `include` rule to include everything:

```yaml
s3:
  include:
    names_regex:
      - ".*"
  exclude:
    names_regex:
      - "^my-real-s3-bucket$"
```

marinalimeira:
Fixes #198.

Refactored the function `ShouldInclude` to behave properly when there is a `exclude` but no `include` config. I added a unit test to this function that catches the bug in question.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -79,27 +79,22 @@ func matches(name string, regexps []Expression) bool {
 	return false
 }
 
-// ShouldInclude - Checks if a name should be included according to the inclusion and exclusion rules
+// ShouldInclude - Checks if a resource's name should be included according to the inclusion and exclusion rules
 func ShouldInclude(name string, includeREs []Expression, excludeREs []Expression) bool {
-	shouldInclude := false
-
-	if len(includeREs) > 0 {
-		// If any include rules are specified,
-		// only check to see if an exclude rule matches when an include rule matches the user
-		if matches(name, includeREs) {
-			shouldInclude = true
-			if matches(name, excludeREs) {
-				shouldInclude = false
-			}
-		}
-	} else if len(excludeREs) > 0 {
-		// Only check to see if an exclude rule matches when there are no include rules defined
-		if matches(name, excludeREs) {
-			shouldInclude = false
-		}
+	// If no rules are defined, should always include
+	if len(includeREs) == 0 && len(excludeREs) == 0 {
+		return true
+		// If a rule that exclude matches, should not include
+	} else if matches(name, excludeREs) {
+		return false
+		// Given the 'name' is not in the 'exclude' list, should include if there is no 'include' list
+	} else if len(includeREs) == 0 {
+		return true
+		// Given there is a 'include' list, and 'name' is there, should include
+	} else if matches(name, includeREs) {
+		return true
+		// If it's not in the 'include' list, should not include
 	} else {
-		shouldInclude = true
+		return false
 	}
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -79,27 +79,22 @@ func matches(name string, regexps []Expression) bool {
...
-		}
-	} else if len(excludeREs) > 0 {
-		// Only check to see if an exclude rule matches when there are no include rules defined
-		if matches(name, excludeREs) {
-			shouldInclude = false
-		}
+	// If no rules are defined, should always include
+	if len(includeREs) == 0 && len(excludeREs) == 0 {
+		return true
+		// If a rule that exclude matches, should not include
```
brikis98, line 87: NIT: indentation on this comment and the ones below seems off...

config/config.go
```diff
@@ -79,27 +79,22 @@ func matches(name string, regexps []Expression) bool {
...
+		// Given the 'name' is not in the 'exclude' list, should include if there is no 'include' list
+	} else if len(includeREs) == 0 {
+		return true
+		// Given there is a 'include' list, and 'name' is there, should include
+	} else if matches(name, includeREs) {
+		return true
+		// If it's not in the 'include' list, should not include
 	} else {
-		shouldInclude = true
+		return false
```
brikis98, line 98: NIT: could be simplified to:

```go
} else {
  return matches(name, includeREs)
}

File: config/config_test.go [hunk 3]
```diff
@@ -331,3 +332,36 @@ func TestConfigSecretsManager_FilterNames(t *testing.T) {
...
+	exclude, err := regexp.Compile(`test.*`)
+	require.NoError(t, err)
+	excludeREs := []Expression { { RE: *exclude } }
+
+	assert.False(t, ShouldInclude("test-openvpn-123", includeREs, excludeREs),
+		"Should not include when matches from the 'exclude' list")
+	assert.True(t, ShouldInclude("tf-state-bucket", includeREs, excludeREs),
+		"Should include when doesn't matches from the 'exclude' list")
+}
+func TestShouldInclude_IncludeWhenMatches(t *testing.T) {
```
brikis98, line 356: Nice unit tests 👍 

NIT: consider adding one for the case where there are both include and exclude values specified.

brikis98 (approved):
Overall, LGTM! 👍

config/config_test.go
marinalimeira: Added another test at 43bcacbc35e

config/config.go
marinalimeira: Ah, nice catch! Fixed b0add83

config/config.go
marinalimeira: Fixed b0add83

marinalimeira:
Thank you for the review, Jim! Addressed the suggestions.

brikis98 (approved):
LGTM!

rhoboat (approved):
LGTM!! Thanks for fixing this.

PR #216, opened 2021-10-12
Fix regex formatting from nuke_config.yml

marinalimeira:
Build is current broken with the following error:
```
❯ houston exec phxops -- go run main.go aws \
  --older-than 1h \
  --force \
  --config ./.circleci/nuke_config.yml \
  --exclude-resource-type iam
ERROR: Error reading config - ./.circleci/nuke_config.yml - yaml: unmarshal errors:
  line 1: cannot unmarshal !!seq into string
exit status 1
ERROR: ErrorWithExitCode{ Err = exit status 1, ExitCode = 1 }
```

### DIFF
--- a/.circleci/nuke_config.yml
+++ b/.circleci/nuke_config.yml
@@ -14,8 +14,8 @@ s3:
       - "^kafka-zk-standalone-[a-zA-Z0-9]{6}$"
       - "^zookeeper-cluster-test-[a-zA-Z0-9]{6}$"
       - "^terragrunt-test-bucket-[a-zA-Z0-9]{6}.*"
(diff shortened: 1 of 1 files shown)


infraredgirl:
Why was this not caught in the original PR? Do we need to add a regression test?

yorinasub17:
We don't have a unit test that sanity checks our own internal circleci nuke config. I think adding a regression test that confirms it's parseable makes sense!

infraredgirl:
Great, I added an issue for this ⬆️.

marinalimeira:
Thank you for creating the issue, Ana!

PR #218, opened 2021-10-12
Pull request 169

ina-stoyanova:
Running tests for #169


infraredgirl:
Why not trigger test run on the original PR?

ina-stoyanova:
And also I believe there's actually the limiter that we've got a PR that's been raised from someone as an external contributor, so the tests won't be triggered automatically for that PR.

Issue #226, opened 2021-10-29
Could we simplify making AWS API calls with paging?


ina-stoyanova:
**Context:** 
In quite a few of our `aws` functions we have to make a batched call into AWS API to get some data (E.g. `s3` module & `dynamodb`, and others). 

**Question:** Is it possible to extract this logic somehow? E.g. so not every module has to figure out how to paging and batch calls on its own?

hongil0316:
For any future resources we implement, we can leverage the pages get calls to handle pagination, instead of having our own batching logic. A lot of the resources already use this API for that. 

Closing this issue since this is no longer relevant - may not make sense to make this change to all existing resources.

PR #233, opened 2021-11-12
Nuke cloudwatch dashboard

### LINKED ISSUE #232: Feature: Add ability to nuke CloudWatch Dashboards
We have a bunch of leftover cloudwatch dashboards in phxdevops, which accumulate to a good amount of money, so we should add the ability to nuke them through cloud-nuke.

yorinasub17:
Fixes #232

### DIFF
--- a/aws/cloudwatch_dashboard.go
+++ b/aws/cloudwatch_dashboard.go
@@ -1,0 +1,89 @@
+package aws
+
+import (
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/gruntwork-io/go-commons/errors"
+
+	"github.com/gruntwork-io/cloud-nuke/config"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+)
+
+func getAllCloudWatchDashboards(session *session.Session, excludeAfter time.Time, configObj config.Config) ([]*string, error) {
+	svc := cloudwatch.New(session)
+
+	allDashboards := []*string{}
+	input := &cloudwatch.ListDashboardsInput{}
+	err := svc.ListDashboardsPages(
+		input,
+		func(page *cloudwatch.ListDashboardsOutput, lastPage bool) bool {
+			for _, dashboard := range page.DashboardEntries {
+				if shouldIncludeCloudWatchDashboard(dashboard, excludeAfter, configObj) {
+					allDashboards = append(allDashboards, dashboard.DashboardName)
+				}
+			}
+			return !lastPage
+		},
+	)
+	return allDashboards, errors.WithStackTrace(err)
+}
+
+func shouldIncludeCloudWatchDashboard(dashboard *cloudwatch.DashboardEntry, excludeAfter time.Time, configObj config.Config) bool {
+	if dashboard == nil {
+		return false
+	}
+
+	if dashboard.LastModified != nil && excludeAfter.Before(*dashboard.LastModified) {
+		return false
+	}
+
+	return config.ShouldInclude(
+		aws.StringValue(dashboard.DashboardName),
+		configObj.CloudWatchDashboard.IncludeRule.NamesRegExp,
+		configObj.CloudWatchDashboard.ExcludeRule.NamesRegExp,
+	)
+}
+
+func nukeAllCloudWatchDashboards(session *session.Session, identifiers []*string) error {
+	region := aws.StringValue(session.Config.Region)
+
+	svc := cloudwatch.New(session)
+
+	if len(identifiers) == 0 {
+		logging.Logger.Infof("No CloudWatch Dashboards to nuke in region %s", region)
+		return nil
+	}
+
+	// NOTE: we don't need to do pagination here, because the pagination is handled by the caller to this function,
+	// based on CloudWatchDashboard.MaxBatchSize, however we add a guard here to warn users when the batching fails and has a
+	// chance of throttling AWS. Since we concurrently make one call for each identifier, we pick 100 for the limit here
+	// because many APIs in AWS have a limit of 100 requests per second.
+	if len(identifiers) > 100 {
+		logging.Logger.Errorf("Nuking too many CloudWatch Dashboards at once (100): halting to avoid hitting AWS API rate limiting")
+		return TooManyCloudWatchDashboardsErr{}
+	}
+
+	logging.Logger.Infof("Deleting CloudWatch Dashboards in region %s", region)
+	input := cloudwatch.DeleteDashboardsInput{DashboardNames: identifiers}
+	_, err := svc.DeleteDashboards(&input)
+	if err != nil {
+		logging.Logger.Errorf("[Failed] %s", err)
+		return errors.WithStackTrace(err)
+	}
+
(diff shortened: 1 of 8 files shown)


config/config_test.go
```diff
@@ -370,16 +371,16 @@ func TestShouldInclude_IncludeWhenMatches(t *testing.T) {
 func TestShouldInclude_WhenMatchesIncludeAndExclude(t *testing.T) {
 	include, err := regexp.Compile(`test.*`)
 	require.NoError(t, err)
-	includeREs := []Expression { { RE: *include } }
+	includeREs := []Expression{{RE: *include}}
 
 	exclude, err := regexp.Compile(`.*openvpn.*`)
 	require.NoError(t, err)
-	excludeREs := []Expression { { RE: *exclude } }
+	excludeREs := []Expression{{RE: *exclude}}
```
ina-stoyanova, line 378: NIT: just noting if this was intentional - I don't think it's a massive problem, but some different formatting.

config/config.go
```diff
@@ -84,13 +85,13 @@ func ShouldInclude(name string, includeREs []Expression, excludeREs []Expression
 	// If no rules are defined, should always include
 	if len(includeREs) == 0 && len(excludeREs) == 0 {
 		return true
-	// If a rule that exclude matches, should not include
+		// If a rule that exclude matches, should not include
 	} else if matches(name, excludeREs) {
 		return false
-	// Given the 'name' is not in the 'exclude' list, should include if there is no 'include' list
+		// Given the 'name' is not in the 'exclude' list, should include if there is no 'include' list
```
ina-stoyanova, line 91: NIT: same here - seems to have added some extra spacing before the comments

aws/cloudwatch_dashboard_types.go
```diff
@@ -0,0 +1,35 @@
...
+// ResourceIdentifiers - The instance ids of the ec2 instances
+func (cwdb CloudWatchDashboards) ResourceIdentifiers() []string {
+	return cwdb.DashboardNames
+}
+
+func (cwdb CloudWatchDashboards) MaxBatchSize() int {
+	return 100
+}
+
+// Nuke - nuke 'em all!!!
```
ina-stoyanova, line 28: 😁

aws/cloudwatch_dashboard.go
```diff
@@ -0,0 +1,89 @@
...
+	region := aws.StringValue(session.Config.Region)
+
+	svc := cloudwatch.New(session)
+
+	if len(identifiers) == 0 {
+		logging.Logger.Infof("No CloudWatch Dashboards to nuke in region %s", region)
+		return nil
+	}
+
+	// NOTE: we don't need to do pagination here, because the pagination is handled by the caller to this function,
```
ina-stoyanova, line 60: just to confirm - is that a pattern we should adopt more or less in all cases for pagination?

aws/cloudwatch_dashboard_test.go
```diff
@@ -0,0 +1,164 @@
...
+	require.NoError(t, err)
+
+	session, err := session.NewSession(&aws.Config{Region: aws.String(region)})
+	require.NoError(t, err)
+	svc := cloudwatch.New(session)
+
+	cwdbName := createCloudWatchDashboard(t, svc, region)
+	defer deleteCloudWatchDashboard(t, svc, cwdbName, true)
+
+	// Assert CWDB is picked up without filters
```
ina-stoyanova, line 49: NIT: could we maybe swap this comment abbreviation out for the full name? Just to keep it easier to read

aws/cloudwatch_dashboard_test.go
```diff
@@ -0,0 +1,164 @@
...
+	region, err := getRandomRegion()
+	require.NoError(t, err)
+
+	session, err := session.NewSession(&aws.Config{Region: aws.String(region)})
+	require.NoError(t, err)
+	svc := cloudwatch.New(session)
+
+	cwdbNames := []*string{}
+	for i := 0; i < 3; i++ {
+		// We ignore errors in the delete call here, because it is intended to be a stop gap in case there is a bug in nuke.
```
ina-stoyanova, line 97: Q: for my knowledge - so this stop gap is meant to try & delete the dashboard once, but if that call fails, then it leaves it at that in case the deletion failed for a good reason?

ina-stoyanova (dismissed):
Overall looks good! I haven't had a chance to give it a test, but I assume you'll try running it against PHX anyway very soon. 

I've left some questions - those are there for me to be able to build better knowledge and mental model of cloud-nuke in general, and then some NITs - mostly around formatting, so feel free to ignore too.

config/config.go
yorinasub17: Ah yea this is not good. Again, I think this was caused by my upgrade to go 1.17. It looks like this is now the new standard formatting. I moved the comments around to make it more readable: f1cf9600d599cd669a482881bc62178416d43928

aws/cloudwatch_dashboard.go
yorinasub17: Yes I think we should. I've been doing this on any new resource I introduce, but haven't been going back and retroactively porting it.

ina-stoyanova: Hm, that's awesome! I've raised an issue actually some weeks ago about this: https://github.com/gruntwork-io/cloud-nuke/issues/226 so really good to see this!

aws/cloudwatch_dashboard_test.go
yorinasub17: Ah no. So the logic here is that the test makes a call to `nukeAllCloudWatchDashboards`, which will go ahead and delete the dashboards that it created in the test, if it works. However, if there is a bug in the function and it fails, those dashboards will be lying around. To avoid having the test pollute the environment with those dashboards, we attempt an alternative cleaning function here to try to make sure those get deleted.

The reason we can't error check is that in the happy path, the dashboards would be deleted by the nuke function and they won't exist, thus this function will throw an error.

aws/cloudwatch_dashboard_test.go
yorinasub17: Done! cdaafce633cc5c357544350150269852ec2b583d

yorinasub17:
> I haven't had a chance to give it a test, but I assume you'll try running it against PHX anyway very soon.

Ah I already ran it. I wrote this because I found a bunch of dashboards across multiple regions lying around in PHX and wanted to easily delete them, while also having a more semi-permanent way to periodically delete them.

config/config.go
ina-stoyanova: perfect! thanks for addressing it

aws/cloudwatch_dashboard_test.go
ina-stoyanova: ahhh! that makes total sense! thanks, Yori :)

ina-stoyanova:
>Ah I already ran it. I wrote this because I found a bunch of dashboards across multiple regions lying around in PHX and wanted to easily delete them, while also having a more semi-permanent way to periodically delete them.

 OK, that's great! 
 
 All looks good now, so approving 👍

ina-stoyanova (approved):
Looks good! 

(Caveat: the failing test doesn't seem to be linked to this new feature, so all good there, we can handle it separately)

yorinasub17:
Thanks for thorough review! Merging and releasing now!

Issue #69, opened 2019-09-11
Feature Request - Removing Config Rules


marklehmann26:
Just wondering if you are looking to in the future add support for removing all config rules.

Use Case would be if anyone has been playing with config rules and setup and forgot any rules they have configured and are being triggered will cause additional charges to users accounts.

It would be good if the tool could also clear these out so if someone wanted to start their account from scratch they can.

Thoughts anyone?

brikis98:
Are you referring to AWS Config? If so, that sounds like a reasonable addition. PR welcome!

marklehmann26:
Hi, @brikis98 yeah AWS config, apologies I forgot to add this to the title.

TryTryAgain:
FYI, would need to also clean up the delivery-channel as well: https://docs.aws.amazon.com/config/latest/developerguide/manage-delivery-channel.html "The AWS Config console does not provide the option to delete the delivery channel, so you must use the AWS CLI, the AWS Config API, or one of the AWS SDKs."

gitsstewart:
Closing this issue out given the fact we have implemented it already. See https://github.com/gruntwork-io/cloud-nuke/blob/75137b08f0b7d565026557ffa3f8259c59d7fcff/aws/resources/config_service.go#L67

PR #240, opened 2021-11-16
Support AWS Config

ekristen:
This implements nuking AWS Config Rules, but excludes any generated internally by AWS that can't be removed. These generally are prefixed with `aws-service-rule`

- Resolves https://github.com/gruntwork-io/cloud-nuke/issues/69

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -642,6 +642,38 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		}
 		// End Dynamo DB tables
 
+		// ConfigService Rules
+		configServiceRules := ConfigService{}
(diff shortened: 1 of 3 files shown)


marinalimeira:
Thank you for your contribution, Erik!

The files are missing a `config_test.go`.  You can check out the `aws/*_test.go` for examples. We create these tests where we (1) create the resource, (2) delete it and (3) confirm that the resource is not there anymore.

zackproser:
Thanks again for your contribution! I wanted to let you know that we've now got support for AWS config recorders and rules in cloud-nuke as of https://github.com/gruntwork-io/cloud-nuke/pull/405

Therefore, I'm going to close out this PR.

PR #241, opened 2021-11-16
Support SecurityHub

ekristen:
This implements SecurityHub for cloud-nuke. This has a helper function that makes an API call to determine if SecurityHub is enabled first in each region and skips if it is disabled so that errors do not occur.

### DIFF
--- a/aws/securityhub.go
+++ b/aws/securityhub.go
@@ -1,0 +1,97 @@
+package aws
+
+import (
+	"strings"
+
+	awsgo "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/securityhub"
+	"github.com/gruntwork-io/cloud-nuke/logging"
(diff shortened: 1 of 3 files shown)


ellisonc:
@ekristen We're getting caught up on old PRs.  If this is still a feature you need, could you resolve the merge conflicts, add telemetry as exists in other modules, add any relevant tests, and update the readme?

aws/securityhub.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	logging.Logger.Info("Deleting all Security Hub Insights")
+
+	deletedinsights := 0
+	svc := securityhub.New(session)
+	multiErr := new(multierror.Error)
+
+	for _, name := range insights {
+		if *name == "disable-securityhub" {
```
zackproser, line 54: Ah for consistency's sake - we should use the aws SDK method for dereferencing a pointer.

hongil0316:
Closing this PR request since we haven't received any response for the past 2 months and the PR has been stale for almost 2 years. Let's re-open this if we need this in the future.

PR #242, opened 2021-11-16
support nuking ec2 keypairs

ekristen:
This adds support for nuking ec2 keypairs


denis256:
Interesting implementation, however, noticed a couple of issues:
 * there are no tests to know in future that removal will continue to work;
 * filtering of keys seems to not work, tried to filter keys with `"key-0fff.*"` but it included all keys;

ellisonc:
This feature has been implemented by another PR

PR #243, opened 2021-11-16
support deleting vpcs that are non-default

ekristen:
This takes the existing VPC code and implements it for non-default per region nuking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ The currently supported functionality includes:
 - Deleting all SQS queues in an AWS account
 - Deleting all S3 buckets in an AWS account - except for buckets tagged with Key=cloud-nuke-excluded Value=true
 - Deleting all default VPCs in an AWS account
-- Deleting all non-default VPCs in an AWS Account - except for default vps which is handled by default-aws
+- Deleting VPCs in an AWS Account (except for default VPCs which is handled by the dedicated `defaults-aws` subcommand)
 - Deleting all IAM users in an AWS account
 - Deleting all Secrets Manager Secrets in an AWS account
 - Deleting all NAT Gateways in an AWS account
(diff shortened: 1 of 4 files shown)


bwhaley (commented):
Changes LGTM!

One request: could you please add VPCs to the list of resources in the README, and also add a subsection under Usage that explains that that default VPCs are excluded when nuking VPCs? I agree with you that excluding them makes sense since we have a dedicated `defaults-aws` command for that. Also, nuking the default VPCs would break some internal use cases we have. But we ought to explain our reasoning in the docs.

Thanks!

ekristen:
@bwhaley changes made as requested.

README.md
```diff
@@ -26,6 +26,7 @@ The currently supported functionality includes:
 - Deleting all SQS queues in an AWS account
 - Deleting all S3 buckets in an AWS account - except for buckets tagged with Key=cloud-nuke-excluded Value=true
 - Deleting all default VPCs in an AWS account
+- Deleting all non-default VPCs in an AWS Account - except for default vps which is handled by default-aws
```
bwhaley, line 29:
```suggestion
- Deleting VPCs in an AWS Account (except for default VPCs which is handled by the dedicated `defaults-aws` subcommand)
```

bwhaley (commented):
Thanks. One fix and I think this is ready to ship.

bwhaley (approved):
LGTM!

bwhaley:
https://github.com/gruntwork-io/cloud-nuke/releases/tag/v0.7.1

ekristen:
Thanks @bwhaley I've got a number of other PRs open ;-) -- I'll make sure to update the README on each one too.

PR #247, opened 2021-11-17
Update explanation of bumping the minor in docs.

rhoboat:
Filed https://github.com/gruntwork-io/cloud-nuke/issues/256 for the follow-up work described below:

----

We need a PR template that includes a checkbox such as:

- [ ] Does this PR introduce a new nuked resource type? Maintainer: Please release this as a minor version update.
or
- [ ] This PR introduces a new nuked resource type and must be released as a minor version update.


CONTRIBUTING.md
```diff
@@ -70,3 +78,14 @@ to include the following:
 
 The maintainers for this repo will review your code and provide feedback. If everything looks good, they will merge the
 code and release a new version, which you'll be able to find in the [releases page](../../releases).
+
+### Bump the minor
```
yorinasub17, line 82: Ah we actually already have this in the README (https://github.com/gruntwork-io/cloud-nuke#choosing-a-new-release-tag), so we should reuse that and link there to avoid duplication.

yorinasub17 (commented):
Makes sense to have this in `CONTRIBUTING` as well! Just a minor note that we already have the bulk of the content in the `README`, so we should be linking there.

ina-stoyanova:
The PR template is a great idea here @rhoboat. Did you plan for that to be part of this PR too?

CONTRIBUTING.md
rhoboat: Thanks for finding that! Will make the change.

zackproser (approved):
LGTM :+1:

rhoboat:
Thanks for helping me push this through, Zack! Merging now.

PR #248, opened 2021-11-17
fix a bug where deleted nat gateways were included in nuke list

ekristen:
This fixes a bug where if the nat gateway was already deleted and in the deleted state, subsequent runs would still list it as needing to be deleted, however thankfully the AWS API would not error on subsequent runs, but it would appear that resources were not being cleaned up.


yorinasub17:
Build failed, but for unrelated reason, so we can bring this in. Thanks for your contribution!

yorinasub17 (approved):
LGTM

Issue #149, opened 2020-11-01
Cloud formation templates within the scope of the nuke should be accountable in AWS 


andrewpsp:
Since you can't destroy an EKS cluster with nodegroups attached, the nuking fails for that resource. This is a request to also  target the deletion of cloudformation templates.

hongil0316:
Closing this issue since we had no activities for this ticket for the past 3 years.

Issue #95, opened 2020-02-27
Support deletion of cloudformation stacks


andreamaruccia:
Would be nice to be able to delete cloudformation stacks older than x days

brikis98:
Seems like a reasonable feature request, but we don't make much use of CloudFormation here, so we're not likely to get to this any time soon. PRs welcome.

ekristen:
@brikis98 I'm looking at submitting a PR to solve this. There are stack and stack sets and sets can contain stacks, would you all rather there be a type like `cf-stackset` and `cf-stack` and just ensure we delete all sets before stacks or would you all rather it be implemented like the VPC nuke code where we just know X resources have to be nuked before the VPC can finally be deleted? 

It's possible to delete stacks without stacksets only if the sets don't refer to the stacks being deleted.

hongil0316:
Closing this PR as we've not made any changes or updates since 2021.

PR #249, opened 2021-11-17
Support Nuking CloudFormation StackSets and Stacks

ekristen:
This implements the removal of StackSets and Stacks with wait groups to ensure that the status moves to DELETED before considering the item nuked.

**Note:** there is a chance that when targeting stacks only that if it's part of a StackSet it could trigger an error which will be logged.

- [ ] Update Documentation
- [ ] Update Tests
- [x] Update Code

These are issues that this PR potentially resolves.

- Potentially Resolves https://github.com/gruntwork-io/cloud-nuke/issues/95
- Potentially Resolves https://github.com/gruntwork-io/cloud-nuke/issues/149

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -221,6 +221,36 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		// The order in which resources are nuked is important
 		// because of dependencies between resources
 
+		// Cloudformation StackSets -- This is early on because it can control other infrastructure
+		stacksets := CloudformationStackSets{}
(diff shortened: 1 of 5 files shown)


ekristen:
I'm working on docs and tests. Also found an issue when stacks depend on each other if not deleted in the correct way the current implementation won't work, working on a fix.

ekristen:
It looks like to delete properly each stack's output has to be queried for import use, then a dependency graph built and deleted in order.

ellisonc:
@ekristen thanks for your contribution, is this still something you plan on working on? We're in the process of cleaning old PRs and will be closing this in the coming weeks if not.

Issue #227, opened 2021-11-03
Fix tests failing 


ina-stoyanova:
Currently a few tests on master are [failing](https://app.circleci.com/pipelines/github/gruntwork-io/cloud-nuke/5287/workflows/61334a04-9d9b-48ac-bbf2-f68116936fa3/jobs/26128):
- [x] TestDeleteFullIamUser (fixed by #255)
- [x] TestListSecretsManagerSecrets (fixed by https://github.com/gruntwork-io/cloud-nuke/pull/259)
- [x] TestTimeFilterExclusionNewlyCreatedSecret (fixed by https://github.com/gruntwork-io/cloud-nuke/pull/259)
- [ ] TestCanNukeAllEcsClustersOlderThan24Hours

denis256:
Test `TestCanNukeAllEcsClustersOlderThan24Hours` seems to pass now

![image](https://user-images.githubusercontent.com/10694338/148781277-b8fd421d-4aff-4534-acf4-233658d85ac1.png)

![image](https://user-images.githubusercontent.com/10694338/148781217-43ec19f5-da6e-4e57-8723-9fbcb2e422ed.png)

brikis98:
Hm, I'm still seeing a lot of flaky tests. Consider the test history on [this branch](https://app.circleci.com/pipelines/github/gruntwork-io/cloud-nuke?branch=bugfix%2Fissue-262&filter=all). I re-ran the tests a bunch of times and got different failures each time:

* `TestDeleteFullIamUser`
* `TestListEIPAddress`
* `TestNukeTransitGatewayVpcAttachment`
* `TestListSecretsManagerSecrets`
* `TestCanListAllOpenSearchDomainsOlderThan24hours`

So I think there's more work to be done to make these tests resilient.

hongil0316:
All the tests have been fixed and refactored. Closing this issue.

PR #255, opened 2021-11-29
#227 Fix TestDeleteFullIamUser

denis256:
Attempt to fix TestDeleteFullIamUser by adding retries in the assignment of MFA device

Execution result:
image

Fixes:
#227

### DIFF
--- a/aws/iam_test.go
+++ b/aws/iam_test.go
@@ -379,7 +379,7 @@ func createFullTestUser(t *testing.T, session *session.Session) (*userInfos, err
 			UserName:              aws.String(userName),
 		})
 		if lastError != nil {
-			time.Sleep(1 * time.Second)
+			time.Sleep(5 * time.Second)
 			continue
 		}
 		break
--- a/aws/iam_test.go
+++ b/aws/iam_test.go
@@ -376,20 +376,26 @@ func createFullTestUser(t *testing.T, session *session.Session) (*userInfos, err
 	})
 	require.NoError(t, err)
 	virtualMFADevice := output.VirtualMFADevice
-	// We now need to generate two codes to enable the MFA device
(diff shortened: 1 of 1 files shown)


infraredgirl:
So this solution essentially retries enabling the MFA device 3 times, in hopes that it will eventually succeed. This could certainly improve the test, but I wonder if there's a better way to address the failure. 

Looking at the error in the original test failure: 
```
EntityTemporarilyUnmodifiable: MFA cannot be enabled while the Login Profile  for principal cloud-nuke-test-full-user-dsiSR3 is being created.
```
So it seems that the problem is a race condition between creating the Login Profile and enabling the MFA. Is there a way we can check/poll the results of the `CreateLoginProfile` call and ensure that the call to `EnableMFADevice` only happens one the Login Profile has been created?

denis256:
Since `CreateLoginProfile` returns same completed structure each time, updated approach to detect in-progress profile creation by sending update requests and check errors

https://docs.aws.amazon.com/sdk-for-go/api/service/iam/

infraredgirl:
Hmm, unfortunately the test failed with `'Verify if profile is ready' unsuccessful after 10 retries`. So, while adding retries does not guarantee to solve the issue with this test, I still think it's an improvement to the current state. I think we should go ahead and merge, but before doing so, maybe give it one last try of tweaking the timeout a bit. Right now it's 10 retries with 5 seconds in between, for a total of 50 seconds. Maybe try something like 4 retires with 30 seconds in between, a total of 2 minutes. What do you think @denis256?

denis256:
Updated timing configurations

infraredgirl (approved):
Thanks, Denis, looks good! Approving and merging.

PR #265, opened 2021-12-17
Config file for lambda and ebs

yorinasub17:
Add Config File support for Lambda functions and EBS Volumes for name filtering.

Note that this does not include the unit tests for the config file for the new file, like we've been doing with the other resources. After some thought, I don't think they add anything particularly useful beyond what we currently have, since they only check if we added the attribute + `yaml` tag to the field, which ends up being a tautology (that is, we are only matching the string between the mock and what's in the code).


denis256 (dismissed):

denis256 (approved):

yorinasub17:
Thanks for review! Going to merge this in now that the build is passing.

PR #267, opened 2022-01-05
Add ability to nuke Elasticache

### LINKED ISSUE #236: Feature: Add ability to nuke Elasticache
We have a bunch of leftover Elasticache clusters in phxdevops, which accumulate to a good amount of money, so we should add the ability to nuke them through cloud-nuke.

tonerdo:
Fixes #236 

Open question: do we want to add support for Elasticache to the config file?

### DIFF
--- a/aws/elasticache_test.go
+++ b/aws/elasticache_test.go
@@ -26,6 +26,14 @@ func createTestElasticacheCluster(t *testing.T, session *session.Session, name s
 	if err != nil {
 		assert.Failf(t, "Could not create test Elasticache cluster", errors.WithStackTrace(err).Error())
 	}
+
+	err = svc.WaitUntilCacheClusterAvailable(&elasticache.DescribeCacheClustersInput{
+		CacheClusterId: awsgo.String(name),
+	})
+
+	if err != nil {
+		assert.Failf(t, "Error waiting for cluster to become available", errors.WithStackTrace(err).Error())
+	}
 }
 
 func TestListElasticacheClusters(t *testing.T) {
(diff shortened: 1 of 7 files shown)


aws/elasticache_test.go
```diff
@@ -0,0 +1,99 @@
...
+		CacheClusterId: awsgo.String(name),
+		CacheNodeType:  awsgo.String("cache.t3.micro"),
+		Engine:         awsgo.String("memcached"),
+		NumCacheNodes:  awsgo.Int64(1),
+	}
+
+	_, err := svc.CreateCacheCluster(&param)
+	if err != nil {
+		assert.Failf(t, "Could not create test Elasticache cluster", errors.WithStackTrace(err).Error())
+	}
```
yorinasub17, line 29: NIT: We've been slowly transitioning error assertions to use `require.NoError(t, err)`, which simplifies the code while providing useful error logs when the assertion fails, and also halting the test if there is an error in the step.

yorinasub17 (commented):
LGTM! Only had minor nits.

Would also be great if you can add in config file support! We've been adding it in for all new resources, and slowly retroactively adding support for existing resources as well.

aws/elasticache_test.go
tonerdo: Thanks! I've switched over to using it

tonerdo:
> Would also be great if you can add in config file support!

Config file support added

yorinasub17 (commented):
Updates LGTM! The only issue I see is that you will have a compile error for the `config_test.go` file when it tries to create the empty config for testing: https://github.com/gruntwork-io/cloud-nuke/blob/master/config/config_test.go#L12

Once that's addressed, I can kick off the regression build for testing!

tonerdo:
All done @yorinasub17. Qq: do we no longer run the CI tests for every push?

yorinasub17:
Just kicked off the build!

> Qq: do we no longer run the CI tests for every push?

We do, but we heightened the security ACLs on CircleCI such that only certain github users can trigger builds that access the secrets. I think we are still in discussion about whether contractors should have those permissions or not. Until we get that sorted out, we will have to have one of the FTEs trigger the builds.

tonerdo:
There's a failure in `TestListElasticacheClustersWithConfigFile`, will look into it tomorrow

tonerdo:
Hey @yorinasub17, I've pushed up a fix for the failing test. Please help run the CI again when you get the chance

denis256 (approved):
Adding approval to test CI job

yorinasub17:
Just kicked it off! https://app.circleci.com/pipelines/github/gruntwork-io/cloud-nuke/5933/workflows/0ebb1166-5457-456c-a3b4-3f2ac8fd46a5

PR #279, opened 2022-02-01
Avoid nuking KMS customer keys in our env

yorinasub17:


denis256 (approved):

yorinasub17:
Thanks for review! Merging now to avoid KMS nukes in our test accounts.

PR #287, opened 2022-03-10
Add config file support for ElasticIP, AutoScalingGroup, LaunchConfiguration, EC2

brandonstrohmeyer:
Adds `names_regex` config file support for:
- ElasticIP
- AutoScalingGroup
- LaunchConfiguration
- EC2 Instances


aws/eip.go
```diff
@@ -70,15 +71,35 @@ func getAllEIPAddresses(session *session.Session, region string, excludeAfter ti
...
+		return false
+	}
+
+	if excludeAfter.Before(firstSeenTime) {
+		return false
+	}
+
+	// If Name is unset, GetEC2ResourceNameTagValue returns error and zero value string
+	// Ignore this error and pass empty string to config.ShouldInclude
+	allocationName, _ := GetEC2ResourceNameTagValue(address.Tags)
```
tonerdo, line 94: We should be using the allocation Ids and not the name

tonerdo (changes requested):
Looks mostly good. I requested changes to how to decide whether or not to include EIP addresses

brandonstrohmeyer:
Thanks for the review,

It should be noted that the EC2 Instance filter also looks at instance name and not instance ID. Additionally, the NAT Gateway filter does as well and is a past PR submission of mine (!266), it was not challenged therefore I didn't think it was an issue.

@tonerdo Is this a topic that is up for discussion? In my experience with Cloud Nuke there is _significant_ utility in being able to use a regex pattern to match a resource name and not just specify individual resource IDs. Especially in an IaC controlled environment where resources may be spun up and down, changing the underlying instance ID but retaining the same Name value. 

If this is a hard rule not up for discussion then I will make the requested changes and also open an issue to request README updates that clarify the "Names" pattern filter does not actually filter the resource names but instead filters the AWS Resource ID as it is not clear.

tonerdo:
@brandonstrohmeyer you're right. I looked through the codebase and founds other examples. This PR is good to go

tonerdo (approved):

Issue #213, opened 2021-10-01
[AWS] Fails to delete s3 buckets that have bucket policies on


avineshwar:

infraredgirl:
Thanks for the bug report! This should definitely be fixed, but I'm not sure when we'll have time to address it. If you would like to submit a PR with a fix, that would be greatly appreciated!

denis256:
Fix included in https://github.com/gruntwork-io/cloud-nuke/releases/tag/v0.11.6

PR #294, opened 2022-04-11
Delete s3 BucketPolicy before nuke Bucket to avoid errors when policy… 

maxhaensel:
… refuses deletion

This fixes a bug where you run into the problem when deleting s3 buckets but they have a bucket policy that prohibits deleting the entire bucket. 

The fix ensures that the bucket policy is cleared before attempting to delete the bucket.

Solves the Issue #213

### DIFF
--- a/aws/s3_test.go
+++ b/aws/s3_test.go
@@ -651,7 +651,7 @@ func TestFilterS3Bucket_Config(t *testing.T) {
 
 // TestNukeS3BucketWithBucketPolicy tests deletion of S3 buckets with a policy that denies deletion
 func TestNukeS3BucketWithBucketPolicy(t *testing.T) {
-	awsParams, err := newS3TestAWSParams("eu-central-1")
+	awsParams, err := newS3TestAWSParams("")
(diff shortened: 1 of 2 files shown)


maxhaensel:
https://gist.github.com/maxhaensel/43a9ad7b395a1cdd8e2fd555491d5882

aws/s3_test.go
```diff
@@ -321,7 +322,7 @@ type TestNukeS3BucketArgs struct {
 
 // testNukeS3Bucket - generates the test function for TestNukeS3Bucket
 func testNukeS3Bucket(t *testing.T, args TestNukeS3BucketArgs) {
-	awsParams, err := newS3TestAWSParams("")
+	awsParams, err := newS3TestAWSParams("eu-central-1")
```
denis256, line 325: I was wondering why is required to use specific region

maxhaensel: @denis256 my organization has a SCP in place so i have only access to Resources in eu-central-1. This should not be a problem for the test with S3 buckets. If it is absolutely necessary, I can repeat the test again in a private account.

Thanks

PR #295, opened 2022-04-24
Add config support for EKS and exclude ghost cluster in our account

yorinasub17:
In our account where we are running `cloud-nuke`, we have a ghost EKS cluster that shows up in the list API, but will return a 404 if you try to get the details. This EKS cluster will presumably be garbage collected by AWS eventually, but in the meantime, it causes problems by halting execution of `cloud-nuke` everytime it tries to nuke our account.

This PR works around this limitation by adding config support to EKS cluster nuking, and adding an explicit exclude rule to ignore the ghost cluster.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	EC2                   ResourceType `yaml:"EC2"`
 	CloudWatchLogGroup    ResourceType `yaml:"CloudWatchLogGroup"`
 	KMSCustomerKeys       ResourceType `yaml:"KMSCustomerKeys"`
+	EKSCluster            ResourceType `yaml:"EKSCluster"`
 }
 
 type ResourceType struct {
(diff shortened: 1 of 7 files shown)


rhoboat (approved):
Okay to ignore the 👻 EKS cluster. LGTM!

config/config.go
```diff
@@ -32,6 +32,7 @@ type Config struct {
 	EC2                   ResourceType `yaml:"EC2"`
 	CloudWatchLogGroup    ResourceType `yaml:"CloudWatchLogGroup"`
 	KMSCustomerKeys       ResourceType `yaml:"KMSCustomerKeys"`
+	EKSCluster            ResourceType `yaml:"EKSCluster"`
```
rhoboat, line 35: Ah! Great!

yorinasub17:

PR #309, opened 2022-06-08
Attach nuke permissions by circleci context, not env var

yorinasub17:
## Description

This updates the nuke circleci rules to use the CircleCI context instead of using env vars.


denis256 (approved):

yorinasub17:

PR #319, opened 2022-06-23
Update codeowners

rhoboat:
Adds the IaC US team engineers to CODEOWNERS and removes others. Addresses https://github.com/gruntwork-io/cloud-chasers/issues/27. We should add back the maintenance tier owners as well.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-@yorinasub17 @zackproser @rhoboat
+* @yorinasub17 @zackproser @rhoboat @denis256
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1 +1 @@
-* @marinalimeira @infraredgirl @robmorgan @ina-stoyanova
+@yorinasub17 @zackproser @rhoboat
```
rhoboat, line 1:
```suggestion
* @yorinasub17 @zackproser @rhoboat @denis256
```

yorinasub17 (approved):

PR #323, opened 2022-07-06
Implement support for Macie member accounts

### LINKED ISSUE #322: Feaure: Add support for Macie member accounts
Implement support for running cloud-nuke against Macie member accounts, thereby deleting the associations and disabling Macie in the given region. 

Note that we may want to add support for cleaning up resources from the Organizations membership side of things separately. Member accounts seem to be able to disassociate from their admin accounts and disable Macie in their own regions whenever they wish.

zackproser:
## Description

Fixes #322 by implementing support for "nuking" Macie accounts that were created by Invitation only. (Macie member accounts created via AWS Organizations do not have sufficient permissions to disassociate _themselves_ from the Org). 

Programmatic testing with Macie is pretty awkward. I found that you can continuously accept an invite from another account to become a member account, then disassociate yourself (nuke), then re-use the same standing invitation to re-associate yourself as a member account. I'm currently using this workaround to leave a "standing" invite from our phxdevops account in `us-east-1` for sanity. 

We could expand our test coverage out to other regions, but that would just require maintaining a map of regions to standing invitation IDs.

Furthermore, Macie member accounts don't expose anything useful to us in terms of filtering resources by their creation time, so the usual semantics of `excludeAfter` won't work here. As a result, I just include all found member Macie accounts as nuke-able. 

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### Migration Guide

Implement support for inspecting and nuking Macie member accounts that were created via Invitation only. (Macie member accounts created via AWS Organizations are not able to disassociate themselves).

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -734,7 +734,7 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		// Macie member accounts
 		macieAccounts := MacieMember{}
 		if IsNukeable(macieAccounts.ResourceName(), resourceTypes) {
-			accountIds, err := getAllMacieAccounts(session, excludeAfter, configObj)
+			accountIds, err := getAllMacieMemberAccounts(session, excludeAfter, configObj)
 			if err != nil {
 				return nil, errors.WithStackTrace(err)
 			}
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -732,7 +732,7 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 		}
 		// End GuardDuty detectors
 		// Macie member accounts
-		macieAccounts := Macie{}
+		macieAccounts := MacieMember{}
 		if IsNukeable(macieAccounts.ResourceName(), resourceTypes) {
 			accountIds, err := getAllMacieAccounts(session, excludeAfter, configObj)
 			if err != nil {
@@ -843,7 +843,7 @@ func ListResourceTypes() []string {
 		KmsCustomerKeys{}.ResourceName(),
 		CloudWatchLogGroups{}.ResourceName(),
 		GuardDuty{}.ResourceName(),
-		Macie{}.ResourceName(),
+		MacieMember{}.ResourceName(),
 	}
 	sort.Strings(resourceTypes)
 	return resourceTypes
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -731,6 +731,20 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 
 		}
 		// End GuardDuty detectors
+		// Macie member accounts
+		macieAccounts := Macie{}
+		if IsNukeable(macieAccounts.ResourceName(), resourceTypes) {
+			accountIds, err := getAllMacieAccounts(session, excludeAfter, configObj)
+			if err != nil {
+				return nil, errors.WithStackTrace(err)
+			}
+			if len(accountIds) > 0 {
+				macieAccounts.AccountIds = accountIds
+				resourcesInRegion.Resources = append(resourcesInRegion.Resources, macieAccounts)
+			}
+
+		}
+		// End Macie member accounts
 
 		if len(resourcesInRegion.Resources) > 0 {
 			account.Resources[region] = resourcesInRegion
@@ -829,6 +843,7 @@ func ListResourceTypes() []string {
 		KmsCustomerKeys{}.ResourceName(),
 		CloudWatchLogGroups{}.ResourceName(),
 		GuardDuty{}.ResourceName(),
+		Macie{}.ResourceName(),
 	}
 	sort.Strings(resourceTypes)
 	return resourceTypes
--- a/aws/macie.go
+++ b/aws/macie.go
@@ -60,8 +60,6 @@ func nukeAllMacieMemberAccounts(session *session.Session, identifiers []string) 
 	svc := macie2.New(session)
 	region := aws.StringValue(session.Config.Region)
 
-	disassociatedAccountIds := []string{}
-
 	if len(identifiers) == 0 {
 		logging.Logger.Infof("No Macie member accounts to nuke in region %s", *session.Config.Region)
 		return nil
@@ -82,7 +80,6 @@ func nukeAllMacieMemberAccounts(session *session.Session, identifiers []string) 
 		}
 
 		logging.Logger.Infof("[OK] Macie account association for accountId %s deleted in %s", accountId, region)
-		disassociatedAccountIds = append(disassociatedAccountIds, accountId)
 	}
 
 	return nil
--- a/aws/macie.go
+++ b/aws/macie.go
@@ -19,23 +19,7 @@ func getAllMacieMemberAccounts(session *session.Session, excludeAfter time.Time,
 
 	allMacieAccounts := []string{}
 	output, err := svc.GetAdministratorAccount(&macie2.GetAdministratorAccountInput{})
-	// If the current account does have an Administrator account relationship, and it is enabled, then we consider this a macie member account
-	if output.Administrator != nil && output.Administrator.RelationshipStatus != nil {
-		if aws.StringValue(output.Administrator.RelationshipStatus) == macie2.RelationshipStatusEnabled {
-
-			input := &sts.GetCallerIdentityInput{}
-			output, err := stssvc.GetCallerIdentity(input)
-			if err != nil {
-				return allMacieAccounts, errors.WithStackTrace(err)
-			}
-
-			currentAccountId := aws.StringValue(output.Account)
-
-			allMacieAccounts = append(allMacieAccounts, currentAccountId)
-		}
-	}
 	if err != nil {
-
 		// There are several different errors that AWS may return when you attempt to call Macie operations on an account
 		// that doesn't yet have Macie enabled. For our purposes, this is fine, as we're only looking for those accounts and
 		// regions where Macie is enabled. Therefore, we ignore only these expected errors, and return any other errror that might occur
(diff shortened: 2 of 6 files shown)


aws/macie.go
```diff
@@ -0,0 +1,87 @@
...
+		var rnfe *macie2.ResourceNotFoundException
+
+		switch {
+		case goerror.As(err, &ade):
+			logging.Logger.Debugf("Macie AccessDeniedException means macie is not enabled in account, so skipping")
+		case goerror.As(err, &rnfe):
+			logging.Logger.Debugf("Macie ResourceNotFoundException means macie is not enabled in account, so skipping")
+		default:
+			return allMacieAccounts, errors.WithStackTrace(err)
+		}
```
yorinasub17, line 52: This condition check should happen first, and should return `nil, nil` at the end. The reason is because we don't want to touch `output` at all if there is an error since the structure of the object is undefined, so to prevent any potential bugs, we should always check the error first.

aws/macie.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/macie2"
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/gruntwork-io/cloud-nuke/config"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/go-commons/errors"
+)
+
+func getAllMacieAccounts(session *session.Session, excludeAfter time.Time, configObj config.Config) ([]string, error) {
```
yorinasub17, line 16: NIT: this should be named `getAllMacieMeberAccounts`.

aws/macie.go
```diff
@@ -0,0 +1,87 @@
...
+		case goerror.As(err, &rnfe):
+			logging.Logger.Debugf("Macie ResourceNotFoundException means macie is not enabled in account, so skipping")
+		default:
+			return allMacieAccounts, errors.WithStackTrace(err)
+		}
+	}
+	return allMacieAccounts, nil
+}
+
+func nukeAllMacieAccounts(session *session.Session, identifiers []string) error {
```
yorinasub17, line 57: NIT: This should be named `nukeAllMacieMemberAccounts`.

aws/macie.go
```diff
@@ -0,0 +1,87 @@
...
+	region := aws.StringValue(session.Config.Region)
+
+	disassociatedAccountIds := []string{}
+
+	if len(identifiers) == 0 {
+		logging.Logger.Infof("No Macie member accounts to nuke in region %s", *session.Config.Region)
+		return nil
+	}
+
+	logging.Logger.Infof("Deleting Macie account membership in %s", region)
```
yorinasub17, line 68: NIT:
```suggestion
	logging.Logger.Infof("Deleting Macie account membership and disabling Macie in %s", region)
```

aws/macie_types.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/gruntwork-io/go-commons/errors"
+)
+
+type Macie struct {
+	AccountIds []string
+}
+
+func (r Macie) ResourceName() string {
+	return "macie"
```
yorinasub17, line 13: This should be `macie-member` to differentiate with `macie-admin`.

yorinasub17 (commented):
Mostly LGTM! Had a few nits on terminology, and a potential bug in the if logic.

aws/aws.go
```diff
@@ -731,6 +731,20 @@ func GetAllResources(targetRegions []string, excludeAfter time.Time, resourceTyp
 
 		}
 		// End GuardDuty detectors
+		// Macie member accounts
+		macieAccounts := Macie{}
```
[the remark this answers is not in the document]
zackproser: Makes sense - done in 63926ff

aws/macie.go
zackproser: Done in 61e86b1

aws/macie.go
zackproser: Done in 61e86b1

aws/macie_types.go
zackproser: Done in 2dfdc00

aws/macie_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	retrievedAccountIds, lookupErr := getAllMacieMemberAccounts(session, time.Now(), config.Config{})
+	require.NoError(t, lookupErr)
+
+	assert.Contains(t, retrievedAccountIds, accountId)
+}
+
+// Macie is not very conducive to programmatic testing. In order to make this test work, we  maintain a standing invite
+// from our phxdevops test account to our nuclear-wasteland account. We can continuously "nuke" our membership because
+// Macie supports a member account *that was invited* to remove its own association at any time. Meanwhile, diassociating
```
rhoboat, line 37: Yeah, I read!

```suggestion
// Macie supports a member account *that was invited* to remove its own association at any time. Meanwhile, disassociating
```

rhoboat (dismissed):
LGTM!

aws/macie_test.go
zackproser: Nice catch, thanks!

zackproser:
Thanks for reviews @yorinasub17 and @rhoboat - looks like we're actually running cloud-nuke against phxdevops, so the Macie test is now passing. 

Current failure is a single known flaky test, so I think this is ready for another look.

aws/macie.go
```diff
@@ -0,0 +1,87 @@
...
+		var rnfe *macie2.ResourceNotFoundException
+
+		switch {
+		case goerror.As(err, &ade):
+			logging.Logger.Debugf("Macie AccessDeniedException means macie is not enabled in account, so skipping")
+		case goerror.As(err, &rnfe):
+			logging.Logger.Debugf("Macie ResourceNotFoundException means macie is not enabled in account, so skipping")
+		default:
+			return allMacieAccounts, errors.WithStackTrace(err)
+		}
```
yorinasub17, line 36: I think this is missing a return at the end for the two cases where we want to skip.

aws/macie_test.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Macie is not very conducive to programmatic testing. In order to make this test work, we  maintain a standing invite
+// from our phxdevops test account to our nuclear-wasteland account. We can continuously "nuke" our membership because
+// Macie supports a member account *that was invited* to remove its own association at any time. Meanwhile, disassociating
+// in this manner does not destroy or invalidate the original invitation, which allows us to to continually re-accept it
+// from our nuclear-wasteland account (where cloud-nuke tests are run), just so that we can nuke it again
+//
+// Macie is also regional, so for the purposes of cost-savings and lower admin overhead, we're initially only testing this
+// in the one hardcoded region
```
yorinasub17, line 42: Does this have risk of breaking our macie test in the other repos? E.g., in CIS service catalog: https://github.com/gruntwork-io/terraform-aws-cis-service-catalog/blob/master/test/security/macie_test.go

yorinasub17 (commented):
Mostly LGTM! Found two potential bugs.

aws/macie.go
```diff
@@ -0,0 +1,87 @@
...
+			return errors.WithStackTrace(disassociateErr)
+		}
+
+		_, err := svc.DisableMacie(&macie2.DisableMacieInput{})
+		if err != nil {
+			return errors.WithStackTrace(err)
+		}
+
+		logging.Logger.Infof("[OK] Macie account association for accountId %s deleted in %s", accountId, region)
+		disassociatedAccountIds = append(disassociatedAccountIds, accountId)
```
denis256, line 83: Values appended to `disassociatedAccountIds` aren't used, most probably should be returned

aws/macie.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/macie2"
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/gruntwork-io/cloud-nuke/config"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/go-commons/errors"
+)
+
+func getAllMacieMemberAccounts(session *session.Session, excludeAfter time.Time, configObj config.Config) ([]string, error) {
```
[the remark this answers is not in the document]
zackproser: Yeah - unfortunately we can't really implement this pattern with Macie member accounts. Since account membership is determined by the invitation acceptance status, and since the API doesn't expose when an account last accepted an invite, we don't have much to go off.

aws/macie_test.go
zackproser: As far as I can tell, that test is using a very similar pattern: 
- The Sandbox account is the administrator account, which invites phxdevops
- Phxdevops accepts the invite, to become a member account of Sandbox 

However, Macie is regional, and the CIS service catalog test is using regions in the EU, whereas this cloud-nuke Macie test is currently hardcoded to us-east-1. 

I think that will prevent conflict - so long as neither test has its testing regions expanded.

aws/macie.go
zackproser: Fixed in faf92ec

aws/macie_test.go
zackproser: Added more context in 7438eea

zackproser:
Looks like the current test failure is `TestListElasticacheClustersWithConfigFile` which is marked as Flaky.

aws/macie.go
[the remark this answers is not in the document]
yorinasub17: Perhaps remove it from the arg list and add a comment indicating that we can't support it?

yorinasub17 (approved):
Mostly LGTM! Just had one nit regarding the extraneous args.

zackproser:
Thanks for review! I'll merge this PR now and open a follow-up to make the improvement you suggested.

PR #327, opened 2022-07-12
Fix circleci pipeline to use scheduled pipelines

yorinasub17:
## Description

It turns out that we haven't been running the nuking routines due to the CircleCI triggers changing it's permissions model to no longer support scheduled workflows. Instead, we need to migrate to using [scheduled pipelines](https://circleci.com/docs/scheduled-pipelines). This PR updates the CCI config to start supporting scheduled pipelines.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

(no functional changes)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   docker:
     - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.16-tf1.0-tg31.1-pck1.7
-version: 2
+version: 2.1
 jobs:
   test:
     <<: *defaults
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -73,6 +73,10 @@ jobs:
 workflows:
   version: 2
```
rhoboat, line 74: Does this have to be `2`? What happens if it's `2.1`, or left off?

rhoboat (approved):
Good catch! Left a question, but LGTM. We won't know if the config works until the schedule tries to run.

.circleci/config.yml
yorinasub17: AFAIK, `2` is the latest version for workflows, and is the default (note that this is different from the global circleci version, which is at the top)

yorinasub17:
> We won't know if the config works until the schedule tries to run.

Yup unfortunately, but I think we can iterate on it quickly once Jim creates the initial workflow.

yorinasub17:
Ok triggers are created now, so going to merge in!

PR #335, opened 2022-07-21
Support nuking elasticache replication groups

### LINKED ISSUE #333: Error deleting elasticache within replication group
```
[cloud-nuke] INFO[2022-07-21T13:27:28-04:00] Terminating 1 resources in batches
[cloud-nuke] INFO[2022-07-21T13:27:28-04:00] Deleting 1 Elasticache clusters in region us-east-1
[cloud-nuke] ERRO[2022-07-21T13:27:28-04:00] [Failed] InvalidCacheClusterState: Cannot delete the cache cluster since it is the only member of a replication group. Delete the replication group instead.
        status code: 400, request id: 4846d313-aa5c-40d7-95f0-f7d868407904
[cloud-nuke] INFO[2022-07-21T13:27:28-04:00] [OK] 0 Elasticache clusters deleted in us-east-1
```

Looks like we may need to issue a call to [DeleteReplicationGroup](https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_DeleteReplicationGroup.html) if the elasticache node exists within a replication group.

zackproser:
## Description

Fixes #333 

These changes introduce support for nuking elasticache clusters that are members of a replication group. The single nuke method has been extended to handle both cases: 
1. Cache cluster is not a member of a replication group, and so can be deleted directly 
2. Cache cluster is a member of a replication group, which itself must then be deleted, because attempting to delete the member cache cluster itself will fail as long as the replication group exists. 

In testing against a deployed Ref Arch, these changes resolve the issue observed in #333.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### DIFF
--- a/aws/elasticache.go
+++ b/aws/elasticache.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/elasticache"
 	"github.com/gruntwork-io/cloud-nuke/config"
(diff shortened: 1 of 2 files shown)


aws/elasticache.go
```diff
@@ -57,11 +120,20 @@ func nukeAllElasticacheClusters(session *session.Session, clusterIds []*string)
...
+		cacheCluster, describeErr := getSingleCacheCluster(svc, clusterId)
+		if describeErr != nil {
+			return describeErr
+		}
+
+		var err error
+		if cacheCluster.ReplicationGroupId == nil {
+			err = nukeNonReplicationGroupElasticacheCluster(svc, clusterId)
+		} else {
+			err = nukeReplicationGroupMemberElasticacheCluster(svc, cacheCluster)
```
yorinasub17, line 134: It looks like this will delete the replication group, but not the cluster. I think it needs to call `nukeNonReplicationGroupElasticacheCluster` after the replication group is deleted, right?

yorinasub17 (commented):
Mostly LGTM. Just had one question which may be identifying a bug.

aws/elasticache.go
zackproser: That's what I thought as well, and I initially had it wired up that way, but was somewhat surprised that this lead to an error about the target cluster not being found. 

It looks like deleting the replication group actually deletes the member node as well.

yorinasub17: Ah gotcha! I think the comments need to be updated then, because the wording of the comment on line 124 ("because members of a replication group require that the replication group be deleted *first*", emphasis mine) made me think that the cache cluster needs to be deleted as well.

zackproser: Roger - done in [cc586f7](https://github.com/gruntwork-io/cloud-nuke/pull/335/commits/cc586f7e13b28067e1b8fe2a2c90f5ac112b520a)

yorinasub17 (dismissed):
LGTM, pending build!

yorinasub17 (dismissed):

zackproser:
@yorinasub17 Thanks for reviews! Looks like the tests are finally passing, so this one is ready for another look when you have a moment.

aws/elasticache.go
```diff
@@ -70,21 +172,16 @@ func nukeAllElasticacheClusters(session *session.Session, clusterIds []*string)
...
-		logging.Logger.Infof("Confirming deletion of %d Elasticache clusters in region %s", len(deletedClusterIds), *session.Config.Region)
+	logging.Logger.Infof("[OK] %d Elasticache clusters deleted in %s", len(deletedClusterIds), *session.Config.Region)
+	return nil
+}
 
-		for _, clusterId := range deletedClusterIds {
-			params := elasticache.DescribeCacheClustersInput{
-				CacheClusterId: clusterId,
-			}
+// Custome errors
```
yorinasub17, line 179:
```suggestion
// Custom errors
```

yorinasub17 (approved):
LGTM!

aws/elasticache.go
zackproser: Ah thanks! I'll open a subsequent PR for the typo.

zackproser:
Thanks for reviews! Going to merge this in now.

aws/elasticache.go
zackproser: Done here: https://github.com/gruntwork-io/cloud-nuke/pull/336

PR #348, opened 2022-08-08
Fix elasticache cluster nuking

### LINKED ISSUE #341: Issue tearing down Elasticache in CIS-Benchmark Ref Arch
As observed with @pete0emerson when nuking a CIS-Benchmark compliant Ref Arch:

`[cloud-nuke] ERRO[2022-08-01T10:38:11-07:00] [Failed] InvalidCacheClusterState: Cannot delete the cache cluster since it is the only member of a replication group. Delete the replication group instead.`

zackproser:
## Description

Fixes #341.

- Look up replication groups separately. Previously we were trying to look up all cache nodes via `DescribeCacheClusters`, which _will_ return nodes that are members of a ReplicationGroup - but won't signify which need to be passed through the ReplicationGroup code path. Instead, we first look up replication groups via their dedicated method `DescribeReplicationGroups` then issue a call to `DescribeCacheClusters`, specifying that the call should only return cache nodes that are not members of a replication group. 
- When then flatten both cache cluster types together, returning both types from the list method
- When nuking an elasticache cluster - return immediately if the cluster is found to be of type `Replication` - previously, incorrect fallthrough behavior was causing replication groups to end up in the code path for deleting single cache nodes.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/aws/elasticache.go
+++ b/aws/elasticache.go
@@ -16,13 +16,31 @@ import (
 // Returns a formatted string of Elasticache cluster Ids
 func getAllElasticacheClusters(session *session.Session, region string, excludeAfter time.Time, configObj config.Config) ([]*string, error) {
 	svc := elasticache.New(session)
-	result, err := svc.DescribeCacheClusters(&elasticache.DescribeCacheClustersInput{})
-	if err != nil {
-		return nil, errors.WithStackTrace(err)
+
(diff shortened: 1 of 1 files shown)


yorinasub17 (approved):
LGTM! Also confirmed test failures are unrelated.

zackproser:

PR #415, opened 2023-02-24
CORE-381/Add deletion of IAM Service Linked Roles

pete0emerson:
This code allows for IAM Service Linked Roles to be deleted.

In the Gruntwork Reference Architecture, we need to clean up AWSServiceRoleForAutoScaling.

Note that this is a naive approach; it requests the resource be deleted, but doesn't verify that they are actually deleted.

We could do this verification by capturing the DeletionTaskId and then querying for the deletion status until the status is (hopefully) SUCCEEDED. Alternate statuses are NOT_STARTED, IN_PROGRESS, and FAILED.

Thoughts here appreciated.

## Description

Fixes CORE-481

## TODOs

- [ ] Update the docs.

## Release Notes

Added ability to nuke [IAM Service Linked Roles](https://docs.aws.amazon.com/IAM/latest/UserGuide/using-service-linked-roles.html)

### DIFF
--- a/aws/iam_service_linked_role.go
+++ b/aws/iam_service_linked_role.go
@@ -1,0 +1,160 @@
+package aws
+
+import (
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/gruntwork-io/cloud-nuke/config"
+	"github.com/gruntwork-io/cloud-nuke/logging"
+	"github.com/gruntwork-io/cloud-nuke/report"
+	"github.com/gruntwork-io/go-commons/errors"
+	"github.com/hashicorp/go-multierror"
+)
+
+// List all IAM Roles in the AWS account
+func getAllIamServiceLinkedRoles(session *session.Session, excludeAfter time.Time, configObj config.Config) ([]*string, error) {
+	svc := iam.New(session)
+
+	allIAMServiceLinkedRoles := []*string{}
+	err := svc.ListRolesPages(
+		&iam.ListRolesInput{},
+		func(page *iam.ListRolesOutput, lastPage bool) bool {
+			for _, iamServiceLinkedRole := range page.Roles {
+				if shouldIncludeIAMServiceLinkedRole(iamServiceLinkedRole, excludeAfter, configObj) {
+					allIAMServiceLinkedRoles = append(allIAMServiceLinkedRoles, iamServiceLinkedRole.RoleName)
+				}
+			}
+			return !lastPage
+		},
+	)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+	return allIAMServiceLinkedRoles, nil
+}
+
+func deleteIamServiceLinkedRole(svc *iam.IAM, roleName *string) error {
+	_, err := svc.DeleteServiceLinkedRole(&iam.DeleteServiceLinkedRoleInput{
+		RoleName: roleName,
+	})
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+	logging.Logger.Infof("Successfully Deleted IAM ServiceLinked Role %s", aws.StringValue(roleName))
+	return nil
+}
+
+// Delete all IAM Roles
+func nukeAllIamServiceLinkedRoles(session *session.Session, roleNames []*string) error {
+	region := aws.StringValue(session.Config.Region)
+	svc := iam.New(session)
+
+	if len(roleNames) == 0 {
+		logging.Logger.Debug("No IAM Service Linked Roles to nuke")
+		return nil
+	}
+
+	// NOTE: we don't need to do pagination here, because the pagination is handled by the caller to this function,
+	// based on IAMRoles.MaxBatchSize, however we add a guard here to warn users when the batching fails and has a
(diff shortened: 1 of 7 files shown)


pete0emerson:
Failed tests are unrelated (Macie, EKS Cluster).

pete0emerson:
> do you have a sense for how long it takes for deletion once the request is submitted?

Yep, it's quite fast. I don't think I've gotten a immediate request for status in faster than it could actually do the deletion. So, we're talking seconds, so far in my experience.

aws/iam_service_linked_role.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	return nil
+}
+
+func shouldIncludeIAMServiceLinkedRole(iamServiceLinkedRole *iam.Role, excludeAfter time.Time, configObj config.Config) bool {
+	if iamServiceLinkedRole == nil {
+		return false
+	}
+
+	if strings.Contains(aws.StringValue(iamServiceLinkedRole.RoleName), "OrganizationAccountAccessRole") {
```
zackproser, line 140: Should we add a note to the README to explain which exclusions / special exc eptions we make when deleting IAM roles and why? It looks like it is possible to delete this role, for example, so in the case where someone is trying to clean up accounts that have this role present, this implicit behavior would probably be pretty confusing to them.

pete0emerson: Fixed in ce254a6

Ah, that was left over from my copy from the IAM Role. I've removed it and then added a comment to the iam_role.go file.

aws/iam_service_linked_role.go
```diff
@@ -0,0 +1,191 @@
...
+			allErrs = multierror.Append(allErrs, err)
+			logging.Logger.Debugf("[Failed] %s", err)
+		}
+	}
+	finalErr := allErrs.ErrorOrNil()
+	if finalErr != nil {
+		return gruntworkerrors.WithStackTrace(finalErr)
+	}
+
+	for _, roleName := range roleNames {
```
MoonMoon1919, line 129: Out of curiosity - do we need this second loop through `roleNames` or do you think the log line in `deleteIamServiceLinkedRole` is sufficient (currently line 82)? They both say the same think, but this log line includes the `[OK]` prefix and region.

aws/iam_service_linked_role_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	name := "cloud-nuke-test-" + util.UniqueID()
+	awsServiceName := "autoscaling.amazonaws.com"
+	iamServiceLinkedRoleName := "AWSServiceRoleForAutoScaling_" + name
+
+	err = createTestServiceLinkedRole(t, session, name, awsServiceName)
+	require.NoError(t, err)
+
+	err = nukeAllIamServiceLinkedRoles(session, []*string{&iamServiceLinkedRoleName})
+	require.NoError(t, err)
```
MoonMoon1919, line 95: Should we add an assertion here that the role is not in the result of `getAllIamServiceLinkedRoles`? I think that validates that the `nuke` was performed to our expectations, beyond not throwing an error. WDYT?

MoonMoon1919 (commented):
Add too small non-blocking comments about logging and a test. Otherwise looks great!

aws/iam_service_linked_role.go
pete0emerson: Fixed in ec4c03c

Nah, it's duplicate. I've removed line 82 to make it consistent with what is going on in `iam_role.go`.

aws/iam_service_linked_role_test.go
pete0emerson: Just to make sure I understand, are you suggesting creating a role that should not get deleted, and verify that it does indeed, not get deleted?

pete0emerson: Oh, I think I see what you're saying. Re-run getAllIamServiceLinkedRoles and make sure that the list doesn't contain the role any more.

MoonMoon1919: exactly!

Something like...

```go
roleNames, err = getAllIamServiceLinkedRoles(session, time.Now(), config.Config{})
require.NoError(t, err)

//AWSServiceRoleForAutoScaling_cloud-nuke-test
assert.NotContains(t, awsgo.StringValueSlice(roleNames), iamServiceLinkedRoleName)
```

pete0emerson: Addressed in 8572f99

MoonMoon1919 (approved):
LGTM

pete0emerson:
All of my tests are passing.

MoonMoon1919:
> All of my tests are passing.

:shipit:

zackproser:
I noticed the TODOs on the PR description were left unchecked - there's one that we're particularly interested in: 

> Attention Grunts - if this PR adds support for a new resource, ensure the nuke_sandbox and nuke_phxdevops jobs in .circleci/config.yml have been updated with appropriate exclusions (either directly in the job or via the .circleci/nuke_config.yml file) to prevent nuking IAM roles, groups, resources, etc that are important for the test accounts.

Was this step intentionally skipped for this PR? It's hard to tell because none of the TODOs are marked as done. 

If it's intentional, is that because we're confident there are no service linked roles in our phxdevops or sandbox accounts whose out of band deletion will interfere with any of our other tests, or some other reason?

pete0emerson:
Ugh. No, not intentionally skipped. I just blew it. I'll put up a PR to address.

pete0emerson:
I put up a PR in https://github.com/gruntwork-io/cloud-nuke/pull/418

PR #440, opened 2023-04-10
Bug: Filter out AMIs created by AWS Backup - CORE-664

arsci:
## Description

Closes [CORE-664](https://gruntwork.atlassian.net/browse/CORE-664)

Updated AMI filtering to exclude any images created by AWS Backup by checking for the reserved `aws:backup:source-resource` tag

Test run against `phxdevops` excluded the two images created by AWS Backup in `us-west-2`

## TODOs

- [x] Update the docs.

## Release Notes

Updated AMI filtering to exclude images created by AWS Backup

[CORE-664]: https://gruntwork.atlassian.net/browse/CORE-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aws/ami.go
+++ b/aws/ami.go
@@ -1,9 +1,10 @@
 package aws
 
 import (
+	"time"
+
 	"github.com/gruntwork-io/cloud-nuke/telemetry"
 	commonTelemetry "github.com/gruntwork-io/go-commons/telemetry"
-	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	awsgo "github.com/aws/aws-sdk-go/aws"
@@ -20,6 +21,14 @@ func getAllAMIs(session *session.Session, region string, excludeAfter time.Time)
 
 	params := &ec2.DescribeImagesInput{
 		Owners: []*string{awsgo.String("self")},
+		// Filters: []*ec2.Filter{
+		// 	{
+		// 		Name: awsgo.String("tag-key"),
+		// 		Values: []*string{
+		// 			awsgo.String("!aws2:backup:source-resource"),
+		// 		},
+		// 	},
+		// },
 	}
 
 	output, err := svc.DescribeImages(params)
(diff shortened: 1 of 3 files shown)


aws/ami.go
```diff
@@ -35,14 +36,30 @@ func getAllAMIs(session *session.Session, region string, excludeAfter time.Time)
...
 		}
 	}
 
 	return imageIds, nil
 }
 
+// Check if the image has an AWS Backup tag
+// Resources created by AWS Backup are listed as owned by self, but are actually
+// AWS managed resources and cannot be deleted here.
+func ImageHasAWSBackupTag(tags []*ec2.Tag) bool {
```
MoonMoon1919, line 50: I think this might be applicable beyond just AMIs, do we have a generic place we can put this so it's easy to find for other consumers?

zackproser: I suppose we could home it in go-commons and load it from there?

MoonMoon1919: I think it makes sense in this package specifically, i'm just wondering if we should put it in `helpers.go` or something like that?

arsci:
> Code changes LGTM! Is there anywhere in the README or code that we should document this functionality? Would anyone potentially be confused by the AMIs being excluded?

Good call, I will add some documentation that describes this!

aws/ami.go
MoonMoon1919: @arsci  I know that the EBS snapshots that back the AMIs get a similar tag.

I'd imagine the behavior would be similar for the AWS services listed here: https://docs.aws.amazon.com/aws-backup/latest/devguide/whatisbackup.html#supported-resources Looking at that list I think I'd probably be most immediately concerned by RDS - e.g., if someone was using backup to manage RDS snapshots -> tried to run cloud nuke they might experience the same problem we're solving for here.

That being said, I think for now this is probably fine and we can circle up internally on how to verify the behavior for other resources.

Issue #494, opened 2023-07-10
New Test Pattern for Unit Testing in Cloud-Nuke


hongil0316:
It seems like we are creating various AWS resources on cloud for unit testing. This approach works but it incurs additional cost and we see quite a lot of transient issues (e.g., QuotaLimit, Timeout, ...). We might want to try switching to this recommendation where we use mock implementation to test the behaviour we want - https://aws.amazon.com/blogs/developer/mocking-out-then-aws-sdk-for-go-for-unit-testing/.

hongil0316:
https://www.notion.so/gruntwork/Cloud-Nuke-Restructure-Proposal-845f291e952844ae8b0054ad0e466cfc

PR #497, opened 2023-07-12
Mocking AWS Calls for Unit Testing for API Gateway

hongil0316:
## Description

Making changes in code to have proper unit testing without calling AWS API - https://aws.amazon.com/blogs/developer/mocking-out-then-aws-sdk-for-go-for-unit-testing/. 

Currently, we call AWS API within unit tests and it causes several problems: 
* QuotaLimit
* Timeout

Also, I believe we can reduce the code inside `GetAllResources` function significantly with the new pattern of encapsulating "get resource" logic inside the ResourceType. Afterall, a better name for `ResourceType` would be just `Resource` but will make changes incrementally once this change looks good for everyone. 

Tackling issue: https://github.com/gruntwork-io/cloud-nuke/issues/494

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated [Mocking AWS Calls for Unit Testing for API Gateway]

### DIFF
--- a/aws/test_utils.go
+++ b/aws/test_utils.go
@@ -1,0 +1,19 @@
+package aws
+
+import (
+	"testing"
+
+	terratestaws "github.com/gruntwork-io/terratest/modules/aws"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/stretchr/testify/require"
+)
+
+func GetTestSession(t *testing.T, approvedRegions []string, forbiddenRegions []string) *session.Session {
+	region := terratestaws.GetRandomStableRegion(t, approvedRegions, forbiddenRegions)
+	session, err := session.NewSession(&aws.Config{Region: aws.String(region)})
+	require.NoError(t, err)
+
+	return session
+}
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -68,6 +68,7 @@ type ResourceType struct {
 
 type FilterRule struct {
 	NamesRegExp []Expression `yaml:"names_regex"`
+	// TODO(james): consider adding time filter rule here instead of passing in as function argument in other place.
```
ellisonc, line 71: I like this todo idea, it always seemed weird how that got tossed into ever function call

aws/test_utils.go
```diff
@@ -0,0 +1,19 @@
...
+	"testing"
+
+	terratestaws "github.com/gruntwork-io/terratest/modules/aws"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/stretchr/testify/require"
+)
+
+func GetTestSession(t *testing.T, approvedRegions []string, forbiddenRegions []string) *session.Session {
```
denis256, line 13: Hm, maybe I miss something, but the function `GetTestSession()` is not invoked in any place

hongil0316: I modified the implementation a bit afterward. It's no longer being used. Will delete it.

PR #517, opened 2023-07-23
Refactor SQS Resource Type

hongil0316:
## Description

Part of https://github.com/gruntwork-io/cloud-nuke/issues/494. 

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated [Refactor SQS Resource Type]

### DIFF
--- a/aws/sqs.go
+++ b/aws/sqs.go
@@ -1,14 +1,13 @@
 package aws
 
 import (
+	"github.com/gruntwork-io/cloud-nuke/config"
 	"github.com/gruntwork-io/cloud-nuke/telemetry"
 	commonTelemetry "github.com/gruntwork-io/go-commons/telemetry"
-	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	awsgo "github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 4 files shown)


aws/sqs.go
```diff
@@ -1,24 +1,21 @@
...
 	"github.com/gruntwork-io/cloud-nuke/logging"
 	"github.com/gruntwork-io/cloud-nuke/report"
 	"github.com/gruntwork-io/go-commons/errors"
 )
 
 // Returns a formatted string of SQS Queue URLs
-func getAllSqsQueue(session *session.Session, region string, excludeAfter time.Time) ([]*string, error) {
-	svc := sqs.New(session)
-
+func (sq SqsQueue) getAll(configObj config.Config) ([]*string, error) {
```
zachreborn, line 18: I have what might be a nitpick here, the shortening of SQS Queue to `sq` instead of `sqs`. I get why, because your other shorthand naming conventions but this one feels a bit too short. Getting into that "what's this mean?" territory.

zachreborn (changes requested):

zachreborn (approved):

aws/sqs.go
hongil0316: One caveat is it conflicts with the import of aws sqs library. I agree that it's more readable if we name is `sqs` but just to avoid conflicts with the import, `sq` might work better here.

Issue #651, opened 2024-03-11
Merge `awsNuke` and `awsDefaults` Logics


hongil0316:
## Context/Background

Right now, you will notice that we have separate routes for `awsNuke` and `awsDefaults`: 
- awsDefaults: nuking default VPCs
- awsNuke: nuking other resources

However, there seems to be quite a lot of changes we can merge and streamline. 

* `ec2.go` file currently contains nuking operations for a lot of resources (e.g., internet gateway, subnet, route tables). Let's separate it out to individual resources. 
* we support functionality on individual resources to support nuking default vs. non-default resources
* we make changes in `awsDefaults` to follow similar logic as `awsNuke` to nuke default resources. 
* we deprecate `awsDefaults` and create a parameter or config in `awsNuke` to target default resources.

hongil0316:
HI @Hrushi20, I'm actually working on this change!

Hrushi20:
Are there any issues I can work on?

PR #673, opened 2024-04-05
Refactor nuking default resources code.

hongil0316:
## Description

Partially Fixing https://github.com/gruntwork-io/cloud-nuke/issues/651
TODO: make sure to use default on all EC2 resources. 

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/aws/resources/ec2_vpc.go
+++ b/aws/resources/ec2_vpc.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	cerrors "errors"
 	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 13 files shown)


aws/resources/ec2_vpc.go
```diff
@@ -57,14 +57,21 @@ func (v *EC2VPCs) getFirstSeenTag(vpc ec2.Vpc) (*time.Time, error) {
 	return nil, nil
 }
 
-func (v *EC2VPCs) getAll(c context.Context, configObj config.Config) ([]*string, error) {
+func (v *EC2VPCs) getAll(_ context.Context, configObj config.Config) ([]*string, error) {
+	// Note: This filter initially handles non-default resources and can be overridden by passing the only-default filter to choose default VPCs.
+	var isdefault = "false"
```
denis256, line 62: make sense to use boolean variables? if left as string, in the future(during refactoring maybe) someone may try to put different value than true/false

hongil0316: Hey @denis256, this is the value we pass in as part of the `DescribeVpcs` endpoint. 
We use boolean inside the configObj for this. Essentailly you need to set either `true` or `false` as a string. That's how the API is defined.

denis256: Yes, I was thinking of representing value before passing it to AWS, and don't allow invalid value to be set in:

```
var isdefault := false
...
Values: awsgo.StringSlice([]string{fmt.Sprintf("%v", isdefault})
...
...
```

`isdefault` can take only valid values, and will be transformed in string when passed to AWS

denis256 (approved):

PR #674, opened 2024-04-08
SME-256 S3 deletion

### LINKED ISSUE #659: Issue with S3 deletion
## Description

The `emptyBucket` function in `s3.go` only lists and deletes object versions if the bucket is versioned. However, the bucket versioning can be turned off while still retaining versions and this prevents the bucket from being deleted.

A good example is when Terraform is able to delete the `aws_s3_bucket_versioning` resource, but fails to delete the bucket. The bucket now cannot be deleted with `cloud-nuke`.

## Possible solution

Ignore the version setting and delete object versions always.

autero1:
## Description

Fixes #659.

- As bucket versioning is managed separately and you can turn off versioning after the bucket is created, we need to check if there are any versions in the bucket regardless of the versioning status.

- Restore EIP deletion in PHXDevOps

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,6 @@ jobs:
               --exclude-resource-type ecr \
               --exclude-resource-type config-recorders \
               --exclude-resource-type config-rules \
-              --exclude-resource-type eip \
               --exclude-resource-type nat-gateway \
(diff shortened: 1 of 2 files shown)


hongil0316 (approved):
LGTM

PR #823, opened 2024-12-23
Skip if tag value is empty or set to true

### LINKED ISSUE #822: Feature Request: Exclude Resources by Tag Existence
### Feature Request: Exclude Resources by Tag Existence in Cloud Nuke

**Current Behavior:**
Cloud Nuke currently supports excluding resources based on a specified key only if its value is `"true"`.

**Feature Request:**
Add support for excluding resources based on the existence of a tag key, regardless of its value.

**Use Case:**
For example, if a resource has the tag `Component: rds`, I would like to exclude it by specifying `isTagExist: Component`. This ensures that all resources with the key `Component` (regardless of its value) are excluded from nuking.

**Proposed Solution:**
Introduce a new parameter `isTagExist` that accepts a tag key. If the specified key exists on a resource, the resource will be excluded.

**Benefits:**
- Provides more flexibility for users to manage exclusions.
- Simplifies scenarios where values of tags are dynamic or irrelevant.

**Example Configuration:**
```yaml
excludeTags:
  isTagExist: Component
```

Would exclude any resource with the tag `Component` present, such as:
- `Component: rds`
- `Component: database`
- `Component: test`

Let me know if more details are needed!

wakeful:
## Description

Fixes #822, change the exclude by tag behaviour to allow matching by an empty tag value.

## TODOs

- [x] Update the docs.

## Release Notes (draft)
change the `exclude by tag` behaviour to allow matching by an empty tag value, fix #822.

### Migration Guide
n/a


james03160927 (commented):
I suggest changing the existing behavior slightly: if the exclusion tag has no value, we should assume the tag itself indicates that the resource should be excluded. If a value is present, we should check whether it is set to "true."

wakeful:
> I suggest changing the existing behavior slightly: if the exclusion tag has no value, we should assume the tag itself indicates that the resource should be excluded. If a value is present, we should check whether it is set to "true."

Agreed, this simplifies the whole logic quite a lot.

james03160927:
https://github.com/gruntwork-io/cloud-nuke/issues/822#issuecomment-2562274489 - To accommodate this, we could introduce a TagValue property within the FilterRule. The TagValue could default to true, be set to a specific value being targeted, or use *, indicating that it can match any value.

How does this sound?

wakeful:
> [#822 (comment)](https://github.com/gruntwork-io/cloud-nuke/issues/822#issuecomment-2562274489) - To accommodate this, we could introduce a TagValue property within the FilterRule. The TagValue could default to true, be set to a specific value being targeted, or use *, indicating that it can match any value.
> 
> How does this sound?

I added `tag_value` to the configuration as a regular expression, primarily because there is already a `matches(name string, regexps []Expression) bool` function.


The original request should now be easy to address using this configuration. cc: @ErezWeiss
```yaml
s3:
  exclude:
    tag: Component
    tag_value: .*
```

ErezWeiss:
@james03160927
@wakeful 
 it seems amazing.
Can we proceed?

james03160927 (approved):
LGTM. Passed all the tests.

ErezWeiss:
@james03160927 thank you.
Can you guys build a version for it?

PR #896, opened 2025-07-22
fix: tag inclusion filtering to properly exclude untagged resources

### LINKED ISSUE #886: Filtering resources with including tags also includes resources with no tags
I'm currently working on adding tag filtering logic for OIDC Providers (should be no big deal and coming soon) and I think I discovered a bug in general resource filtering by tags:

When I configure include filter with tags I also get resources in the result which have no tags at all. I think the problem lies here: 
https://github.com/gruntwork-io/cloud-nuke/blob/ff4b0cb8f03875b3973e692c25c5eb9f4414d58b/config/config.go#L517

When I remove the check for `nil` and empty map, it works for my case with including resources by tags. But not sure what about excluding resources by tag. 

WDYT? Can this be changed? I would be eager to file a PR. 

<sub>Jan Roehrich <jan.roehrich@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/LEGAL_IMPRINT.md)</sub>

james03160927:
## Summary
Fixes #886 - Resources without tags are now correctly excluded when tag inclusion filters are specified.

## Changes
- Modified `ShouldIncludeBasedOnTag` to properly handle untagged resources when inclusion rules exist
- Added comprehensive unit tests for tag filtering behavior

## Testing
Added test cases covering all tag filtering scenarios including untagged resources with inclusion/exclusion rules.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -514,7 +514,7 @@ func (r ResourceType) ShouldInclude(value ResourceValue) bool {
 		return false
 	} else if value.Time != nil && !r.ShouldIncludeBasedOnTime(*value.Time) {
 		return false
-	} else if value.Tags != nil && len(value.Tags) != 0 && !r.ShouldIncludeBasedOnTag(value.Tags) {
+	} else if !r.ShouldIncludeBasedOnTag(value.Tags) {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


james03160927:
HI @roehrijn, PTAL.

roehrijn (approved):

denis256 (approved):

PR #899, opened 2025-08-04
fix(ekscluster): filtering for tags is omitted

### LINKED ISSUE #898: Bug in filtering for ekscluster
In the `filter()` function for `EKSClusters`, there is a shortcut implemented: the `DescribeCluster()` API endpoint is only called the name-based filter is matching. However, if one doesn't want to match by name, but by tags or by time for example, this doesn't work because the code never reaches the point where it is able to evaluate these properties. 

How to reproduce:
e.g. configure a inclusion rule with a tag but empty name matcher. 

Expected result:
The EKS cluster having the tag should be included in the result.

Actual result:
The EKS cluster having the tag is not included in the result.

I created a unit test to reproduce the issue. It is part of the fix PR I'm going to attach soon.

roehrijn:
## Description

Fixes #898.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `ekscluster` resource to allow proper filtering for time and tags without a name matcher.

### DIFF
--- a/aws/resources/eks.go
+++ b/aws/resources/eks.go
@@ -10,7 +10,6 @@ import (
 	"github.com/gruntwork-io/cloud-nuke/config"
 	"github.com/gruntwork-io/cloud-nuke/logging"
 	"github.com/gruntwork-io/cloud-nuke/report"
-	"github.com/gruntwork-io/cloud-nuke/util"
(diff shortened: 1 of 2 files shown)


james03160927 (approved):
LGTM

PR #914, opened 2025-09-01
feat(rds-cluster): add deletion protection support

james00012:
## Summary
- Add support for RDS cluster deletion protection to prevent accidental deletion of protected clusters
- Automatically disable deletion protection before nuking clusters
- Improve code readability with constants and better documentation

## Features
- **Deletion Protection Support**: RDS clusters with deletion protection enabled are now handled correctly
- **Configurable Behavior**: New `IncludeDeletionProtected` config option to control whether protected clusters should be included for deletion
- **Automatic Protection Removal**: When nuking clusters, deletion protection is automatically disabled first

### Core Functionality
- Modified `getAll()` to skip deletion-protected clusters by default (unless `IncludeDeletionProtected` is true)
- Enhanced `nukeAll()` to disable deletion protection before attempting cluster deletion
- Added `ModifyDBCluster` API support to handle protection removal
- Updated config to use `AWSProtectectableResourceType` for RDS clusters

### Code Improvements
- Added named constants for retry configuration in `waitUntilRdsClusterDeleted()`
- Improved comment clarity throughout the codebase
- Enhanced test documentation with numbered test cases

## Test plan
- [x] Added tests for deletion-protected cluster handling
- [x] Verified clusters with deletion protection are excluded by default
- [x] Verified clusters with deletion protection are included when `IncludeDeletionProtected: true`
- [x] Tested time-based exclusion still works correctly
- [x] All existing tests pass

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,6 +35,8 @@ type Config struct {
 	CloudWatchAlarm                 ResourceType                  `yaml:"CloudWatchAlarm"`
 	CloudWatchDashboard             ResourceType                  `yaml:"CloudWatchDashboard"`
 	CloudWatchLogGroup              ResourceType                  `yaml:"CloudWatchLogGroup"`
+	CloudMapNamespace               ResourceType                  `yaml:"CloudMapNamespace"`
+	CloudMapService                 ResourceType                  `yaml:"CloudMapService"`
 	CloudtrailTrail                 ResourceType                  `yaml:"CloudtrailTrail"`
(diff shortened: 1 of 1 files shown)


gcagle3:
This looks really well done! We should update the README to include something like a "Notes for deleting RDS Clusters with deletion protection enabled" or similar that explains how/when to use the new IncludeDeletionProtected option. I think this would be valuable to help alleviate confusion between nukeall or preserving deletion protection.

PR #922, opened 2025-09-08
Feat/ecscluster tag support

ErezWeiss:
Implements tag-based filtering for ECS clusters to support include/exclude
rules based on resource tags. This allows users to selectively target
or protect ECS clusters using tag expressions with AND/OR logic.

Key changes:
- Added getAllTags() method to retrieve cluster tags
- Modified getAll() method to include tags in ShouldInclude filtering
- Added comprehensive test cases for tag filtering scenarios
- Enhanced mock to return different tags for different test clusters

Supports all existing cloud-nuke tag functionality:
- Default exclusion tag (cloud-nuke-excluded: true)
- Time-based protection (cloud-nuke-after: date)
- Custom include/exclude tag rules with regex patterns
- Multiple tag conditions with AND/OR operators


ErezWeiss:
@james00012 can we publish a new version? It's also needed to update the README with all the new capabilities

PR #958, opened 2025-11-03
refactor: reorganize commands directory for better code organization

james00012:
## Summary

Refactors the monolithic `commands/cli.go` (~800 lines) into focused, modular files for better code organization and maintainability.

## Changes

**Commands Directory:**
- `aws_commands.go` (new) - AWS command handlers
- `gcp_commands.go` (new) - GCP command handlers  
- `flags.go` (new) - Reusable flag definitions
- `helpers.go` (new) - Shared utility functions
- `cli.go` (simplified) - CLI setup only (~100 lines)
- `cli_test.go` (updated) - Enhanced tests
- `errors.go` (updated) - New structured error types

**Supporting:**
- `config/config.go` - Added `ApplyTimeFilters()` helper (10 lines)

## Notes

- GCP `--resource-type` and `--exclude-resource-type` flags are currently ignored (will be implemented in future PR)
- Added timeout validation to prevent negative/zero values
- No functional changes to CLI behavior

**Stats:** 10 files changed, 963 insertions(+), 755 deletions(-)

### DIFF
--- a/commands/aws_commands.go
+++ b/commands/aws_commands.go
@@ -3,7 +3,6 @@ package commands
 import (
 	"github.com/gruntwork-io/cloud-nuke/aws"
 	"github.com/gruntwork-io/cloud-nuke/config"
-	"github.com/gruntwork-io/cloud-nuke/logging"
 	"github.com/gruntwork-io/cloud-nuke/telemetry"
 	"github.com/gruntwork-io/cloud-nuke/ui"
 	"github.com/gruntwork-io/go-commons/errors"
@@ -18,12 +17,7 @@ import (
 // awsNuke is the main command handler for nuking (deleting) AWS resources.
 // It supports resource filtering, time-based filtering, and config file overrides.
 func awsNuke(c *cli.Context) error {
-	telemetry.TrackEvent(commonTelemetry.EventContext{
-		EventName: "Start aws",
-	}, map[string]interface{}{})
-	defer telemetry.TrackEvent(commonTelemetry.EventContext{
-		EventName: "End aws",
-	}, map[string]interface{}{})
+	defer telemetry.TrackCommandLifecycle("aws")()
 
 	// Handle the --list-resource-types flag
 	if c.Bool(FlagListResourceTypes) {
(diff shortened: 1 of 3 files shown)


commands/aws_commands.go
```diff
@@ -0,0 +1,236 @@
...
+func awsDefaults(c *cli.Context) error {
+	telemetry.TrackEvent(commonTelemetry.EventContext{
+		EventName: "Start aws-defaults",
+	}, map[string]interface{}{})
+	defer telemetry.TrackEvent(commonTelemetry.EventContext{
+		EventName: "End aws-defaults",
+	}, map[string]interface{}{})
+
+	// Parse and set log level
+	parseErr := logging.ParseLogLevel(c.String(FlagLogLevel))
```
denis256: Should invoke parseLogLevel()? same as on line 30

commands/flags.go
```diff
@@ -0,0 +1,170 @@
...
+	return &cli.StringFlag{
+		Name:     FlagProjectID,
+		Usage:    "GCP Project ID to nuke resources from.",
+		Required: true,
+	}
+}
+
+// CombineFlags combines multiple flag slices into one
+func CombineFlags(flagSets ...[]cli.Flag) []cli.Flag {
+	var combined []cli.Flag
```
denis256: Cibsuder to pre-allocate `combined` slice

commands/helpers.go
```diff
@@ -0,0 +1,198 @@
...
+	if timeout != nil {
+		configObj.AddTimeout(timeout)
+	}
+
+	return nil
+}
+
+// confirmNuke handles the nuke confirmation prompt and countdown
+// Returns true if the nuke should proceed, false otherwise
+func confirmNuke(c *cli.Context, hasResources bool) (bool, error) {
```
denis256: Context state is ignored; I suspect it should be checked if context is done

commands/gcp_commands.go
```diff
@@ -0,0 +1,173 @@
...
+		gcp.NukeAllResources(account, nil)
+		ui.RenderRunReportWithFormat(outputFormat, outputFile)
+	}
+
+	return nil
+}
+
+// handleGetGcpResourcesWithFormat retrieves all GCP resources matching the filters and renders them
+// in the specified output format. This is used for both inspect and nuke operations.
+func handleGetGcpResourcesWithFormat(c *cli.Context, configObj config.Config, projectID string, outputFormat string, outputFile string) (
```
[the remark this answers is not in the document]
james00012: I think we should pass the context over to leverage e.g., contextual logging going forward. Right now the code is quite basic so not really using it at the moment.

commands/helpers.go
james00012: ditto.

commands/flags.go
james00012: Good catch! In this case, CombineFlags is only called during CLI initialization (not in a hot path), and we're combining a small number of flag sets. The performance benefit of pre-allocation would be negligible here, while the two-pass approach adds complexity (e.g., calculating the length and preallocating). I think the simpler, more readable version is better suited for this use case. WDYT?

PR #975, opened 2025-12-16
feat: use mise.toml for tool version management

### LINKED ISSUE #974: Use `mise.toml`
Switch over to mise for management of Golang+ dependencies in this project.

See how [Terragrunt handled this](https://github.com/gruntwork-io/terragrunt/blob/main/mise.toml). Note that this also controls the versions used in CI/CD, which saves us a lot of hassle in coordinating versions in local development and in CI.

james00012:
## Summary
- Add `mise.toml` with Go 1.24.0 and golangci-lint 1.64.8
- Update CircleCI lint job to install tools via mise
- Centralizes tool version management, matching terragrunt's approach

Closes #974

## Test plan
- [x] Verify lint job passes with mise-installed tools

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,18 +18,19 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Install Go 1.24
+          name: Install mise
           command: |
-            curl -sSL https://go.dev/dl/go1.24.0.linux-amd64.tar.gz | tar -xz -C /tmp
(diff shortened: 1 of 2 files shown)


james00012:
Got it. Thanks for the comment @denis256 @yhakbar. I can definitely move things to github action instead of relying on circleCI for us to run lint & tests & scheduling jobs. I can merge this and create a follow-up PR for that.

PR #1040, opened 2026-03-08
feat: add GCP Artifact Registry resource

Amit2465:
### How it works
**List:** Unlike Cloud Functions, Artifact Registry does *not* support the wildcard `locations/-`. The code first queries all available regions for the project via `ListLocations`, and then iterates through each region to query repositories via `ListRepositories`. Results are filtered by name and time (`CreateTime`) using the existing config filter logic.

**Delete:** Sends a delete request and waits for the long-running operation to complete. Already-deleted repositories (404) are handled gracefully. A 5-second delay is applied between deletions to avoid hitting API quota limits.

**Identifier format:** Full resource name — `projects/{project}/locations/{location}/repositories/{repository}`

**Note on [go.mod](cci:7://file://wsl.localhost/Ubuntu/home/amit/cloud-nuke/go.mod:0:0-0:0) changes:** Adding `cloud.google.com/go/artifactregistry/apiv1` caused Go to automatically upgrade several shared transitive dependencies (grpc, opentelemetry, google.golang.org/api) via minimum version selection. These are not manual upgrades; they are required by the new package.

### Test output:
```text
=== RUN   TestArtifactRegistryRepositories_ResourceName
--- PASS: TestArtifactRegistryRepositories_ResourceName (0.00s)
=== RUN   TestArtifactRegistryRepositories_MaxBatchSize
--- PASS: TestArtifactRegistryRepositories_MaxBatchSize (0.00s)
PASS
ok  	github.com/gruntwork-io/cloud-nuke/gcp/resources	0.022s
```

## TODOs

- [x] Update the docs.
- [x] Attention Grunts - if this PR adds support for a new resource, ensure the `nuke_sandbox` and `nuke_phxdevops` jobs in [.circleci/config.yml](cci:7://file://wsl.localhost/Ubuntu/home/amit/cloud-nuke/.circleci/config.yml:0:0-0:0) have been updated with appropriate exclusions (either directly in the job or via the `.circleci/nuke_config.yml` file) to prevent nuking IAM roles, groups, resources, etc that are important for the test accounts. *(Note: Checked because `nuke_sandbox` and `nuke_phxdevops` only run `aws`, so no exclusions are needed for this new GCP resource).*

## Release Notes (draft)

Added GCP Artifact Registry (`artifact-registry`) as a new supported resource type.

### Migration Guide

This PR introduces `artifact-registry` as a new GCP resource type. Since `cloud-nuke` automatically includes all registered resource types, Artifact Registry repositories will be nuked by default if no config file is provided. 
To opt out, add the following to your `cloud-nuke.yml` config file:
```yaml
ArtifactRegistry:
  exclude:
    names_regex:
      - ".*"

### CODE REVIEW

File: go.mod [hunk 1]
```diff
@@ -70,6 +71,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/urfave/cli/v2 v2.10.3
 	google.golang.org/api v0.247.0
+	google.golang.org/genproto v0.0.0-20250603155806-513f23925822
```
james00012: `google.golang.org/genproto` is a large module. Could `cloud.google.com/go/location` be used instead for `ListLocationsRequest`/`ListLocationsResponse`? If not, this is fine.

james00012 (dismissed):

james00012 (dismissed):
Clean implementation, follows established patterns well. `InitializationError` over `panic()` is a nice improvement. A few things to address:

**Blocking:**
1. `Nuke()` needs an `InitializationError` check too — nil client panic if called directly
2. `ShouldInclude` gets the full resource path (`projects/.../repositories/foo`), so anchored `names_regex` like `^foo$` won't work. Extract the short repo name for filtering.
3. Consider protecting against nuking system repos (`gcr.io`, `us.gcr.io`, etc.)

**Non-blocking suggestions:**
4. Replace `time.Sleep(5s)` with a context-aware select so cancellation isn't blocked
5. One failed location aborts the entire scan — consider logging and continuing
6. 5s delay is 10x more conservative than the 600 req/min quota requires

james00012:
> One final comment outstanding. The InitializationError field added to resource/resource.go is the right approach, but it's never actually set. artifact_registry.go:39 still uses panic() in InitClient, and there's no recover() in Resource.Init() or GcpResourceAdapter.Init() to catch it. If the Artifact Registry API is disabled, the process crashes instead of gracefully skipping. Fix: set the error directly instead of panicking:
> 
> ```
>   client, err := artifactregistry.NewClient(context.Background())
>   if err != nil {
>       r.InitializationError = fmt.Errorf("failed to create Artifact Registry client: %v", err)
>       return
>   }
> ```
> 
> Everything else looks great — nice work on the short name extraction, GCR-managed repo protection, context-aware delay, and location error handling.

Actually let me address this separately in a follow-up PR.

PR #1061, opened 2026-03-13
fix(gcp): use InitializationError instead of panic in gcs-bucket and cloud-function

Amit2465:
## Description

Fixes the client initialization error handling in `gcs-bucket` and `cloud-function` resources. Both were using `panic()` when the GCP client failed to initialize, which would crash the process if the API is disabled. Updated to set `r.InitializationError` and return instead.       

  ## TODOs

  ## Release Notes (draft)

  Updated `gcs-bucket` and `cloud-function` to gracefully handle client initialization failures instead of panicking.


Amit2465:
Hey @james00012, looks like #1062 handles this at the adapter level by adding a recover() in GcpResourceAdapter.Init(), which is a cleaner approach than fixing each resource individually. Closing this PR since it's covered there.

PR #1144, opened 2026-06-10
feat: Adding parallelism

yhakbar:
## Description

Adds `--parallelism` flag and parallelizes destroys where possible to get them sped up.

We were doing our destroys sequentially, which became unsustainably slow.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Added `--parallelism` CLI flag to control concurrent resource deletions (defaults to system CPU count).
  * Enabled parallel resource scanning and deletion for both AWS and GCP providers.

* **Bug Fixes**
  * Improved resource discovery concurrency and ordering consistency.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -4,8 +4,12 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync"
 	"time"
 
+	"golang.org/x/sync/errgroup"
+
+	"github.com/gruntwork-io/cloud-nuke/aws/resources"
 	"github.com/gruntwork-io/cloud-nuke/util"
 	"github.com/hashicorp/go-multierror"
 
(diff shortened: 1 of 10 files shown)


james00012 (approved):
LGTM overall. The concurrency model is right (single global cap instead of nested), the telemetry race is fixed, and global-last ordering is handled well. Tests build and pass.

One thing to note (not blocking): parallel region nuking can race cross-region dependencies that the old serial loop happened to avoid, e.g. RDS Aurora global clusters, TGW peering attachments, cross-region VPC peering. These are non-fatal and self-heal on the next run, so it's fine, just a possible one-run lag. If we want zero transient errors later, we can put those cross-region globals in the same "delete last" bucket as global.

Follow-ups, also non-blocking:
- Default --parallelism is GOMAXPROCS, which is 2 on our CircleCI small runners, so jobs that don't pass the flag only get ~2x. A fixed default (~10) would unlock the full speedup.
- No adaptive retry, so if we crank parallelism up, throttled lists silently skip. Worth adding RetryModeAdaptive if we raise the default.

None of these block merge.

PR #1145, opened 2026-06-11
Review fixes for parallelism (#1144)

james00012:
Stacked on top of #1144. Addresses the review points, verified against phxdevops.

What changed:
1. Default `--parallelism` is a fixed 10, not `GOMAXPROCS`. On container runners GOMAXPROCS reflects the CPU quota (measured 2 on a CircleCI `small`), not the IO concurrency this workload can use. Also exposes the flag on `inspect-aws` / `inspect-gcp`.
2. Single global concurrency cap. The scan now runs one errgroup over (region, type) instead of nested per-region and per-type limits, which multiplied to parallelism squared. Total in-flight calls are now bounded by `--parallelism`.
3. Adaptive retry. Adds `RetryModeAdaptive` + higher max attempts so the extra concurrency is absorbed by client-side rate limiting instead of surfacing as throttling errors (a throttled list silently skips that resource type).
4. Global nuked last. Regional regions nuke concurrently, then the `global` pseudo-region, preserving cross-region and global dependency order (IAM, RDS global clusters, peering). Within-region order was already preserved by the registry-index re-sort.
5. Telemetry race fix. Account ID is fetched once before the scan instead of from each region goroutine.
6. Made `TestSimpleBatchDeleter` thread-safe so `go test -race` passes.

Verification (read-only `inspect-aws` against phxdevops, 3 regions):

| parallelism | time | resources found | throttling |
|---|---|---|---|
| 1 | 255s | 88 | 0 |
| 10 | 53s | 88 | 0 |
| 20 | 53s | 88 | 0 |

~4.8x faster, identical results, no throttling. `go test -race` green on util/resource/commands/externalcreds. Nuke `global`-last validated by code and unit tests, not by a destructive run.

Ref: OSS-3751

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -45,126 +45,132 @@ func GetAllResources(c context.Context, query *Query, configObj config.Config, c
 
 	parallelism := util.GetParallelism(c)
 
-	var accountMu sync.Mutex
+	// Build per-region sessions and registered resources up front (sequential and
+	// cheap). The account ID is identical across all regions, so it is fetched
+	// once here and stored on the context, avoiding a data race on the telemetry
+	// package global that concurrent per-region writes would otherwise cause.
+	type regionWork struct {
+		region    string
+		resources []*resources.AwsResource
+	}
+	var regionWorks []regionWork
+	accountIDFetched := false
+	for _, region := range query.Regions {
+		cloudNukeSession, errSession := NewSession(region)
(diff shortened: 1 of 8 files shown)


james00012:
Closing this. The review fixes should fold directly into #1144 rather than living as a separate stacked PR. Commit preserved on branch review/parallelism-fixes.

PR #1147, opened 2026-06-12
fix: treat fast-deleting service-linked roles as success on NoSuchEntity

### LINKED ISSUE #1143: `iam-service-linked-role`: `GetServiceLinkedRoleDeletionStatus` returns 404 NoSuchEntity for fast-deleting roles, causing false failures
## Summary

When cloud-nuke deletes IAM service-linked roles, some roles complete deletion so quickly on the AWS side that the deletion task record is already purged by the time cloud-nuke polls `GetServiceLinkedRoleDeletionStatus`. AWS returns `404 NoSuchEntity`, which cloud-nuke treats as a hard failure — even though the role was actually deleted successfully.

This causes cloud-nuke to exit with errors and a non-zero exit code, even when all targeted resources were successfully removed.

- **Introduced in**: v0.47.0 (PR #1046 removed the initial `time.Sleep(3s)` before the first poll)
- **Last working version**: v0.46.0

Prior to v0.47.0, an unconditional `time.Sleep(3 * time.Second)` before the first `GetServiceLinkedRoleDeletionStatus` call accidentally masked this issue for most roles. PR #1046 replaced the hand-rolled polling loop with `util.PollUntil`, which calls the condition immediately with no initial delay, exposing the race.

Observed consistently for the following roles (all are AWS-managed service-linked roles that delete near-instantaneously):

- `AWSServiceRoleForAccessAnalyzer`
- `AWSServiceRoleForObservabilityAdmin`
- `AWSServiceRoleForOrganizations`
- `AWSServiceRoleForSSO`
- `AWSServiceRoleForSupport`
- `AWSServiceRoleForTrustedAdvisor`

The issue is not limited to these roles — any service-linked role that AWS deletes faster than the first poll interval can trigger this.

```
ERROR [Failed] iam-service-linked-role AWSServiceRoleForOrganizations: polling IAM ServiceLinked Role
AWSServiceRoleForOrganizations deletion: operation error IAM: GetServiceLinkedRoleDeletionStatus,
https response error StatusCode: 404, RequestID: 5a4c93b0-9cbe-407c-9da7-********,
NoSuchEntity: Cannot find deletion status for given id.
```

```
1 error occurred:
  * [global] iam-service-linked-role: 6 errors occurred:
    * AWSServiceRoleForAccessAnalyzer: polling IAM ServiceLinked Role ... NoSuchEntity: Cannot find deletion status for given id.
    * AWSServiceRoleForObservabilityAdmin: ...
    * AWSServiceRoleForOrganizations: ...
    * AWSServiceRoleForSSO: ...
    * AWSServiceRoleForSupport: ...
    * AWSServiceRoleForTrustedAdvisor: ...
```

## How to reproduce

1. Use cloud-nuke >= v0.47.0
2. Run `cloud-nuke aws --resource-type iam-service-linked-role` against an AWS account that has any of the affected roles
3. Observe that cloud-nuke reports the roles as failed despite them being deleted

## Root cause

**File**: `aws/resources/iam_service_linked_role.go`

`DeleteServiceLinkedRole` returns a `DeletionTaskId`. For some roles, AWS completes deletion synchronously and purges the task record almost immediately. When `GetServiceLinkedRoleDeletionStatus` is called with a stale task ID, AWS returns `404 NoSuchEntity`.

The current polling code propagates this as a fatal error:

```go
status, err := client.GetServiceLinkedRoleDeletionStatus(ctx, &iam.GetServiceLinkedRoleDeletionStatusInput{
    DeletionTaskId: taskId,
})
if err != nil {
    return false, errors.WithStackTrace(err)  // 404 treated as hard failure
}
```

---

## Proposed fix

A `NoSuchEntity` response on `GetServiceLinkedRoleDeletionStatus` does **not** by itself confirm the role is gone — the task record could be missing for other reasons (invalid ID, AWS-side issue) while the role still exists.

The correct fix is to treat 404 as a signal to verify actual role existence:

```go
if err != nil {
    var notFoundErr *types.NoSuchEntityException
    if goerr.As(err, &notFoundErr) {
        // Task record gone — verify the role itself no longer exists
        _, roleErr := client.GetRole(ctx, &iam.GetRoleInput{
            RoleName: aws.String(roleName),
        })
        var roleNotFound *types.NoSuchEntityException
        if goerr.As(roleErr, &roleNotFound) {
            return true, nil  // Role is gone — actual success
        }
        // Role still exists — this is a real failure
        return false, errors.WithStackTrace(err)
    }
    return false, errors.WithStackTrace(err)
}
```

This avoids falsely reporting success if the task ID was invalid but the role was never actually deleted.

---

## Workaround

Pin to v0.46.0 until a fix is released

james00012:
## Summary

Stop cloud-nuke from falsely reporting IAM service-linked role deletions as failures when AWS deletes the role faster than the first status poll.

## Intent

Fixes #1143. Some service-linked roles (AWSServiceRoleForOrganizations, AWSServiceRoleForSSO, AWSServiceRoleForSupport, AWSServiceRoleForTrustedAdvisor, etc.) delete near-instantaneously. AWS purges the deletion-task record before cloud-nuke polls GetServiceLinkedRoleDeletionStatus, which then returns 404 NoSuchEntity. The old code treated that 404 as a hard failure, so cloud-nuke exited non-zero and reported the roles as failed even though they were deleted. This regressed in v0.47.0 when #1046 replaced the hand-rolled poll loop (which had an initial 3s sleep that masked the race) with util.PollUntil (immediate first poll).

## Decision

On NoSuchEntity from the status poll, verify the role's actual existence with GetRole rather than assuming success. A 404 on the task record alone does not prove the role is gone (the task ID could be invalid for other reasons), so we confirm against the role itself.

GetRole has three outcomes, all handled explicitly:
- NoSuchEntity -> role is gone -> success
- role still present -> task record missing but role not deleted -> clear error
- any other error (e.g. throttling) -> surface that error rather than masking it with the original task-not-found error

Alternative considered: blindly treat the status-poll 404 as success. Rejected because it would hide genuine failures where the task ID was bad but the role still exists. Returning the original 404 when the role survives was also rejected during review (misleading message); we return a role-still-present error instead.

## What changed

- `aws/resources/iam_service_linked_role.go`: added GetRole to the IAMServiceLinkedRolesAPI interface; on NoSuchEntity from the status poll, branch on GetRole as described above.
- `aws/resources/iam_service_linked_role_test.go`: mock now supports injecting errors on GetServiceLinkedRoleDeletionStatus and GetRole; added three delete-path tests.

## Illustration

Before: status poll 404 -> hard failure (role was actually deleted)

After:
```
status poll 404 (NoSuchEntity)
  -> GetRole
       NoSuchEntity   -> success
       role present   -> error "role is still present"
       other error    -> propagate that error
```

## Validation

- Unit tests: `go test ./aws/resources/ -run TestIAMServiceLinkedRoles` — PASS (5 tests). Covers happy path, task-purged-and-role-gone (success), task-missing-but-role-present (error), and GetRole-fails (error surfaced).
- Build + vet + gofmt: `go build ./...`, `go vet ./aws/resources/`, `gofmt -l` — all clean.
- Local E2E (read-only) against the phxdevops test account (087285199408): `cloud-nuke inspect-aws --resource-type iam-service-linked-role` listed ~30 service-linked roles including the exact ones reported in the issue. Confirms the list/identify path works end-to-end with no regression.
- Not run: the actual nuke against live roles. The fast-delete race is non-deterministic and nuking would destroy real SLRs in the shared test account. The deletion branch is covered by unit tests instead.

## Known limitations / follow-ups

- The 404-then-verify branch is exercised only by unit tests; the real race was not reproduced live (non-deterministic, destructive). No follow-up planned unless it recurs in the field.

## Bug Fixes

* **Enhanced IAM Service Linked Role deletion** – Improved reliability of the deletion process by adding verification when deletion status checks return not found, ensuring proper confirmation of role removal in edge cases such as rapid purges.

### DIFF
--- a/aws/resources/iam_service_linked_role.go
+++ b/aws/resources/iam_service_linked_role.go
@@ -94,15 +94,24 @@ func deleteIAMServiceLinkedRole(ctx context.Context, client IAMServiceLinkedRole
 				// record is purged before we poll for its status, yielding a
 				// NoSuchEntity (404) error. That alone does not confirm the role is
 				// gone (the task ID could be invalid for other reasons), so verify the
-				// role itself no longer exists before reporting success.
+				// role itself via GetRole before reporting success.
 				var notFoundErr *types.NoSuchEntityException
 				if goerr.As(err, &notFoundErr) {
 					_, roleErr := client.GetRole(ctx, &iam.GetRoleInput{
 						RoleName: roleName,
 					})
 					var roleNotFound *types.NoSuchEntityException
-					if goerr.As(roleErr, &roleNotFound) {
+					switch {
+					case goerr.As(roleErr, &roleNotFound):
+						// Role is gone: deletion succeeded and AWS purged the task record.
 						return true, nil
+					case roleErr != nil:
+						// Could not confirm the role's state; surface the verification
+						// error rather than masking it with the task-not-found error.
+						return false, errors.WithStackTrace(roleErr)
+					default:
+						// Role still exists: the task record is missing but the role was not deleted.
+						return false, fmt.Errorf("deletion task for IAM ServiceLinked Role %s no longer exists but the role is still present", aws.ToString(roleName))
 					}
 				}
 				return false, errors.WithStackTrace(err)
--- a/aws/resources/iam_service_linked_role.go
+++ b/aws/resources/iam_service_linked_role.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"context"
+	goerr "errors"
 	"fmt"
 	"strings"
 	"time"
@@ -19,6 +20,7 @@ import (
 // IAMServiceLinkedRolesAPI defines the interface for IAM Service Linked Role operations.
 type IAMServiceLinkedRolesAPI interface {
 	ListRoles(ctx context.Context, params *iam.ListRolesInput, optFns ...func(*iam.Options)) (*iam.ListRolesOutput, error)
+	GetRole(ctx context.Context, params *iam.GetRoleInput, optFns ...func(*iam.Options)) (*iam.GetRoleOutput, error)
 	DeleteServiceLinkedRole(ctx context.Context, params *iam.DeleteServiceLinkedRoleInput, optFns ...func(*iam.Options)) (*iam.DeleteServiceLinkedRoleOutput, error)
 	GetServiceLinkedRoleDeletionStatus(ctx context.Context, params *iam.GetServiceLinkedRoleDeletionStatusInput, optFns ...func(*iam.Options)) (*iam.GetServiceLinkedRoleDeletionStatusOutput, error)
 }
@@ -88,6 +90,21 @@ func deleteIAMServiceLinkedRole(ctx context.Context, client IAMServiceLinkedRole
 				DeletionTaskId: deletionData.DeletionTaskId,
 			})
 			if err != nil {
+				// Some roles delete so quickly on the AWS side that the deletion task
+				// record is purged before we poll for its status, yielding a
+				// NoSuchEntity (404) error. That alone does not confirm the role is
+				// gone (the task ID could be invalid for other reasons), so verify the
+				// role itself no longer exists before reporting success.
+				var notFoundErr *types.NoSuchEntityException
+				if goerr.As(err, &notFoundErr) {
+					_, roleErr := client.GetRole(ctx, &iam.GetRoleInput{
+						RoleName: roleName,
+					})
+					var roleNotFound *types.NoSuchEntityException
+					if goerr.As(roleErr, &roleNotFound) {
+						return true, nil
+					}
+				}
 				return false, errors.WithStackTrace(err)
 			}
 
--- a/aws/resources/iam_service_linked_role_test.go
+++ b/aws/resources/iam_service_linked_role_test.go
@@ -18,10 +18,17 @@ type mockedIAMServiceLinkedRoles struct {
 	ListRolesOutput                          iam.ListRolesOutput
 	DeleteServiceLinkedRoleOutput            iam.DeleteServiceLinkedRoleOutput
 	GetServiceLinkedRoleDeletionStatusOutput iam.GetServiceLinkedRoleDeletionStatusOutput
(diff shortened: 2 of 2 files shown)


aws/resources/iam_service_linked_role.go
```diff
@@ -88,6 +90,30 @@ func deleteIAMServiceLinkedRole(ctx context.Context, client IAMServiceLinkedRole
...
 			})
 			if err != nil {
+				// Some roles delete so quickly on the AWS side that the deletion task
+				// record is purged before we poll for its status, yielding a
+				// NoSuchEntity (404) error. That alone does not confirm the role is
+				// gone (the task ID could be invalid for other reasons), so verify the
+				// role itself via GetRole before reporting success.
+				var notFoundErr *types.NoSuchEntityException
+				if goerr.As(err, &notFoundErr) {
+					_, roleErr := client.GetRole(ctx, &iam.GetRoleInput{
```
denis256, line 100: Hm, this will required iam:GetRole on the service-linked-role, make sense to document this?

aws/resources/iam_service_linked_role_test.go
```diff
@@ -108,3 +116,60 @@ func TestIAMServiceLinkedRoles_Delete(t *testing.T) {
...
+// non-NoSuchEntity error (e.g. throttling), we surface that error instead of
+// silently treating the role as deleted.
+func TestIAMServiceLinkedRoles_DeleteVerificationError(t *testing.T) {
+	t.Parallel()
+	client := mockedIAMServiceLinkedRoles{
+		DeleteServiceLinkedRoleOutput: iam.DeleteServiceLinkedRoleOutput{
+			DeletionTaskId: aws.String("task/aws-service-role/organizations.amazonaws.com/AWSServiceRoleForOrganizations/test-task-id"),
+		},
+		GetServiceLinkedRoleDeletionStatusErr: &types.NoSuchEntityException{Message: aws.String("Cannot find deletion status for given id.")},
+		GetRoleErr:                            fmt.Errorf("Throttling: Rate exceeded"),
```
denis256, line 169: Go discourages capitalized error strings

denis256 (approved):

